PR #4, opened 2019-11-11
CLI

liorblob:


demisto_sdk/core.py
```diff
@@ -1,20 +1,23 @@
...
-from .packages.splitter import Splitter
+
+from .yaml_tools.unifier import Unifier
+from .yaml_tools.extractor import Extractor
 from .common.configuration import Configuration
 
 
-class Client:
+class DemistoSDK:
+    SCRIPT = 'script'
```
Itay4: lets add constants file and define these there

demisto_sdk/main.py
```diff
@@ -0,0 +1,51 @@
...
+                                choices=[True, False], type=str2bool, default=None)
+
+    unify_parser.add_argument("-i", "--indir", help="The path to the files to unify", required=True)
+    unify_parser.add_argument("-o", "--outdir", help="The output dir to write the unified yml to", required=True)
+
+    args = parser.parse_args()
+
+    sdk = DemistoSDK()
+
+    if args.command == 'extract':
```
Itay4: is this only temp?

liorblob: it's the general direction.. why?

Itay4: can see how this quickly become messy, can we generlize this ?

PR #10, opened 2019-11-24
Package extractor unifier tests

guyfreund:
Tests for unifier.py & extractor.py


demisto_sdk/yaml_tools/unifier.py
```diff
@@ -140,16 +139,13 @@ def get_code_file(self, script_type):
...
-        script_path = list(filter(lambda x: not re.search(ignore_regex, x),
-                                  glob.glob(self.package_path + '*' + script_type)))[0]
-        return script_path
+        # We assume here the code file has the same name as the integration/script name, plus the addition of the type
+        package_name = os.path.basename(os.path.dirname(self.package_path))
+        code_file_path = self.package_path + package_name + script_type
+        if not os.path.isfile(code_file_path):
+            raise Exception('Code file does not exists or has different name than {}'
+                            .format(package_name + script_type))
+        return code_file_path
```
liorblob: What is this change?

demisto_sdk/yaml_tools/unifier.py
```diff
@@ -191,6 +187,7 @@ def insert_script_to_yml(self, script_type, yml_text, yml_data):
         return yml_text, script_path
 
     def get_script_package_data(self):
+        package_path = self.package_path
```
liorblob: this was fixed, merge from master

tests/extractor_test.py
```diff
@@ -0,0 +1,91 @@
...
+from demisto_sdk.common.constants import DEFAULT_IMAGE_BASE64
+import os
+import base64
+
+
+def test_get_yml_type():
+    from demisto_sdk.yaml_tools.extractor import Extractor
+    configuration = Configuration()
+    with patch.object(Extractor, '__init__', lambda a, b, c, d, e, f, g: None):
+        extractor = Extractor('', '', False, False, '', configuration)
```
liorblob: Can we create this instance once for all the tests?

tests/image_test.py
```diff
@@ -6,7 +6,7 @@ def test_is_not_default_image():
     int_path = os.path.normpath(os.path.join(__file__, '..', 'test_files', 'integration-Zoom.yml'))
     image.INTEGRATION_REGEX = int_path
     image_validator = image.ImageValidator(int_path)
-    assert image_validator.is_not_default_image() is True
+    assert image_validator.is_not_default_image() is False
```
liorblob: Why?

demisto_sdk/yaml_tools/unifier.py
guyfreund: In order to make our automation better we should assume assumptions, talked to @yaakovi and we decided to refactor it that way.

tests/image_test.py
guyfreund: is_not_default_image is just a string comparison, and we check for the base64 string representation of an image, so we don't need to check python's string comparison correctness, So I've changed the image in that integration to be also the default one so that we can use the same file also for other tests :)

demisto_sdk/yaml_tools/unifier.py
guyfreund: ok

tests/extractor_test.py
guyfreund: refactored, please take a look

liorblob: @guyfreund I meant all the instance creation in the unifier and extractor, can it be done one time only?

PR #17, opened 2019-11-26
Secrets Validator

yardensade:


demisto_sdk/core.py
```diff
@@ -126,3 +129,12 @@ def lint(self, project_dir: str, **kwargs):
...
         return linter.run_dev_packages()
+
+    def secrets(self, **kwargs):
+        sys.path.append(self.configuration.content_dir)
+
+        print_color('Starting secrets detection', LOG_COLORS.GREEN)
+
+        validator = SecretsValidator(configuration=self.configuration, **kwargs)
+
+        return validator.main()
```
liorblob: Do we really want to call it `main`?

yardensade: I will rename it to something cheesy soon

yardensade:
This pollution happened 7 days ago... how come it catches it just now, weird.

PR #31, opened 2019-11-27
Added hooks

jochman:
first run pre-commit install.
pre-commit autoupdate
enjoy


jochman:
> Add to readme an explanation how to set this up. Could be a short explanation with reference to:
> https://pre-commit.com/ .
> 
> Also, how do we run these validations in our build? Our build should make sure that mypy and flake8 pass (incase someone didn't use pre-commit).

our tox build is running flake8 and mypy.

glicht:
I see flake8 in tox.ini but no mypy.

glicht:
Also need to add pre-commit, flake8 and mypy to test-requirements.txt.

jochman:
@glicht https://github.com/demisto/demisto-sdk/pull/48 here's the added mypy to tox. we still can't run it on the repo, need to fix all mypy errors.
@roysagi

glicht:
Not sure we want this one: double-quote-string-fixer

jochman:
> Not sure we want this one: double-quote-string-fixer

our convention is to use `'` instead of `"`. so it'll change `"` to `'`.
shall I remove it and merge those two branches? the mypy will not fail the build.

PR #32, opened 2019-11-27
Fixing parent imports

jochman:
may the build will pass with this one


requirements.txt
```diff
@@ -1,3 +1,4 @@
+urllib3
```
liorblob: why?

yardensade: oh snap

PR #36, opened 2019-11-29
added common-server-python

jochman:
added commonserverpython to the root so  mypy and other tests will pass


glicht:
@jochman when is this being used? I don't want duplicate code.

jochman:
@gllicht it is used when validating mypy on files. mypy searches for the imports and validates that the imports exist.
a better solution is to copy the file from the git.

tests/linter_test.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+from demisto_sdk.dev_tools.linter import Linter
+
+
+class TestLinter:
+    path = "./CommonServerPython.py"
+
+    def setup_class(self):
+        if os.path.isfile(self.path):
```
yaakovi: if the test creates the file, this should be in teardown not setup.
if the file should be there regardless to the test, it shouldn't remove it (rename and revert at teardown)

demisto_sdk/dev_tools/linter.py
```diff
@@ -38,13 +40,37 @@ def __init__(self, project_dir: str, no_test: bool = False, no_pylint: bool = Fa
...
         self.run_args = {
             'pylint': not no_pylint,
             'flake8': not no_flake8,
             'mypy': not no_mypy,
             'tests': not no_test
         }
 
+    @staticmethod
+    def _get_common_server_python():
+        # type: () -> bool
```
liorblob: Please use python 3 types

demisto_sdk/dev_tools/linter.py
```diff
@@ -38,13 +40,37 @@ def __init__(self, project_dir: str, no_test: bool = False, no_pylint: bool = Fa
...
+        # Getting common server python in not exists
+        common_server_python_path = "./Scripts/CommonServerPython/CommonServerPython.py"
+        common_server_target_path = "./CommonServerPython.py"
+        common_server_remote_path = "https://raw.githubusercontent.com/demisto/content/master/Scripts/" \
+                                    "CommonServerPython/CommonServerPython.py"
+        if not os.path.isfile(common_server_target_path):
+            try:
+                shutil.copyfile(common_server_python_path, common_server_target_path)
+            except OSError:
+                # File not exists, trying to get from github.
```
liorblob: does not exist*

PR #37, opened 2019-12-01
align local behavior to circle

avidan-H:
always require unique file names when building content artifacts, both when building locally or in circle


yardensade:
Kol AKAVOD

PR #39, opened 2019-12-01
Fix file exists check

avidan-H:
Verifying that the .json file didn't exist wasn't checking the correct file path. Revert changes made to `get_code_file` function in `unifier.py`.


demisto_sdk/yaml_tools/unifier.py
```diff
@@ -177,20 +177,22 @@ def get_data(self, extension):
...
+        ignore_regex = (r'CommonServerPython\.py|CommonServerUserPython\.py|demistomock\.py|_test\.py'
+                        r'|conftest\.py')
+        if not self.package_path.endswith('/'):
+            self.package_path += '/'
+        if self.package_path.endswith('Scripts/CommonServerPython/'):
+            return self.package_path + 'CommonServerPython.py'
+
+        script_path = list(filter(lambda x: not re.search(ignore_regex, x),
+                                  glob.glob(os.path.join(self.package_path, '*' + script_type))))[0]
+        return script_path
```
liorblob: The test will fail now...

avidan-H: no, it doesn't. I'm magical!

PR #40, opened 2019-12-01
Allow specifying working directory

liorblob:
Add an argument to the main parser to specify a working directory, to run the command in that directory.

### DIFF
--- a/demisto_sdk/common/configuration.py
+++ b/demisto_sdk/common/configuration.py
@@ -3,12 +3,14 @@ import logging
 
 
 class Configuration:
-    def __init__(self, default_env_dir=None, log_verbose=False, logging_level=logging.INFO):
+    def __init__(self, env_dir=None, log_verbose=False, logging_level=logging.INFO):
         logging.basicConfig(level=logging_level)
         self.log_verbose = log_verbose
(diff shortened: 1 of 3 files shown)


demisto_sdk/common/configuration.py
```diff
@@ -3,12 +3,14 @@
...
-        if not default_env_dir:
-            self.env_dir = os.path.dirname(os.path.abspath(__file__))
+        self.sdk_env_dir = os.path.dirname(os.path.abspath(__file__))
+        if not env_dir:
+            self.env_dir = self.sdk_env_dir
         else:
-            self.env_dir = default_env_dir
-        self.envs_dirs_base = '{}/dev_envs/default_python'.format(self.env_dir)
+            self.env_dir = env_dir
+        self.envs_dirs_base = '{}/dev_envs/default_python'.format(self.sdk_env_dir)
```
yardensade: shouldn't it be `.format(self.env_dir)`?

liorblob: I don't think so because no normal human being will have those files in their repository

yardensade: What about cyborgs tho?

PR #43, opened 2019-12-02
fix default env_dir

jochman:
Ready

### DIFF
--- a/demisto_sdk/dev_tools/linter.py
+++ b/demisto_sdk/dev_tools/linter.py
@@ -28,7 +28,7 @@ class Linter:
             raise ValueError("Nothing to run as all --no-* options specified.")
 
         self.configuration = configuration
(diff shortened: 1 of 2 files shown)


demisto_sdk/dev_tools/linter.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, project_dir: str, no_test: bool = False, no_pylint: bool = Fa
             raise ValueError("Nothing to run as all --no-* options specified.")
 
         self.configuration = configuration
-        dev_scripts_dir = os.path.join(self.configuration.sdk_env_dir, 'scripts', 'dev_scripts')
+        dev_scripts_dir = os.path.join(self.configuration.env_dir, 'scripts', 'dev_scripts')
```
liorblob: the dev scripts are in the SDK dir. They won't be in the env dir

PR #45, opened 2019-12-02
fix path to dev env and scripts dir

jochman:
rdy


liorblob:
@jochman note the LGTM alert

PR #48, opened 2019-12-02
Add mypy to tox

jochman:
ready
but need to fix all mypy problems before merged


jochman:
something wrong happened with the diff. opening a new one

PR #53, opened 2019-12-04
Rn validator

teizenman:
## Status
In progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/18954
fixes: https://github.com/demisto/etc/issues/18608


tests/test_files/valid-multi-line_CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## [Unreleased]
+  - Comment #1.
+  - Comment #2.
```
ronykoz: think that should be without the space at the beginning of the line(and in all the other multilines)

tests/structure_test.py
```diff
@@ -141,7 +141,8 @@ def test_is_file_valid(self, source, target, answer):
...
         """Tests locked path (as reputations.json) so we won't override the file"""
-        mocker.patch.object(StructureValidator, "is_valid_file_path", return_value=answer)
-        structure = StructureValidator(source)
-        StructureValidator.scheme_name = scheme_name
-        assert structure.is_valid_file() is answer
+        # mocker.patch.object(StructureValidator, "is_valid_file_path", return_value=answer):
+        # structure = StructureValidator(source)
+        # StructureValidator.scheme_name = scheme_name
+        # Currently there isn't a scheme for reputations so test cant pass.
+        # assert structure.is_valid_file() is answer
```
ronykoz: let's fix this please

demisto_sdk/validation/file_validator.py
```diff
@@ -34,6 +34,7 @@
     LOG_COLORS, get_yaml, filter_packagify_changes, collect_ids, str2bool, get_pack_name, is_file_path_in_pack, \
     get_yml_paths_in_dir
 from demisto_sdk.yaml_tools.unifier import Unifier
+from demisto_sdk.common.hook_validations.release_notes import ReleaseNotesValidator
```
ronykoz: make it part of all the other hook_validations section please

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
+from __future__ import print_function
+
+import os
+import re
+from demisto_sdk.common.tools import run_command
+
+from demisto_sdk.common.tools import print_error, get_latest_release_notes_text, \
+    get_release_notes_file_path
+
+release_notes_standard = 'https://github.com/demisto/content/blob/master/docs/release_notes/README.md'
```
ronykoz: make it a class CONST please

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class ReleaseNotesValidator:
+    """Release notes validator is designed to ensure the existence and correctness of the release notes in content repo.
+
+    Attributes:
```
ronykoz: you are missing some attributes

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+        file_path (str): the path to the file we are examining at the moment.
+
+    """
+
+    def __init__(self, file_path):
+        # type: (str) -> None
```
ronykoz: not needed

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+        self.file_path = file_path
+        self.release_notes_path = get_release_notes_file_path(self.file_path)
```
ronykoz: i think that maybe this function should be here, where is it being used?

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+        self.release_notes = get_latest_release_notes_text(self.release_notes_path)
```
ronykoz: same here

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+        self.master_diff = self.get_master_diff()
+
+    def get_master_diff(self):
+        # type: () -> str
```
ronykoz: remove

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+        """Gets difference between current branch and origin/master
+
+        git diff with the --unified=10000 option means that if there exists a
+        difference between origin/master and current branch, the output will have at most 10000
+        lines of context.
```
ronykoz: why do we want so many? i think that 20 should do

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+
+        Returns:
+            str. empty string if no changes made or no origin/master branch, otherwise full difference context.
+
```
ronykoz: remove empty line

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        try:
```
ronykoz: why do we need try here?

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+        print_error(F'File {self.release_notes_path} is not formatted according to '
+                    F'release notes standards.\nFix according to {release_notes_standard}')
+        return False
+
+    def is_valid_release_notes_structure(self):
+        # type: () -> bool
+        """Validates that the release notes written in the correct manner.
+
+        one_line_release_notes_regex meaning:
+            dont start with any of the characters in the first []
```
ronykoz: in the first what?

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+            starts with a letter or number
+            ends with '.'
+
+        multi_line_release_notes_regex meaning:
+            starts with tab, then '-' ,then space
```
ronykoz: not sure about the tab, @yardensade could you share some insights?

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+            ends with '.'
```
ronykoz:
```suggestion
            each line ends with '.'
```

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+        Returns:
+            bool. True if release notes structure valid, False otherwise
+
+        """
+        hyphen_release_notes_regex = r'- ?'
+        one_line_release_notes_regex = r'[a-zA-Z0-9].*\.$'
+        multi_line_release_notes_regex = r'(\t+| {2,4})- .*\.$|- ?$'
+
+        release_notes_comments = self.release_notes.split('\n')
+        print(release_notes_comments)
```
ronykoz: remove

demisto_sdk/common/hook_validations/release_notes.py
ronykoz:
```suggestion
        self.latest_release_notes = get_latest_release_notes_text(self.release_notes_path)
```

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,164 @@
...
+        Returns:
+            bool. True if release notes file exists, False otherwise.
+        """
+        # checks that release notes file exists and contains text
+        if not os.path.isfile(self.release_notes_path):
+            print_error(F'File {self.file_path} is missing release notes, '
+                        F'Please add it under {self.release_notes_path}')
+            return False
+
+        if self.release_notes is None:
```
ronykoz: add that to the first condition

tests/release_notes_test.py
```diff
@@ -0,0 +1,91 @@
+from mock import patch
+import pytest
+from demisto_sdk.common.hook_validations.release_notes import ReleaseNotesValidator
```
ronykoz: split the third party from demisto_sdk imports

tests/validators_test.py
```diff
@@ -93,3 +91,84 @@ def test_is_file_valid(self, source, target, answer, validator):
...
+            copyfile(source_release_notes, target_release_notes)
+            patch.object(ReleaseNotesValidator, 'get_master_diff', side_effect=self.mock_get_master_diff)
+            validator = ReleaseNotesValidator(target_dummy)
+            assert validator.is_valid_release_notes_structure() is answer
+        finally:
+            os.remove(target_dummy)
+            os.remove(target_release_notes)
+
+    @staticmethod
+    def mock_get_master_diff():
```
ronykoz: shouldn't this be changed everytime?

demisto_sdk/common/hook_validations/release_notes.py
teizenman: it doesn't matter only if we take into consideration that the only part of the changelog file that will be changed is the `Unreleased` part. Otherwise I believe that 20 is insufficient. 10000 is just a matter of safety.

demisto_sdk/common/hook_validations/release_notes.py
teizenman: As I understand it, it's either a tab or something between 2 and 4 spaces.

tests/test_files/valid-multi-line_CHANGELOG.md
teizenman: According to https://github.com/demisto/content/blob/master/docs/release_notes/README.md that's the format.

tests/validators_test.py
teizenman: The `master_diff` is not relevant for this test, but we have to mock it for the test to work as expected.

demisto_sdk/common/hook_validations/release_notes.py
teizenman: It's also being used in the `base_validator`. I think it makes sense for it to be in the common tools.

demisto_sdk/common/hook_validations/release_notes.py
teizenman: Same as above.

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -72,45 +65,44 @@ def is_release_notes_changed(self):
...
+        release_notes_comments = self.latest_release_notes.split('\n')
 
         if len(release_notes_comments) == 1:
-            if not (re.match(hyphen_release_notes_regex, self.release_notes) or
-                    re.match(one_line_release_notes_regex, self.release_notes)):
+            if not (re.match(hyphen_release_notes_regex, self.latest_release_notes) or
+                    re.match(one_line_release_notes_regex, self.latest_release_notes)):
                 print_error(F'File {self.release_notes_path} is not formatted according to '
-                            F'release notes standards.\nFix according to {release_notes_standard}')
+                            F'release notes standards.\nFix according to {self.RELEASE_NOTES_STANDARD}')
```
guykeller: I'd consider changing this to LINK_TO_RELEASE_NOTES_STANDARDS.
Had to go up and see what it was to understand

demisto_sdk/common/hook_validations/release_notes.py
ronykoz: i don't want us to reach a point in which that function will give the entire changelog for some reason and then it will read 10000 which will take forever, lets put a more secure limit here, 100 or something of that sort, it should not be longer then even 10

tests/validators_test.py
ronykoz: so when is it relevant?

teizenman: Only to see whether a new comment was added. This test checks only the structure of the changes.

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,149 @@
...
+    def __init__(self, file_path):
+        self.file_path = file_path
+        self.release_notes_path = get_release_notes_file_path(self.file_path)
+        self.latest_release_notes = get_latest_release_notes_text(self.release_notes_path)
+        self.master_diff = self.get_master_diff()
+
+    def get_master_diff(self):
+        """Gets difference between current branch and origin/master
+
+        git diff with the --unified=200 option means that if there exists a
```
ronykoz:
```suggestion
        git diff with the --unified=100 option means that if there exists a
```

ronykoz:
@teizenman add CHANGELOG description please

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,152 @@
...
+                elif line.startswith('-') and not re.match(r'- *$', line):
+                    removes_in_diff += 1
+
+            # means that at least one new line was added
+            if adds_in_diff - removes_in_diff > 0:
+                return True
+
+        print_error(F'File {self.release_notes_path} is not formatted according to '
+                    F'release notes standards.\nFix according to {self.LINK_TO_RELEASE_NOTES_STANDARD}')
+        return False
```
guykeller: 1. Why is that not inside the general "if" statement? we would like to print an error if there's no difference?
2. Let's improve the error, please write the exact problem to fix

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,152 @@
...
+            Option #2:
+                starts with '-' with none or one trailing spaces.
+
+        multi_line_release_notes_regex meaning:
+            Option #1:
+                starts with tab or two to four spaces, then '-' ,then space
+                each line ends with '.'
+            Option #2:
+                starts with '-' with none or one trailing spaces.
+
```
guykeller: Discussed with @teizenman verbally

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,152 @@
...
+        release_notes_comments = self.latest_release_notes.split('\n')
+
+        if not release_notes_comments[-1]:
+            release_notes_comments = release_notes_comments[:-1]
+
+        if len(release_notes_comments) == 1:
+            if not re.match(one_line_release_notes_regex, self.latest_release_notes):
+                print_error(F'File {self.release_notes_path} is not formatted according to '
+                            F'release notes standards.\nFix according to {self.LINK_TO_RELEASE_NOTES_STANDARD}')
+                return False
```
guykeller: Same as above, let's be specific

demisto_sdk/common/hook_validations/release_notes.py
```diff
@@ -0,0 +1,152 @@
...
+        return True
+
+    def is_file_valid(self):
+        """Checks if given file is valid.
+
+        Return:
+            bool. True if file's release notes are valid, False otherwise.
+        """
+        # verifying that the other tests are even necessary
+        if not self.validate_file_release_notes_exists():
```
guykeller: Why not add this to the `validations` array?
`all()` will short circuit the tests anyway

teizenman: it won't, I tried it

demisto_sdk/common/hook_validations/release_notes.py
teizenman: Its difficult to do, since I cant know what the user has wanted to do.
I know what could have gone wrong, but not what the issue is.

demisto_sdk/common/hook_validations/release_notes.py
teizenman: Done

demisto_sdk/common/hook_validations/release_notes.py
teizenman: Done

ronykoz:
@teizenman Pull from master, and update the CHANGELOG please merge after that

PR #54, opened 2019-12-04
added missing constants test

jochman:
also fixed mypy for constants.py


yardensade:
Actually I take it back, you put the tests in the wrong folder, will open PR to fix it.

PR #56, opened 2019-12-05
glob Index Error fix

guyfreund:
fixes: https://github.com/demisto/etc/issues/20496


demisto_sdk/dev_tools/linter.py
```diff
@@ -79,7 +79,11 @@ def remove_common_server_python(self):
 
     def run_dev_packages(self) -> int:
         # load yaml
-        yml_path = glob.glob(self.project_dir + '/*.yml')[0]
+        try:
+            yml_path = glob.glob(self.project_dir + '/*.yml')[0]
```
liorblob: I think before we access the `[0]` item we should check if it returned any results instead of try/catch.

liorblob:
See this example:
```
 yml_files = glob.glob(os.path.join(self.package_path, '*.yml'))
        if not yml_files:
            raise Exception("No yml files found in package path: {}. "
                            "Is this really a package dir? If not remove it.".format(self.package_path))
```
We also have these lines at: `file_validator.py` lines 399 and 425.

demisto_sdk/dev_tools/linter.py
```diff
@@ -79,7 +79,9 @@ def remove_common_server_python(self):
 
     def run_dev_packages(self) -> int:
         # load yaml
-        yml_path = glob.glob(self.project_dir + '/*.yml')[0]
+        _, yml_path = get_yml_paths_in_dir(self.project_dir, Errors.no_yml_file(self.project_dir))
+        if not yml_path:
+            return 1
```
liorblob: I think we need to raise an exception here

PR #57, opened 2019-12-07
Add Bandit to linter

Itay4:
fixes https://github.com/demisto/etc/issues/20567


demisto_sdk/dev_tools/linter.py
```diff
@@ -298,20 +311,22 @@ def _setup_dev_files(self):
     def _get_lint_files(self):
         unifier = Unifier(self.project_dir)
         code_file = unifier.get_code_file('.py')
+        print('hi1')
```
liorblob: hi

PR #50, opened 2019-12-03
Add parallel linter

roysagi:
Added run_parallel_pkg_dev_tasks.py functionality to lint in demisto-sdk.
Using LintManager and -p flag to run several packages in parallel.

PR #59, opened 2019-12-08
Add multi package lint

roysagi:
opening this instead of https://github.com/demisto/demisto-sdk/pull/50

Added run_parallel_pkg_dev_tasks.py functionality to lint in demisto-sdk.
Using LintManager and -p flag to run several packages in parallel and in a non-parallel matter.
Also added several docstrings and better prints for user readability.
Also added --no-bc flag for lint command to not check the current files with $DIFF_COMPARE env variable.

fixes: https://github.com/demisto/etc/issues/20565
fixes: https://github.com/demisto/etc/issues/20600


demisto_sdk/dev_tools/linter.py
```diff
@@ -110,54 +136,99 @@ def run_dev_packages(self) -> int:
...
+            print("mypy completed for: {}\n".format(lint_files))
         finally:
             self.remove_common_server_python()
+            self.lock.release()
 
     def run_bandit(self, py_num):
+        """Run bandit
 
+        Args:
+            py_num: The python version in use
```
ronykoz: add type

demisto_sdk/dev_tools/linter.py
```diff
@@ -110,54 +136,99 @@ def run_dev_packages(self) -> int:
...
+
+        Returns:
+            None.
```
ronykoz: remove

demisto_sdk/dev_tools/linter.py
```diff
@@ -74,6 +92,9 @@ def get_common_server_python(self) -> bool:
 
     def remove_common_server_python(self):
         """checking if file exists and removing it
+
+            Returns:
+                None.
```
ronykoz: not needed

demisto_sdk/dev_tools/linter.py
```diff
@@ -94,13 +115,18 @@ def run_dev_packages(self) -> int:
             if script_type == 'powershell':
                 # TODO powershell linting
                 return 0
-            print('Script is not of type "python". Found type: {}. Nothing to do.'.format(script_type))
+            print('The script {} is not of type "python". Found type: {}. Nothing to do.'.format(self.project_dir,
+                                                                                                 script_type))
```
ronykoz: why do we have an empty line here?

ronykoz:
@yaakovi reminding you to go over this

demisto_sdk/dev_tools/linter.py
yaakovi: it's not empty

demisto_sdk/dev_tools/lint_manager.py
```diff
@@ -0,0 +1,240 @@
...
+        Returns:
+            bool. True if there is a difference and False otherwise.
+        """
+        diff_compare = os.getenv("DIFF_COMPARE")
+        if not diff_compare:
+            return True
+        if os.getenv('CONTENT_PRECOMMIT_RUN_DEV_TASKS'):
+            # if running in precommit we check against staged
+            diff_compare = '--staged'
+        res = subprocess.run(["git", "diff", "--name-only", diff_compare, "--", pkg_dir], text=True,
```
yaakovi: can we switch to `run_command` (`content:test_utils.py`)?

demisto_sdk/dev_tools/lint_manager.py
```diff
@@ -0,0 +1,240 @@
...
+                 no_bc_check: bool = False, configuration: Configuration = Configuration()):
+
+        if no_test and no_pylint and no_flake8 and no_mypy:
+            raise ValueError("Nothing to run as all --no-* options specified.")
+
+        self.parallel = parallel
+        self.log_verbose = verbose
+        self.root = root
+        self.parallel = parallel
+        self.max_workers = 10 if max_workers is None else max_workers
```
yaakovi: when will it be None?

demisto_sdk/dev_tools/lint_manager.py
```diff
@@ -0,0 +1,240 @@
...
+    def run_parallel_packages(self, pkgs_to_run: List[str]) -> int:
+        """Runs the Lint command in parallel on several threads.
+
+        Args:
+            pkgs_to_run: The packages to run in parallel
+
+        Returns:
+            int. 0 on success and 1 if any package failed
+        """
+        max_workers = int(self.max_workers)
```
yaakovi: redundant, it should already be int

roysagi:
Only 2 tasks remain to make this PR perfect:
1) Currently the feature to run only on packages changed from a git branch doesn't work (no-bc flag).
  - Come to think of it we might want to just add the possibility to run on packages that changed from master then we can have 2 modes: 
    a) for regular builds - run only on changed packages.
    b) for nightly - runs on all packages regardless of whether changes made.  

2) could be irrelevant right now but the prints regarding the docker actions are not subjected to a lock which can be problematic on parallel run... 
The reason this could wait is because if an error occurs the failure message will print all together and I added the failed dir name at the beginning so we can easily identify the failing package.

PR #69, opened 2019-12-16
fixing a lot of stuff in validate

roysagi:
Until now indicator fields were not checked.
After consulting with server we came to the conclusion that indicator fields and incident fields follow the same scheme.

Thus this PR makes indicator fields follow the same course as incident fields.

Also there was a bug in the check where there was a call to incident_field_validator.is_valid().
is_valid is a field not a function, I believe the intention was to call incident_field_validator.is_valid_file() which checks all elements of the file for validity.

lastly I added an error message in case a non supported file type was entered for validation.


roysagi:
@reutshal - if you can also look at the code and tell me what you think.

@amitdavidson234 - FYI

demisto_sdk/validation/file_validator.py
```diff
@@ -281,13 +281,19 @@ def validate_modified_files(self, modified_files):  # noqa: C901
                 if not image_validator.is_valid():
                     self._is_valid = False
 
-            elif re.match(INCIDENT_FIELD_REGEX, file_path, re.IGNORECASE):
+            # incident fields and indicator fields are using the same scheme.
+            elif re.match(INCIDENT_FIELD_REGEX, file_path, re.IGNORECASE) or \
+                    re.match(INDICATOR_FIELDS_REGEX, file_path, re.IGNORECASE):
```
ronykoz: lets create a list for both of them pls

demisto_sdk/validation/file_validator.py
```diff
@@ -281,13 +281,19 @@ def validate_modified_files(self, modified_files):  # noqa: C901
...
                 incident_field_validator = IncidentFieldValidator(structure_validator)
-                if not incident_field_validator.is_valid():
+                if not incident_field_validator.is_valid_file():
                     self._is_valid = False
                 if self.is_backward_check and not incident_field_validator.is_backward_compatible():
                     self._is_valid = False
 
+            else:
+                print_error("The file type of {} is not supported in validate command".format(file_path))
```
ronykoz: add a list of the supported file types please

demisto_sdk/validation/file_validator.py
```diff
@@ -339,15 +345,20 @@ def validate_added_files(self, added_files):
...
-            elif re.match(INCIDENT_FIELD_REGEX, file_path, re.IGNORECASE):
+            # incident fields and indicator fields are using the same scheme.
+            elif re.match(INCIDENT_FIELD_REGEX, file_path, re.IGNORECASE) or \
+                    re.match(INDICATOR_FIELDS_REGEX, file_path, re.IGNORECASE):
                 incident_field_validator = IncidentFieldValidator(file_path)
-                if not incident_field_validator.is_valid():
+                if not incident_field_validator.is_valid_file():
                     self._is_valid = False
+            else:
+                print_error("The file type of {} is not supported in validate command".format(file_path))
```
ronykoz: same here

demisto_sdk/validation/file_validator.py
```diff
@@ -18,10 +18,9 @@
 from demisto_sdk.common.configuration import Configuration
 from demisto_sdk.common.constants import CODE_FILES_REGEX, OLD_YML_FORMAT_FILE, SCHEMA_REGEX, KNOWN_FILE_STATUSES, \
     IGNORED_TYPES_REGEXES, INTEGRATION_REGEX, BETA_INTEGRATION_REGEX, BETA_INTEGRATION_YML_REGEX, SCRIPT_REGEX, \
-    IMAGE_REGEX, INCIDENT_FIELD_REGEX, TEST_PLAYBOOK_REGEX, \
-    INTEGRATION_YML_REGEX, DIR_LIST, PACKAGE_SUPPORTING_DIRECTORIES, \
+    IMAGE_REGEX, TEST_PLAYBOOK_REGEX, INTEGRATION_YML_REGEX, DIR_LIST, PACKAGE_SUPPORTING_DIRECTORIES, \
     YML_BETA_INTEGRATIONS_REGEXES, PACKAGE_SCRIPTS_REGEXES, YML_INTEGRATION_REGEXES, PACKS_DIR, PACKS_DIRECTORIES, \
-    Errors, PLAYBOOKS_REGEXES_LIST, INDICATOR_FIELDS_REGEX
+    Errors, PLAYBOOKS_REGEXES_LIST, JSON_ALL_INCIDENT_FIELD_REGEXES
```
ronykoz: i think that it doesn't include the pack regex for it

PR #73, opened 2019-12-19
sys.exit() fix

reutshal:
Added it in __main__.py so the SDK will actually return the exit code


ronykoz:
@liorblob what's the goal of having both main.py and __main__.py?
@yaakovi @reutshal can we merge this?

liorblob:
/_/_main/_/_.py is for the console entry, it shouldn't contain any logic and import the main.py file and call it.

PR #80, opened 2019-12-22
Adding beta integrations to the structure test

ronykoz:
Adding beta integrations to the structure test


ronykoz:
Not sure why there are so many files changed, must be something related to the pre-commit hook

amshamah419:
Looks like it stripped some whitespace. Either way looks good.

PR #82, opened 2019-12-23
Lint bug fix

roysagi:
The demisto-sdk changes to fix the lint command failure


demisto_sdk/__main__.py
```diff
@@ -1,9 +1,9 @@
-import os
+import sys
 from demisto_sdk.main import main
 
 
 def console_entry() -> None:
-    os._exit(main())
+    sys.exit(main())
```
Itay4: why is this change?

PR #84, opened 2019-12-23
Adding Beta integration image check

roysagi:
Adding Beta integration image check in validate command.


roysagi:
added support for packs as well

PR #85, opened 2019-12-23
Support API modules

liorblob:
Support importing API modules from the Modules pack and ignore the pack when creating content.

fixes: https://github.com/demisto/etc/issues/19965


ronykoz:
@liorblob lets go over this together please, i want to see the Content side of it too

demisto_sdk/yaml_tools/unifier.py
```diff
@@ -246,6 +255,46 @@ def get_script_package_data(self):
...
+    @staticmethod
+    def check_api_module_imports(script_code: str) -> Tuple[str, str]:
+        """
+        Checks integration code for API module imports
+        :param script_code: The integration code
+        :return: The import string and the imported module name
+        """
+
+        # General regex to find API module imports, for example: "from MicrosoftApiModule import *  # noqa: E402"
+        module_regex = r'from ([\w\d]+Module) import \*(?:  # noqa: E402)?'
```
ronykoz:
```suggestion
        module_regex = r'from ([\w\d]+ApiModule) import \*(?:  # noqa: E402)?'
```

demisto_sdk/dev_tools/linter.py
```diff
@@ -304,20 +305,42 @@ def _docker_run(self, docker_image):
...
+        if py_num > 3:
+            unifier = Unifier(self.project_dir)
+            code_file_path = unifier.get_code_file('.py')
+            try:
+                # Look for an import to an API module in the code. If there is such import, we need to copy the correct
+                # module file to the package directory.
+                with io.open(code_file_path, mode='r', encoding='utf-8') as script_file:
+                    _, module_name = unifier.check_api_module_imports(script_file.read())
+                if module_name:
+                    module_path = os.path.join(self.configuration.env_dir, 'Packs', 'ApiModules', 'Scripts',
```
ronykoz: this will not work from outside the content repo

tests/unifier_test.py
```diff
@@ -94,14 +117,22 @@ def test_insert_image_to_yml():
         assert yml_text == yml_text_test
 
 
+def test_check_api_module_imports():
```
ronykoz: can we check that we also do the replace correctly?

demisto_sdk/dev_tools/linter.py
liorblob: I guess we do not intend for it to work outside of content. Should I change this to something else?

ronykoz: maybe check if the path exists, if not raise an error stating that the user is not in the content repo/doesn't have this pack

demisto_sdk/dev_tools/linter.py
```diff
@@ -407,20 +408,46 @@ def _docker_run(self, docker_image):
...
+                                               module_name, module_name + '.py')
+                    print_v('Copying ' + os.path.join(self.configuration.env_dir, 'Scripts', module_path))
+                    if not os.path.exists(module_path):
+                        print_v('API Module {} not found, you might be outside of the content repository or this API'
+                                ' module does not exist'.format(module_name))
```
ronykoz: we should raise an exception here, shouldn't we?

liorblob: it will be caught in the `except` right afterwards. Should we raise there as well?

ronykoz: i would just raise en exception here rather then try and copy a non existing file

PR #96, opened 2019-12-29
Better outputs for secrets

roysagi:
* Created better outputs for secrets command.
* Added catch for general exception in extract.


demisto_sdk/core.py
```diff
@@ -126,7 +126,9 @@ def unify_package(self, package_path, dest_path):
                         "should contain either Integrations or Scripts directories")
 
         unifier = Unifier(package_path, directory_name, dest_path)
-        return unifier.merge_script_package_to_yml()
+        unifier_outputs = unifier.merge_script_package_to_yml()
+        print_color("Created unified yml: {}".format(unifier_outputs[0][0]), LOG_COLORS.GREEN)
```
ronykoz: add this msg in the unifier itself

roysagi: done

demisto_sdk/yaml_tools/unifier.py
```diff
@@ -126,7 +123,8 @@ def merge_script_package_to_yml(self):
             yml_text, desc_path = self.insert_description_to_yml(yml_data, yml_text)
 
         output_map = self.write_yaml_with_docker(yml_text, yml_data, script_obj)
-        return list(output_map.keys()), yml_path, script_path, image_path, desc_path
+        unifier_outputs = list(output_map.keys()), yml_path, script_path, image_path, desc_path
+        print_color("Created unified yml: {}".format(unifier_outputs[0][0]), LOG_COLORS.GREEN)
```
ronykoz: why did you remove the return?

roysagi: no need to return any thing. this is command will print any error it encounters and if it gets to this last line it means it finished successfully.

It used to return what now is entered to the unifier_outputs variable and then print in core.py - it now prints in place so the return is redundant.

PR #99, opened 2019-12-29
Adding git flag to lint command (instead of no-bc)

roysagi:
Adding git flag to lint command (instead of no-bc)

-g or --git will run lint on only changed packages (compared to master).


ronykoz:
Lets go over this together

PR #104, opened 2019-12-30
not checking RN for newly added integrations and scripts

roysagi:
fixes: https://github.com/demisto/etc/issues/20966

Removing validate check for RN in newly added integrations and scripts


ronykoz:
Lets add UT for this before merging

PR #112, opened 2020-01-01
update readme so autocomplete is clear

glicht:


README.md
```diff
@@ -20,7 +20,8 @@ You can use the SDK in the CLI as follows:
 For more information, run `demisto-sdk -h`.
 For more information on a specific command execute `demisto-sdk <command> -h`.
 
-Our CLI supports autocomplete for linux/MacOS machines, you can turn this feature on by running one of the following:
+### Autocomplete
+Our CLI supports autocomplete for Linux/MacOS machines, you can turn this feature on by running one of the following:
```
ronykoz: i think that we should suggest adding those lines in the ~/.bash_profile for mac users or ~/.zshrc for zsh users and ~/.bashrc for linux ones

glicht: I don't know. Please test and check. I just added the title and capitalized Linux.

PR #115, opened 2020-01-02
Reputation validations skip

teizenman:
Currently we do not have a validation or scheme for reputation.
The fix: as long as the file matches the reputation regex (right directory and right name structure), any file validation will be skipped and return OK.

fixes: https://github.com/demisto/etc/issues/20955


ronykoz:
Please note that this is a patch until @yaakovis team will add a proper validator

ronykoz:
@teizenman lets merge this please, fix the conflicts here

PR #116, opened 2020-01-02
Specific file validations

teizenman:
Run `validate` for specific pre-defined file.

fixes: https://github.com/demisto/etc/issues/20959


demisto_sdk/main.py
```diff
@@ -159,7 +160,7 @@ def unify(**kwargs):
     '-i', '--id-set', is_flag=True,
     default=False, show_default=True, help='Create the id_set.json file.')
 @click.option(
-    '-p', '--prev-ver', help='Previous branch or SHA1 commit to run checks against.')
+    '-pv', '--prev-ver', help='Previous branch or SHA1 commit to run checks against.')
```
ronykoz:
```suggestion
    '--prev-ver', help='Previous branch or SHA1 commit to run checks against.')
```

demisto_sdk/main.py
```diff
@@ -169,15 +170,32 @@ def unify(**kwargs):
 @click.option(
     '-g', '--use-git', is_flag=True, show_default=True,
     default=False, help='Validate changes using git - this will check your branch changes and will run only on them.')
+@click.option(
+    '-p', '--path', help='Path of file to validate specifically.'
+)
+# @click.option(
+#    '-t', '--file-type', type=click.Choice(
+#        ['integration', 'script', 'playbook', 'incidnet-field', 'indicator-field', 'image', 'release-notes', 'layout',
+#         'description']), help='If you want to validate a specific file, you can specify it\'s type.')
```
ronykoz: lets remove this

demisto_sdk/main.py
```diff
@@ -169,15 +170,32 @@ def unify(**kwargs):
...
 def validate(config, **kwargs):
     sys.path.append(config.configuration.env_dir)
 
-    validator = FilesValidator(configuration=config.configuration,
-                               is_backward_check=str2bool(kwargs['backward_comp']),
-                               is_circle=str2bool(kwargs['circle']), prev_ver=kwargs['prev_ver'],
-                               validate_conf_json=kwargs['conf_json'], use_git=kwargs['use_git'])
-    return validator.run()
+    # if not kwargs.get('path') and kwargs.get('file_type'):
+    #    click.echo(click.style('Ignoring file-type argument since no specific file was determined', fg='yellow'))
```
ronykoz: ?

demisto_sdk/main.py
```diff
@@ -169,15 +170,32 @@ def unify(**kwargs):
...
+
+    file_path = kwargs['path']
+
+    if not (file_path or os.path.isfile(file_path)):
+        print_error(F'File {file_path} was not found')
+        return 1
```
ronykoz: why ?
it should not be required

demisto_sdk/validation/file_validator.py
```diff
@@ -191,6 +199,17 @@ def get_modified_and_added_files(self, tag='origin/master'):
                 self.get_modified_files(all_changed_files_string,
                                         print_ignored_files=self.print_ignored_files)
 
+            if self.file_path:
```
ronykoz: if we use file_path we should not use all_changed_files_string

demisto_sdk/validation/file_validator.py
```diff
@@ -536,16 +553,22 @@ def is_valid_structure(self):
...
-            print("No using git, validating all files")
-            self.validate_all_files()
+            if self.file_path:
+                branches = run_command('git branch')
+                branch_name_reg = re.search(r'\* (.*)', branches)
+                self.branch_name = branch_name_reg.group(1)
+                self.validate_committed_files()
+            else:
+                print('No using git, validating all files')
+                self.validate_all_files()
```
ronykoz: could you explain this logic

demisto_sdk/main.py
teizenman: Done

demisto_sdk/main.py
teizenman: It was there in case we wanted to pre-define the type, but I just added the logic to the regular flow so it is not needed

demisto_sdk/main.py
teizenman: It was there in case we wanted to pre-define the type, but I just added the logic to the regular flow so it is not needed

demisto_sdk/validation/file_validator.py
teizenman: when the user run the specific file validator, without the `use_git` flag, the logic of receiving the working branch does not happen. So, to verify against master I need to fetch it's branch.
The `else` if for the regular validate without use_git.

teizenman: But, it would be smarter to make it into a function, so Im fixing that.

demisto_sdk/main.py
teizenman: Fixed

demisto_sdk/validation/file_validator.py
```diff
@@ -536,16 +555,20 @@ def is_valid_structure(self):
...
+                print('Validates all of Content repo directories according to their schemas')
                 self.validate_all_files()
         else:
-            print("No using git, validating all files")
-            self.validate_all_files()
+            if self.file_path:
+                self.branch_name = self.get_current_working_branch()
+                self.validate_committed_files()
+            else:
+                print('No using git, validating all files')
```
yaakovi:
```suggestion
                print('Not using git, validating all files')
```

ronykoz:
@teizenman update the readme.md and then merge

PR #118, opened 2020-01-03
Fix commands

ronykoz:
## Status
Ready

## Description
Fix several commands that were causing problems due to merge issues


ronykoz:
@dantavori please check the uploader in this version, it doesn't seem to work for me

ronykoz:
Seems to me that maybe something is not working with Demisto py, as i didn't change the logic of the upload itself

demisto_sdk/main.py
```diff
@@ -67,18 +66,16 @@ def main(config, version, env_dir):
     type=click.Choice([SCRIPT_PREFIX, INTEGRATION_PREFIX])
 )
 @click.option(
-    '--demisto-mock', '-d',
-    help="Add an import for demisto mock, true by default",
-    type=click.Choice(["True", "False"]),
-    default=True,
+    '--no-demisto-mock',
+    help="Don't add an import for demisto mock, false by default",
```
anara123:
```suggestion
    help="Don't add an import for demisto mock",
```
it is more confusing with false/true

demisto_sdk/main.py
```diff
@@ -159,10 +154,10 @@ def unify(**kwargs):
 @click.option(
     '-p', '--prev-ver', help='Previous branch or SHA1 commit to run checks against.')
 @click.option(
-    '-c', '--circle', type=click.Choice(["True", "False"]), default='False',
-    help='Is CircleCi or not')
+    '--post-commit', is_flag=True, help='Whether the validation is done after you committed the files. Mostly for '
```
anara123: not that clear description

demisto_sdk/main.py
```diff
@@ -159,10 +154,10 @@ def unify(**kwargs):
...
+                                        'build validations.')
 @click.option(
-    'b', '--backward-comp', type=click.Choice(["True", "False"]), default='False', show_default=True,
+    '-b', '--backward-comp', is_flag=True, show_default=True,
     help='To check backward compatibility.')
```
anara123: we should want to check by default backward compatibility

demisto_sdk/yaml_tools/unifier.py
```diff
@@ -136,6 +136,8 @@ def merge_script_package_to_yml(self):
         output_map = self.write_yaml_with_docker(yml_text, yml_data, script_obj)
         unifier_outputs = list(output_map.keys()), yml_path, script_path, image_path, desc_path
         print_color("Created unified yml: {}".format(unifier_outputs[0][0]), LOG_COLORS.GREEN)
+        # todo: add more prints here
```
anara123: remove todo comment

demisto_sdk/main.py
```diff
@@ -172,15 +167,17 @@ def validate(config, **kwargs):
...
-                               is_backward_check=str2bool(kwargs['backward-comp']),
-                               is_circle=str2bool(kwargs['circle']), prev_ver=kwargs['prev_ver'],
+                               is_backward_check=kwargs['backward_comp'],
+                               is_circle=kwargs['post_commit'], prev_ver=kwargs['prev_ver'],
                                validate_conf_json=kwargs['conf_json'], use_git=kwargs['use_git'])
     return validator.run()
 
 
 # ====================== create ====================== #
 @main.command(name="create",
```
anara123: we are moving this after all? if not we should discuss this first , and if this command stays they definitely the name should be changed. it is not enough to change the description i

ronykoz: Lets talk about that tomorrow, but i think that we can probably move that back into the content repo.
@yaakovi what do you think about this?

demisto_sdk/main.py
ronykoz: updated

dantavori:
@ronykoz I pip-installed this branch of `demisto-sdk` and the uploader worked for me without any problem. However, in a package case it adds this message (but still succeeds to execute):
image
If this validation is important then maybe we should consider adding a flag attribute to the `Unifier` class when it doesn't expect a legal path or something

ronykoz:
@dantavori did you try running it on most updated server?

dantavori:
@ronykoz yes

demisto_sdk/main.py
yaakovi: it is not something that one will actually use (beside content developers).
this feature can stay in the sdk if we want to support custom content zip files that one can generate and upload.

demisto_sdk/dev_tools/uploader.py
```diff
@@ -23,22 +23,27 @@ def __init__(self, path: str, insecure: bool = False, verbose: bool = False):
...
+
+            # Upload the file to Demisto
+            result = self.client.integration_upload(file=self.path)
+
+            # Print results
+            print_v(f'Result:\n{result.to_str()}', self.log_verbose)
+            print_color(f'Uploaded \'{result.name}\' successfully', LOG_COLORS.GREEN)
+
+        except Exception as ex:
+            raise ex
```
yaakovi: redundant, just leave the try-finally

README.md
```diff
@@ -69,7 +69,7 @@ Extract code, image and description files from a demisto integration or script y
                         will import unless this is CommonServerPython
 
 **Examples**:
-`demisto-sdk extract -i Integrations/integration-MyInt.yml -o Integrations/MyInt -m`
+`demisto-sdk split-yml -i Integrations/integration-MyInt.yml -o Integrations/MyInt -m`
```
anara123: `yml-to-package`?

PR #120, opened 2020-01-05
fixed playbook scheme

teizenman:
## Status
Ready

## Description
`format` now adds a `description` tag to playbook.
Fixed playbook scheme


ronykoz:
why do we need the allowempty here?

PR #125, opened 2020-01-05
Validate command docs

roysagi:
validate command docs - @reutshal FYI


ronykoz:
@roysagi if you can update this based on master please that would be great

PR #140, opened 2020-01-09
fix secrets

anara123:
## Status
In Progress

## Related Issues
fixes https://github.com/demisto/etc/issues/21172
fixes https://github.com/demisto/etc/issues/21173

## Description
Fix bug in secrets


yaakovi:
missing changelog update

PR #149, opened 2020-01-15
Unify fix

yaakovi:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21190
fixes: https://github.com/demisto/etc/issues/21374

## Description
updated unifier to work with dictionary instead of text for handling code, image, and description insertions.

## Screenshots


yaakovi:
content build: https://circleci.com/gh/demisto/content/tree/UnifyTest

tests/unifier_test.py
```diff
@@ -199,12 +227,193 @@ def test_insert_script_to_yml_exceptions(package_path, dir_name, file_path):
...
+        shutil.copy(image_file, os.path.join(package_path, f'{package_name}_image.png'))
+
+
+class TestMergeScriptPackageToYMLIntegration:
+    def setup(self):
+        self.test_dir_path = os.path.join('tests', 'test_files', 'Unifier', 'Testing')
+        os.makedirs(self.test_dir_path)
+        self.package_name = 'SampleIntegPackage'
+        self.export_dir_path = os.path.join(self.test_dir_path, self.package_name)
+        self.expected_yml_path = os.path.join(self.test_dir_path, 'integration-SampleIntegPackage.yml')
```
anara123: where is this file located? `integration-SampleIntegPackage.yml`

yaakovi: good catch, it was missing.

glicht:
@yaakovi when can we get this fixed?

demisto_sdk/commands/unify/tests/unifier_test.py
```diff
@@ -206,12 +234,244 @@ def test_insert_script_to_yml_exceptions(package_path, dir_name, file_path):
...
+            test_dir=self.test_dir_path,
+            package_name=self.package_name,
+            base_yml='demisto_sdk/tests/test_files/Unifier/SampleIntegPackage/SampleIntegPackage.yml',
+            script_code=TEST_VALID_CODE,
+            image_file='demisto_sdk/tests/test_files/Unifier/SampleIntegPackage/SampleIntegPackage_image.png',
+            detailed_description=description,
+        )
+
+        unifier = Unifier(self.export_dir_path, outdir=self.test_dir_path)
+        yml_files, orig_yml, orig_script, orig_image, orig_description = unifier.merge_script_package_to_yml()
```
anara123: what is the reason that `merge_script_package_to_yml()` returns all these file pathes?
shouldn't it just return the path to the generated yml?

demisto_sdk/commands/unify/unifier.py
```diff
@@ -220,43 +226,27 @@ def insert_script_to_yml(self, script_type, yml_text, yml_data):
...
-        elif self.dir_name == 'Integrations' or self.dir_name == 'Beta_Integrations':
-            if yml_data.get('script', {}).get('script'):
-                if yml_data['script']['script'] != '-' and yml_data['script']['script'] != '':
-                    raise ValueError("Please change the script to be blank or a dash(-) for package {}"
-                                     .format(self.package_path))
-        else:
-            raise ValueError('Unknown yml type for dir: {}. Expecting: Scripts/Integrations'.format(self.package_path))
+        if self.is_script_package:
+            if yml_data.get('script') not in ('', '-'):
+                raise ValueError(f'Please change the script to be blank or a dash(-) for package {self.package_path}')
```
anara123: why not just replace the script instead of asking the user to replace it with dash

demisto_sdk/commands/unify/unifier.py
```diff
@@ -220,43 +226,27 @@ def insert_script_to_yml(self, script_type, yml_text, yml_data):
...
+        return yml_unified, script_path
 
     def get_script_package_data(self):
+        # should be static method
         _, yml_path = get_yml_paths_in_dir(self.package_path, error_msg='')
         if not yml_path:
-            raise Exception("No yml files found in package path: {}. "
-                            "Is this really a package dir? If not remove it.".format(self.package_path))
+            raise Exception(f'No yml files found in package path: {self.package_path}. '
+                            'Is this really a package dir? If not remove it.')
```
anara123: remove what?

yaakovi: the given folder that doesn't include any yml file.
this happen when you switch between branches and leaving behind directories without any files (or just the unified file)

demisto_sdk/commands/unify/tests/unifier_test.py
yaakovi: idk, did not want to break something that depends on it (i'm assuming no one actually need them.
should i remove them?

demisto_sdk/commands/unify/unifier.py
yaakovi: we can switch that to a warning, but i don't think we should modify the original files.

anara123: we should, this is not normal situation that they have something in the script. It is better UX in this case

demisto_sdk/commands/unify/tests/unifier_test.py
anara123: no need for them, remove them. Return just the path to the unified yml, and make sure that yml is created as expected

PR #168, opened 2020-01-23
fix-feed-validation

jochman:
## Status
Ready


bakatzir:
fixes: https://github.com/demisto/etc/issues/21394

PR #171, opened 2020-01-26
Fix wrong term in dockerimage validation

jochman:
### Status
Ready

fixes: https://github.com/demisto/etc/issues/21472
fixes: https://github.com/demisto/etc/issues/21229

wait with merge for @GalRabin


yaakovi:
need a unit test for docker change

PR #174, opened 2020-01-26
adding config command

roysagi:
## Status
Ready

## Description
Added demisto-sdk config - this command will help create the DEMISTO_BASE_URL and DEMISTO_API_KEY env vars

Do with it what you will...

### DIFF
--- a/demisto_sdk/commands/config/config.py
+++ b/demisto_sdk/commands/config/config.py
@@ -1,0 +1,30 @@
+from demisto_sdk.commands.common.tools import run_command, print_color, LOG_COLORS
+
+
+class Config:
+    """Configure
+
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/config/config.py
```diff
@@ -0,0 +1,32 @@
...
+        self.base_url = base_url
+        self.api_key = api_key
+
+    def configure(self):
+        """Configure the env variables DEMISTO_BASE_URL and DEMISTO_API_KEY
+
+        Returns:
+            None.
+        """
+        while self.base_url is None or self.base_url == '':
```
yaakovi: add support to pass only one (if empty string don't export)

PR #176, opened 2020-01-26
Test github actions

GalRabin:
## Status
In Progress

## Description
Replacing circle ci to GitHub actions, Should we implement? same functionality
Can also deploy to pypi, much more faster and cheaper


glicht:
Let's stay with Circle.

PR #182, opened 2020-01-30
build optimization

GalRabin:
## Status
Ready

## Description
Splitting the PR as requested

1. Remove flake8, mypy . from tox.ini - due to duplicate verification - we already do it with pre-commit.
2. Enable not failing if one missing interpreter so it will be comforted to develop if you are not using venv and pyenv (Probably if its passing v3.7 it will pass v3.8) - How the verification of both versions will be in the build process.
3. Skipping a specific failing tests path.
4. There is many files that committed which changed by auto-pep8.


Itay4:
@yaakovi ?

GalRabin:
@Itay4 
We cancelled the change , I forgot to close the PR

PR #185, opened 2020-02-02
update playbook format

ronykoz:
## Status
Ready 

## Related Issues
fixes: https://github.com/demisto/etc/issues/21681

## Description
Check if a description already exists for a task and if so don't override it with an empty value

### DIFF
--- a/demisto_sdk/commands/format/update_playbook.py
+++ b/demisto_sdk/commands/format/update_playbook.py
@@ -25,6 +25,10 @@ class PlaybookYMLFormat(BaseUpdateYML):
             self.yml_data['description'] = ''
 
         for task_id, task in self.yml_data.get('tasks', {}).items():
+            if task['task'].get('description'):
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/format/update_playbook.py
```diff
@@ -25,6 +25,10 @@ def add_description(self):
             self.yml_data['description'] = ''
 
         for task_id, task in self.yml_data.get('tasks', {}).items():
+            if task['task'].get('description'):
+                import ipdb; ipdb.set_trace()
```
anara123: why you need this line?

idovandijk: Why do we need debugger here?

idovandijk: Anar commented, removing mine

ronykoz: my bad, removed it

PR #198, opened 2020-02-13
Adding version and from version test for incident fields

esharf:
## Status
Ready.

## Related Issues
fixes: https://github.com/demisto/etc/issues/21682

## Description
ensure that version is -1 and from version is at least 5.0.0.

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/incident_field.py
+++ b/demisto_sdk/commands/common/hook_validations/incident_field.py
@@ -305,12 +305,19 @@ class IncidentFieldValidator(BaseValidator):
         return is_valid
 
     def is_valid_from_version(self):
+        error_msg = None
         try:
             server_base_version = self.current_file.get('fromVersion').split('.')[0]
-        except AttributeError:
-            print_error(f'{self.file_path}: "fromVersion" as an invalid value.')
-        else:
             if server_base_version and int(server_base_version) >= 5:
                 return True
+            else:
+                error_msg = f'{self.file_path}: "fromVersion" mast be at least 5.0.0'
+        except AttributeError:
+            pass
+        except ValueError:
+            pass
 
+        if not error_msg:
+            error_msg = f'{self.file_path}: "fromVersion" as an invalid value.'
+        print_error(error_msg)
         return False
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/common/hook_validations/incident_field.py
```diff
@@ -301,3 +303,10 @@ def is_cliname_is_builtin_key(self):
                 f"{self.file_path}: cliName field can not be {cliname} as it's a builtin key."
             )
         return is_valid
+
+    def is_valid_from_version(self):
+        if self.current_file.get('fromVersion', '') < '5.0.0':
```
bakatzir: please take 1st char, <= 5 (as number)

demisto_sdk/commands/common/tests/incident_field_test.py
```diff
@@ -222,3 +222,31 @@ def test_is_valid_cliname_invalid(self, cliname, group):
...
+        structure.current_file = {"version": version}
+        validator = IncidentFieldValidator(structure)
+        assert validator.is_valid_version() == is_valid, f'is_valid_version({version}) returns {not is_valid}.'
+
+    data_is_valid_from_version = [
+        ('5.0.0', True),
+        ('4.', False),
+        ('', False),
+        ('4.0.0', False),
+    ]
```
bakatzir: add also
5.0.1
10.1.2

esharf: added an over 50 version 😀

demisto_sdk/commands/common/hook_validations/incident_field.py
```diff
@@ -301,3 +303,12 @@ def is_cliname_is_builtin_key(self):
                 f"{self.file_path}: cliName field can not be {cliname} as it's a builtin key."
             )
         return is_valid
+
+    def is_valid_from_version(self):
+        server_base_version = str(self.current_file.get('fromVersion', '0.0.0')).split('.')[0]
```
yuvalbenshalom: you should catch `ValueError` exception if in case the it will to translate it to integer, and return an error that say the version is invalid and print it.

And add a test for that please.

demisto_sdk/commands/common/hook_validations/incident_field.py
```diff
@@ -301,3 +303,21 @@ def is_cliname_is_builtin_key(self):
...
+            if server_base_version and int(server_base_version) >= 5:
+                return True
+            else:
+                error_msg = f'{self.file_path}: "fromVersion" mast be at least 5.0.0'
+        except AttributeError:
+            pass
+        except ValueError:
+            pass
+
+        if not error_msg:
```
bakatzir: Please handle the error message in the expect and print the error catched.
also, please use only 1 return in the function, and return a true/false value according to the flow.

PR #208, opened 2020-02-18
create content fix - Yamlordereddict -> ryaml load + test infra

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22093

## Description
using the ryaml load instead of the orderedloader
also added sanity unit test to the copy_dir_files in the content creator


avidan-H:
I'd feel more comfortable with @yaakovi's review as he worked intimately on this stuff

demisto_sdk/commands/create_artifacts/tests/contet_creator_test.py
```diff
@@ -0,0 +1,31 @@
+import filecmp
+from demisto_sdk.commands.create_artifacts.content_creator import *
+
+
+def test_copy_dir_files():
+    current_dir = os.path.dirname(__file__)
+    content_repo_example = os.path.join(current_dir, '/test_files/content_repo_example')
```
yaakovi: you already using join...
```suggestion
    content_repo_example = os.path.join(current_dir, 'test_files', 'content_repo_example')
```

demisto_sdk/commands/create_artifacts/tests/contet_creator_test.py
```diff
@@ -0,0 +1,31 @@
...
+    # delete all files in the content_bundle
+    for filename in os.listdir(content_bundle_full_path):
+        file_path = os.path.join(content_bundle_full_path, filename)
+        try:
+            if os.path.isfile(file_path) or os.path.islink(file_path):
+                os.unlink(file_path)
+            elif os.path.isdir(file_path):
+                shutil.rmtree(file_path)
+        except Exception as err:
+            print('Failed to delete %s. Reason: %s' % (file_path, err))
```
yaakovi: should be in teardown as this is not part of the test.

demisto_sdk/commands/create_artifacts/tests/contet_creator_test.py
```diff
@@ -0,0 +1,31 @@
+import filecmp
+from demisto_sdk.commands.create_artifacts.content_creator import *
+
+
+def test_copy_dir_files():
```
yaakovi: test is unclear, what are you trying to test here?

demisto_sdk/commands/create_artifacts/tests/contet_creator_test.py
```diff
@@ -0,0 +1,31 @@
+import filecmp
```
yaakovi: please update file name: `content_creator_test.py`

demisto_sdk/commands/create_artifacts/content_creator.py
```diff
@@ -169,17 +170,23 @@ def copy_dir_yml(self, dir_path, bundle):
...
 
             with open(path, 'r') as file_:
                 yml_info = yaml.safe_load(file_)
-
             ver = yml_info.get('fromversion', '0')
             print(f' - processing: {ver} ({path})')
-            copy_func(path, os.path.join(bundle, os.path.basename(path)), yml_info)
+
+            if dir_name in ['Playbooks', 'TestPlaybooks']:
+                self.copy_playbook_yml(path, os.path.join(bundle, os.path.basename(path)))
```
yaakovi: `TestPlaybooks` folder can contain scripts, make sure this is handled correctly

demisto_sdk/commands/create_artifacts/content_creator.py
```diff
@@ -169,17 +170,23 @@ def copy_dir_yml(self, dir_path, bundle):
...
+            else:
+                ryaml = YAML()
+                ryaml.allow_duplicate_keys = True
+                with io.open(path, mode='r', encoding='utf-8') as file_:
+                    yml_unified = ryaml.load(file_)
```
yaakovi: no actual reason to open the file twice (here and on lines 177-178).
read it once using `ruamel` package.

extra: you can also modify the function `Unifier.write_yaml_with_docker` to use only the unified version of the dict so you don't have to pass it twice.
extra2: same goes for the script object, it can be calculated inside the function.

demisto_sdk/commands/create_artifacts/tests/contet_creator_test.py
```diff
@@ -0,0 +1,31 @@
...
+from demisto_sdk.commands.create_artifacts.content_creator import *
+
+
+def test_copy_dir_files():
+    current_dir = os.path.dirname(__file__)
```
yaakovi: use absolute paths using `git_path()` (see other tests)

demisto_sdk/commands/create_artifacts/tests/contet_creator_test.py
anara123: `copy_dir_files` is internal details. 
instead test public functionality like `ContentCreator.run` function. 
And add test description, what is the behaviour that you are testing here

bakatzir: It's one path after the `create_content`, and I wanted to avoid the zipping on circle.
I agree we should do it, bur would preferred it in a 2nd batch, as this part have no tests at all.

demisto_sdk/commands/create_artifacts/content_creator.py
bakatzir: Done(I think)

demisto_sdk/commands/create_artifacts/tests/contet_creator_test.py
bakatzir: added in the test description.

demisto_sdk/commands/create_artifacts/content_creator.py
bakatzir: You don't want to get the scripts in the Testplaybook folder to the content bundle afaik.
what do you want to handle aside from that?

yaakovi: those scripts should go to the test_bundle as they are being used by the test playbooks

demisto_sdk/commands/create_artifacts/content_creator.py
```diff
@@ -168,25 +167,24 @@ def copy_dir_yml(self, dir_path, bundle):
         :return: None
         """
         scan_files, _ = get_yml_paths_in_dir(dir_path, error_msg='')
+        print(scan_files)
```
yaakovi: remove

demisto_sdk/commands/create_artifacts/tests/content_creator_test.py
```diff
@@ -0,0 +1,46 @@
...
+from tempfile import mkdtemp
+
+from demisto_sdk.commands.create_artifacts.content_creator import *
+from demisto_sdk.commands.common.git_tools import git_path
+
+
+class TestContentCreator:
+    def setup(self):
+        current_dir = os.path.normpath(os.path.join(__file__,
+                                                    f'{git_path()}/demisto_sdk/commands/create_artifacts', 'tests'))
```
yaakovi: `git_path` returns absolute path to the root of the repo, no need to use the current file location.
```
current_dir = f'{git_path()}/demisto_sdk/commands/create_artifacts/tests'
```

demisto_sdk/commands/create_artifacts/tests/content_creator_test.py
```diff
@@ -0,0 +1,46 @@
...
+        - copying the content folder to a content bundle(flatten files)
+        Then
+        - ensure files are being flatten correctly
+        - ensure no !!omap are added(due to ordereddict yaml problems)
+        """
+        content_creator = ContentCreator(artifacts_path=self.content_repo, content_version='2.5.0',
+                                         content_bundle_path=self._dir,
+                                         preserve_bundles=False)
+        print(self.scripts_full_path)
+        print(self._dir)
```
yaakovi: don't forget to remove

demisto_sdk/commands/create_artifacts/content_creator.py
bakatzir: @yaakovi tended to it. first added the test case and fixed accordingly. hopefully I did not miss an edge case.

PR #215, opened 2020-02-21
Vulture

Itay4:
Added [Vulture](https://github.com/jendrikseipp/vulture) to linter
[Running example](https://circleci.com/gh/demisto/content/41892#queue-placeholder/containers/1)

I've set the confidence level to 60 in order to catch things like:
image

image

Might lead to some false positives, but we can whitelist them


glicht:
I am concerned with the amount reported. Thinking of starting with 100% confidence instead of 60.

Itay4:
@glicht with 100% confidence level we will lose the whole point of vulture, as you can see there are only 3 alerts with 100% confidence, and most of the reports are actual dead code
for example, we will miss unused functions
i think the whitelist mechanism should help us handle false positives

glicht:
We have over 100 reported integrations/scripts. Need to discuss how we go about this. 

I am not sure how easy the whitelisting is. I really don't like that they don't have inline comments. See: https://github.com/jendrikseipp/vulture/issues/182 . I suggest taking one of the reported issues and creating a whitelist.

Itay4:
@glicht please see [this commit](https://github.com/demisto/content/commit/7803363b94515296e6a94e591b40fb4b73b9ae0c), which include unused vars and whitelisting of reports that came up in GmailSingleUser
i've used `--make-whitelist` to generate the whitelist automatically

anara123:
I think we have enough linters

anara123:
I think we have enough linters

anara123:
I think we have enough linters

Itay4:
@glicht updated confidence level to 100

demisto_sdk/commands/lint/linter.py
```diff
@@ -287,6 +293,37 @@ def run_bandit(self, py_num) -> int:
...
+
+        Args:
+            py_num: The python version in use
+
+        Returns:
+            int. 0 on successful vulture run, 1 otherwise.
+        """
+        lint_files = self._get_lint_files()
+        python_exe = 'python2' if py_num < 3 else 'python3'
+        cmd_args = [python_exe, '-m', 'vulture', lint_files, '--min-confidence 100']
```
glicht: Let's get the confidence level from an environment variable (with a default of 100). That way if a developer wants to run it with a lower confidence level she can simply set an environment variable.

RJ722:
Hello there, vulture dev here. Just an update: Vulture now supports `# noqa` comments -- changes are on master, and we'd probably make a release later today/tomorrow.

Itay4:
@RJ722 cool, we have been waiting for this feature
thanks for the update

PR #219, opened 2020-02-24
Add ability to unify with force

amshamah419:
## Status
Ready

## Related Issues
fixes: demisto/etc#22213

## Description
Adds a new flag to allow forcing of the unification process. Essentially the programmer's version of "You kids are going to stay in a room until you like each other"


amshamah419:
@bakatzir @yuvalbenshalom Only question I have is will the parser they are using allow for a boolean to be passed, or will I have to do some string to bool conversion?

amshamah419:
Did in fact need to use a form of str to bool. Tested and works well.

demisto_sdk/__main__.py
```diff
@@ -159,6 +159,9 @@ def extract_code(config, **kwargs):
 @click.option(
     "-o", "--outdir", help="The output dir to write the unified yml to", required=False
 )
+@click.option(
+    "-f", "--force", help="Forcefully overwrites the preexisting yml if one exists", required=False
+)
```
ronykoz: change that to a flag

amshamah419: Good call. Done.

PR #226, opened 2020-02-27
Reputation validations

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21016

## Description
ReputationValidator and unskip reputation-*.json from the validate flow.

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/reputation.py
+++ b/demisto_sdk/commands/common/hook_validations/reputation.py
@@ -11,13 +11,19 @@ class ReputationValidator(BaseValidator):
     def is_valid_file(self, validate_rn=True):
         """Check whether the reputation file is valid or not
         """
-        return all(
-            [
-                self.is_valid_version(),
-                self.is_valid_expiration(),
+        is_reputation_valid = all([
+            self.is_valid_version(),
+            self.is_valid_expiration()
+        ])
+
+        # check only on added files
+        if not self.old_file:
+            is_reputation_valid = all([
+                is_reputation_valid,
                 self.is_id_equals_name()
-            ]
-        )
+            ])
+
+        return is_reputation_valid
 
     def is_valid_version(self):
         # type: () -> bool
--- a/demisto_sdk/commands/common/hook_validations/reputation.py
+++ b/demisto_sdk/commands/common/hook_validations/reputation.py
@@ -50,17 +50,15 @@ class ReputationValidator(BaseValidator):
 
     def is_id_equals_name(self):
         # type: () -> bool
-        """Validate that the expiration field of a 5.5 reputation file is numeric."""
+        """Validate that the id equal name."""
         error_msg = None
         is_valid = True
 
-        from_version = self.current_file.get("fromVersion", "0.0.0")
-        if LooseVersion(from_version) >= LooseVersion("5.5.0"):
-            id_ = self.current_file.get('id', None)
-            name = self.current_file.get('name', None)
-            if not id_ or not name or id_ != name:
-                error_msg = f'{self.file_path}: id and name field are not equal.'
-                is_valid = False
+        id_ = self.current_file.get('id', None)
+        name = self.current_file.get('name', None)
+        if not id_ or not name or id_ != name:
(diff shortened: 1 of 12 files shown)


demisto_sdk/commands/common/hook_validations/reputation.py
```diff
@@ -1,19 +1,67 @@
...
+
+    def is_valid_expiration(self):
+        # type: () -> bool
+        """Validate that the expiration field of a 5.5 reputation file is numeric."""
+        error_msg = None
+        is_valid = True
+
+        from_version = self.current_file.get("fromVersion", "0.0.0")
+        if LooseVersion(from_version) >= LooseVersion("5.5.0"):
+            expiration = self.current_file.get('expiration', "")
```
yaakovi: if expiration doesn't exist it will fail the validation, is that what we want? if not, change default value to a number.

demisto_sdk/commands/common/hook_validations/reputation.py
```diff
@@ -1,19 +1,67 @@
...
+                error_msg = f'{self.file_path}: expiration field should have a numeric value.'
+                is_valid = False
+
+        if error_msg:
+            print_error(error_msg)
+        return is_valid
+
+    def is_id_equals_name(self):
+        # type: () -> bool
+        """Validate that the expiration field of a 5.5 reputation file is numeric."""
```
yaakovi: update docs

demisto_sdk/commands/common/hook_validations/reputation.py
```diff
@@ -1,19 +1,67 @@
...
+        error_msg = None
+        is_valid = True
+
+        from_version = self.current_file.get("fromVersion", "0.0.0")
+        if LooseVersion(from_version) >= LooseVersion("5.5.0"):
```
yaakovi: why is it matter what version we are on? why not always enforcing it for new types?

demisto_sdk/commands/validate/file_validator.py
```diff
@@ -339,6 +340,11 @@ def validate_modified_files(self, modified_files):  # noqa: C901
                 if self.is_backward_check and not incident_field_validator.is_backward_compatible():
                     self._is_valid = False
 
+            elif checked_type(file_path, [REPUTATION_REGEX]):
```
yaakovi: should also include `Packs` path option?

demisto_sdk/commands/common/tests/reputation_test.py
```diff
@@ -0,0 +1,47 @@
...
+    @pytest.mark.parametrize('expiration, is_valid', data_is_valid_expiration)
+    def test_is_valid_expiration(self, expiration, is_valid):
+        structure = StructureValidator("")
+        structure.current_file = {"fromVersion": "5.5.0", "expiration": expiration}
+        validator = ReputationValidator(structure)
+        assert validator.is_valid_expiration() == is_valid, f'is_valid_expiration({expiration})' \
+                                                            f' returns {not is_valid}.'
+
+    data_is_id_equals_name = [
+        # ("CIDR", "CIDR", True),
```
yaakovi: why is this disabled?

demisto_sdk/commands/common/tests/reputation_test.py
```diff
@@ -0,0 +1,47 @@
+import pytest
+
+from demisto_sdk.commands.common.hook_validations.reputation import ReputationValidator
+from demisto_sdk.commands.common.hook_validations.structure import StructureValidator
+
+
+class TestReputationValidator:
```
yaakovi: there is no meaning for the class, usually test classes are use to gather different use cases of same area (for example, multiple tests per function: sanity, special params, exceptions...)

demisto_sdk/commands/common/hook_validations/reputation.py
bakatzir: After 5.5, having expiration is mandatory.

demisto_sdk/commands/validate/file_validator.py
bakatzir: No. indicators are only in Misc repo

PR #229, opened 2020-02-29
Read me md xcheck

orhovy:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/21123

## Description
added new hook for validating readme.md files


demisto_sdk/commands/common/hook_validations/readme.py
```diff
@@ -7,27 +7,34 @@ class ReadMeValidator:
...
-        except Exception as err:
-            print_warning(f'There is no node installed on the machine, Test Skipped, error {err}')
-            return None
+                # if modules are not installed, the readme file would not be validated
+                print_warning(f'There are some modules missing, Test Skipped,\n'
+                              f'Status of node {is_node.returncode}\n '
+                              f'Status of mdx package {is_mdx.returncode}\n status of fs_extra {is_fs_extra.returncode}'
+                              f'\n '
+                              f'Status of commander {is_commander.returncode}')
+        except Exception as err:  # check if you can catch a more exact exception
```
bakatzir:
```suggestion
        except Exception as err:
```

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 ### 0.3.10
 * Added a BC check for incident fields - changing from version is not allowed.
 * Fixed an issue in create-content-artifacts where scripts in Packs in TestPlaybooks dir were copied with a wrong prefix.
+* Added a new hook for validating README.md files
```
bakatzir: we are in 4.0.0, pull from master

demisto_sdk/commands/common/constants.py
```diff
@@ -240,6 +240,8 @@ def invalid_v2_script_name(file_path):
 CAN_START_WITH_DOT_SLASH = '(?:./)?'
 NOT_TEST = '(?!Test)'
 INTEGRATIONS_DIR = 'Integrations'
+README = 'README'
+CHANGELOG = 'CHANGELOG'
```
bakatzir: where is this used?

demisto_sdk/commands/common/hook_validations/readme.py
```diff
@@ -0,0 +1,48 @@
...
+        if self.ready:
+            # run the java script mdx parse validator
+            res = subprocess.run(['node', mdx_parse, '-f', self.file_path], text=True, timeout=10,
+                                 capture_output=True)
+            if res.returncode != 0:
+                print_error(f'Failed verifying README.md, Path: {self.file_path}. Error Message is: {res.stderr}')
+                self.is_valid = False
+        return self.is_valid
+
+    def are_modules_installed_for_verify(self):
```
yuvalbenshalom: we should consider printing the stdout also, if any of dose runs will not return 0.
the check should be after every run.

demisto_sdk/commands/common/hook_validations/readme.py
```diff
@@ -0,0 +1,48 @@
+import subprocess
+import os
+from demisto_sdk.commands.common.tools import print_error, print_warning
+
+
+class ReadMeValidator:
+    def __init__(self, file_path):
+        self.file_path = file_path
```
yuvalbenshalom: @bakatzir lets check if we can avoid this field.

demisto_sdk/commands/common/hook_validations/readme.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        is_readme_valid = all([
+            self.is_mdx_file(),
+        ])
+        return is_readme_valid
+
+    def is_mdx_file(self) -> bool:
+        mdx_parse = f'{os.path.dirname(os.path.abspath(__file__))}/../mdx-parse.js'
+        self.are_modules_installed_for_verify()
+        if self.ready:
```
yuvalbenshalom: I dont see a use for this field. from what I understand this can be the return value of `are_modules_installed_for_verify()`

bakatzir: Indeed this function should be a return value and not a field

demisto_sdk/commands/common/hook_validations/readme.py
```diff
@@ -0,0 +1,48 @@
+import subprocess
+import os
+from demisto_sdk.commands.common.tools import print_error, print_warning
+
+
+class ReadMeValidator:
+    def __init__(self, file_path):
+        self.file_path = file_path
+        self.is_valid = True
+        self.ready = False
```
bakatzir: lets remove these 2 as well

demisto_sdk/commands/common/hook_validations/readme.py
orhovy: There is no schema for md files in the sdk , as a result this class cant inherit from base_validator and this field can not be avoided.

demisto_sdk/commands/common/hook_validations/readme.py
orhovy: removed

bakatzir:
@orhovy what are we lacking to merge this?
we said that the node-modules will nor reside in the sdk - please remove them.
also, the env_Var should start with `DEMISTO_` prefix, e.g: DEMISTO_API
please rename and add proper documentation for it, take for example: 
https://github.com/demisto/demisto-sdk/blob/master/demisto_sdk/commands/run_playbook/run_playbook_command.md

demisto_sdk/commands/common/hook_validations/readme.py
```diff
@@ -0,0 +1,76 @@
+import subprocess
+import os
+from demisto_sdk.commands.common.tools import print_error, print_warning
+
+NODE_MODULES_DIRECTORY = os.path.dirname(os.path.dirname(
```
bakatzir: if used only in `are_modules_installed_for_verify` , why do we need it as global? please move to the function

demisto_sdk/commands/common/hook_validations/readme.py
```diff
@@ -0,0 +1,76 @@
...
+        ready = False
+        try:
+            # check if requiring modules in node exist
+            is_node = subprocess.run(['node', '-v'], text=True, timeout=10, capture_output=True,
+                                     cwd=NODE_MODULES_DIRECTORY)
+            is_mdx = subprocess.run(['npm', 'ls', '@mdx-js/mdx'], text=True, timeout=10, capture_output=True,
+                                    cwd=NODE_MODULES_DIRECTORY)
+            is_fs_extra = subprocess.run(['npm', 'ls', 'fs-extra'], text=True, timeout=10, capture_output=True,
+                                         cwd=NODE_MODULES_DIRECTORY)
+            is_commander = subprocess.run(['npm', 'ls', 'commander'], text=True, timeout=10, capture_output=True,
```
bakatzir: good that you added the timeout

demisto_sdk/commands/common/hook_validations/readme.py
```diff
@@ -0,0 +1,76 @@
...
+        if os.environ.get('DEMISTO_README_VALIDATION'):
+            is_readme_valid = all([
+                self.is_mdx_file(),
+            ])
+            return is_readme_valid
+        else:
+            return True
+
+    def is_mdx_file(self) -> bool:
+        mdx_parse = f'{os.path.dirname(os.path.abspath(__file__))}/../mdx-parse.js'
```
bakatzir: if used aonly in the `if ready:` block, lets move this line inside,

bakatzir:
@orhovy Let's post a message in content-team channel regarding the env perquisites for this validator

PR #230, opened 2020-03-01
V2misleadingfix

orhovy:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22450

## Description
Changed the error message for v2 validation hook


Itay4:
dont we have test for this?

bakatzir:
@Itay4 We have tests. changes occur only in the printing of the error message(which we do not test)

PR #231, opened 2020-03-01
Remove invalid `display` field from init script

amshamah419:
## Status
Ready

## Related Issues
fixes: demisto/etc#22459

## Description
Removed the `display` field from init script.


amshamah419:
Be careful with this review. It's a bit of a beast.

PR #234, opened 2020-03-01
added beta rn collector

jochman:
## Status
Ready

Added beta rn collector


bakatzir:
reference: https://github.com/demisto/etc/issues/20431

PR #236, opened 2020-03-03
disabled pykwalify info log

orhovy:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22478

## Description
Disabled pykwalify info prints


orhovy:
image
after change

demisto_sdk/commands/common/hook_validations/structure.py
```diff
@@ -93,9 +93,11 @@ def is_valid_scheme(self):
         Returns:
             bool. Whether the scheme is valid on self.file_path.
         """
+        INFO = 20
         if self.scheme_name in [None, 'image']:
             return True
         try:
+            logging.disable(INFO)
```
Itay4: use `logging.INFO`

orhovy: changed it

demisto_sdk/commands/common/hook_validations/structure.py
```diff
@@ -96,6 +96,7 @@ def is_valid_scheme(self):
         if self.scheme_name in [None, 'image']:
             return True
         try:
+            logging.disable(logging.INFO)
```
Itay4: please add a comment what are we doing here

demisto_sdk/commands/common/hook_validations/structure.py
```diff
@@ -8,7 +8,7 @@
 from typing import Optional
 
 import yaml
-from pykwalify.core import Core
+from pykwalify.core import Core, logging
```
Itay4: this might cause some issues in the future if we would like to use `logging`
lets just `import logging` (not from `pykwalify`) and capture the required logger using `logging.getLogger`

demisto_sdk/commands/common/hook_validations/structure.py
orhovy: added

demisto_sdk/commands/common/hook_validations/structure.py
orhovy: changed it

demisto_sdk/commands/common/hook_validations/structure.py
```diff
@@ -96,6 +98,9 @@ def is_valid_scheme(self):
         if self.scheme_name in [None, 'image']:
             return True
         try:
+            logging.getLogger(__name__)
```
Itay4: i think you're catching the wrong module name here, should be something like:
```
log = logging.getLogger('pykwalify.core')
log.disable(logging.INFO)
```

demisto_sdk/commands/common/tools.py
```diff
@@ -550,3 +551,17 @@ def find_type(path: str):
...
+    disable logs from channel name starting from equal and bigger than level
+
+    Arguments:
+        name -  channel name
+        level - enum for a specific level which represent the max level from which logs will be printed.
+                e.g. :Logger.ERROR = 40 - all logs that are ERROR level an grater will be printed , all lower will not
+    """
+    log = logging.getLogger(name)
+    log.level = level
+    log.manager._clear_cache()
```
Itay4: this is a bit of overkill, why did we create a function for that?

orhovy: I added name and level and made helper function ,so we will be able to disable any logger and any messages in other parts of the code if needed.

Itay4: if i understand correctly, the cache clearing is required just because we're setting minimum log level
if thats the case, i dont like the fact we're touching an internal mechanism just for that
i prefer we do it like you did in the first place, where you disable log outputs only for `pykwalify`, dont think we really need this generalization

orhovy: The problem with what I did before is that logger.disable, disabled all logger's logs pykwalify and others. What you suggested could work but the problem is that there is no disable function by levels to log = logger.getLogger() and as a result I had to set level and clean cache

Itay4: did you try `.setLevel(logging.WARNING)` ?

orhovy: SetLevel did exactly what I did in the helper function. So I changed it.

Itay4: not sure i follow.. if it does what you want, why didnt you use it in the first place and created a function?
is the code currently leads to the expected results?

orhovy: Yes the code leads to expected results, I did not found the SetLevel function and as a result I created one that does the samething

demisto_sdk/commands/common/tools.py
```diff
@@ -27,7 +27,7 @@ class LOG_COLORS:
     YELLOW = '\033[0;33m'
 
 
-def get_yml_paths_in_dir(project_dir: str, error_msg: str,) -> Tuple[list, str]:
+def get_yml_paths_in_dir(project_dir: str, error_msg: str, ) -> Tuple[list, str]:
```
Itay4: whats that change for ?

PR #237, opened 2020-03-03
increase line width limit in scripts

yaakovi:
## Status
Ready

## Description
some scripts / integrations contains very long lines (like `WhoIs` below).
this is causing the sdk to split those lines in the unified yaml creating SyntaxErrors in code.

## Screenshots
image (7)

### DIFF
--- a/demisto_sdk/commands/unify/unifier.py
+++ b/demisto_sdk/commands/unify/unifier.py
@@ -51,7 +51,7 @@ class Unifier:
 
         self.ryaml = YAML()
         self.ryaml.preserve_quotes = True
-        self.ryaml.width = 3000  # make sure long lines will not break (relevant for code section)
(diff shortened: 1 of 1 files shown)


yaakovi:
what you expect the test will look like (i have one pending)

bakatzir:
@yaakovi Next time, it might be a good opportunity to add it to the `content_creator_test`

bakatzir:
something like this maybe:
Screen Shot 2020-03-03 at 21 35 41
```
content_creator.copy_dir_files(f'{self.integrations_full_path}/Securonix', content_creator.content_bundle)
        print(f'{self._bundle_dir}/Panorama.yml')
        assert filecmp.cmp(f'{self.integrations_full_path}/Securonix/Securonix_unified.yml',
                           f'{self._bundle_dir}/Securonix_unified.yml')
```

PR #240, opened 2020-03-05
added is_root_connected_to_all_tasks with tests

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22570

## Description
added `is_root_connected_to_all_tasks` with tests to PlaybookValidator, which validates that the root is connected to all tasks in the playbook.


demisto_sdk/commands/common/hook_validations/playbook.py
```diff
@@ -62,3 +64,21 @@ def is_no_rolename(self):  # type: () -> bool
...
+        tasks = self.current_file.get('tasks', {})
+        tasks_bucket = set()
+        next_tasks_bucket = set()
+        for task_id, task in tasks.items():
+            if task_id != start_task_id:
+                tasks_bucket.add(task_id)
+            next_tasks = task.get('nexttasks', {})
+            for next_task_ids in next_tasks.values():
+                next_tasks_bucket.update(next_task_ids)
+        return tasks_bucket.issubset(next_tasks_bucket)
```
yuvalbenshalom: I think we should add a warning print here. maybe even consider printing the list of task names that ar unreachable from the root.

PR #241, opened 2020-03-05
Dashboard validator

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22360


demisto_sdk/commands/common/hook_validations/dashboard.py
```diff
@@ -1,11 +1,108 @@
...
+    def is_id_equals_name(self):
+        # type: () -> bool
+        """Check whether the dashboard ID is equal to its name.
+
+        Returns:
+            bool. Whether the file id equals to its name
+        """
+        return super(DashboardValidator, self)._is_id_equals_name('dashboard')
+
+    def is_not_excluding_fields(self):
```
yuvalbenshalom: maybe change to "contains_forbidden_fields" or something like that?
"is_not_excluding" is confusing.

demisto_sdk/commands/common/hook_validations/dashboard.py
```diff
@@ -1,11 +1,108 @@
...
+        # type: () -> bool
+        """Return if root and widgets exclude the unnecessary fields.
+
+        Returns:
+            True if exclude, else False.
+        """
+        error_msg = ""
+        is_valid = True
+        fields_to_exclude = ['system', 'isCommon', 'shared', 'owner',
```
yuvalbenshalom: lets re check if we want to have a list of excluded fields or rather have a list of permitted fields.

PR #242, opened 2020-03-08
add incident type validator

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/20084


demisto_sdk/commands/common/hook_validations/incident_type.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        return super(IncidentTypeValidator, self)._is_valid_version()
+
+    def is_current_valid_from_version(self):
+        # type: () -> bool
+        """Check if the current file fromversion is valid.
+        Returns:
+            bool. Whether the current fromversion is valid or not.
+        """
+        error_msg = None
```
yuvalbenshalom: why do you need this var? isn't it better to just print any error that you find?

demisto_sdk/commands/common/hook_validations/incident_type.py
```diff
@@ -0,0 +1,127 @@
...
+                is_bc_broke = True
+        return is_bc_broke
+
+    def is_including_int_fields(self):
+        # type: () -> bool
+        """Check if including required fields, only from 5.0.0.
+        Returns:
+            bool. Whether the fields .
+        """
+        error_msg = ""
```
yuvalbenshalom: here again, not sure if we need this var.

bakatzir:
@yuvalbenshalom I can, but this is not the convention of the hook_validations that we have. and I find it readable.

yuvalbenshalom:
@bakatzir I don't understand how it is more reliable to put an error you found in a variable, in hope that it will be printed later, rather than print it instantly. what are the pros for this?

bakatzir:
@yuvalbenshalom Agreed.

PR #243, opened 2020-03-08
Support CommonServerPython being in 2 possible directories

liorblob:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/22548

## Description
Support the CommonServerPython file being in 2 possible directories.
We should delete the script directory check in the future.


liorblob:
@bakatzir The linter only works on packages, it doesn't matter if it's in a pack or not.

PR #244, opened 2020-03-08
Add a warning message when not running the latest sdk version

DeanArbel:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22587

## Description
Added a warning message when not running on the latest version.

## Screenshots
image

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -197,8 +197,11 @@ def get_last_remote_release_version():
             if isinstance(latest_release, str):
                 # remove v prefix
                 return latest_release[1:]
-    except Exception:
-        print_warning('Could not get latest demisto-sdk version.')
+    except Exception as exc:
+        exc_msg = str(exc)
+        if isinstance(exc, requests.exceptions.ConnectionError):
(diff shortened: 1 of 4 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -182,17 +182,24 @@ def get_child_files(directory):
...
+        releases_request = requests.get(SDK_API_GIHUB_RELEASES, verify=False)
+        releases_request.raise_for_status()
+        releases = requests.get(SDK_API_GIHUB_RELEASES, verify=False).json()
+        if isinstance(releases, list) and isinstance(releases[0], dict):
+            latest_release = releases[0].get('tag_name')
+            if isinstance(latest_release, str):
+                # remove v prefix
+                return latest_release[1:]
+    except Exception:
+        print_warning('Could not get latest demisto-sdk version.')
```
yuvalbenshalom: I think you should print in the second line the error that you get in the exception, and especially check it when you are disconnected from the internet. make sure the error is readable and add extra handling for this case if it is not.

DeanArbel: image

PR #249, opened 2020-03-10
Fix lint manager

liorblob:


demisto_sdk/commands/lint/lint_manager.py
```diff
@@ -154,7 +154,7 @@ def run_parallel_packages(self, pkgs_to_run: List[str]) -> int:
         # when we modify the file for mypy includes
         if 'Scripts/CommonServerPython' in pkgs_to_run:
             pkgs_to_run.remove('Scripts/CommonServerPython')
-            common_server_path = get_common_server_path(self.configuration.env_dir)
+            common_server_path = get_common_server_dir('')
```
yuvalbenshalom: why did you remove the `env_dir`?

liorblob: so it will be like the previous version

liorblob: previously it was just `Scripts/CommonServerPython`

bakatzir:
i'm restoring the branch to be able to test the content build with it

PR #251, opened 2020-03-10
fix non parallel run of common

glicht:
## Status
Ready

## Description
Fix bug in non-parallel run when common is in a base pack. Add utility functions for pwsh common.

## Screenshots

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -568,7 +568,7 @@ def get_common_server_path(env_dir):
 
 
 def get_common_server_path_pwsh(env_dir):
-    common_server_dir = get_common_server_dir(env_dir)
+    common_server_dir = get_common_server_dir_pwsh(env_dir)
     return os.path.join(common_server_dir, 'CommonServerPowerShell.py')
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -564,14 +564,25 @@ def find_type(path: str):
...
-
     return os.path.join(common_server_dir, 'CommonServerPython.py')
 
 
-def get_common_server_dir(env_dir):
-    common_server_pack_path = os.path.join(env_dir, 'Packs', 'Base', 'Scripts', 'CommonServerPython')
-    common_server_script_path = os.path.join(env_dir, 'Scripts', 'CommonServerPython')
+def get_common_server_path_pwsh(env_dir):
+    common_server_dir = get_common_server_dir_pwsh(env_dir)
+    return os.path.join(common_server_dir, 'CommonServerPowerShell.py')
```
yuvalbenshalom:
```suggestion
    return os.path.join(common_server_dir, 'CommonServerPowerShell.ps1')
```
I think this should be the correct one.

PR #263, opened 2020-03-16
split-yml: Removed the yml_type argument

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22757

## Description
code and unit tests refactor as well.

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -79,19 +79,11 @@ def main(config, version, env_dir):
     '-h', '--help'
 )
 @click.option(
-    '--infile', '-i',
-    help='The yml file to extract from',
-    required=True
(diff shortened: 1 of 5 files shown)


bakatzir:
relates to: https://github.com/demisto/etc/issues/22758

demisto_sdk/__main__.py
```diff
@@ -79,20 +79,12 @@ def main(config, version, env_dir):
     '-h', '--help'
 )
 @click.option(
-    '--infile', '-i',
-    help='The yml file to extract from',
-    required=True
+    '-i', '--input', help='The yml file to extract from', required=True
```
anara123: I think we agreed on `in` and `out`

bakatzir: `in` is a saved word, so it is not possible.

PR #264, opened 2020-03-16
Pack metadata validations

liorblob:
## Status
In Progress

## Related Issues
fixes: https://github.com/demisto/etc/issues/22433

This is what I added so far. What else can I add?
I'll add a test for that when it's done.


demisto_sdk/commands/common/constants.py
```diff
@@ -496,6 +496,30 @@ def invalid_v2_script_name(file_path):
...
+PACK_METADATA_DESC = 'description'
+PACK_METADATA_MIN_VERSION = 'serverMinVersion'
+PACK_METADATA_CURR_VERSION = 'currentVersion'
+PACK_METADATA_AUTHOR = 'author'
+PACK_METADATA_URL = 'url'
+PACK_METADATA_EMAIL = 'email'
+PACK_METADATA_CATEGORIES = 'categories'
+PACK_METADATA_TAGS = 'tags'
+PACK_METADATA_CREATED = 'created'
+PACK_METADATA_UPDATED = 'updated'
```
IkaDemisto: I think we can remove this field and remove it also in init pack_metadata

demisto_sdk/commands/common/constants.py
```diff
@@ -496,6 +496,30 @@ def invalid_v2_script_name(file_path):
...
 
+PACK_METADATA_NAME = 'name'
+PACK_METADATA_DESC = 'description'
+PACK_METADATA_MIN_VERSION = 'serverMinVersion'
+PACK_METADATA_CURR_VERSION = 'currentVersion'
+PACK_METADATA_AUTHOR = 'author'
+PACK_METADATA_URL = 'url'
+PACK_METADATA_EMAIL = 'email'
+PACK_METADATA_CATEGORIES = 'categories'
+PACK_METADATA_TAGS = 'tags'
```
IkaDemisto: Also maybe should validate that tags and all keys that have list with values, aren't empty. Like we had with empty string as tags

yuvalbenshalom: @IkaDemisto Are you sure that all of the list fields will be populated in a valid Packs?
what are the list fields that we know right now how to populate in every pack that exist?

liorblob: I added a validation for that, if those lists have 1 item and it's empty it will fail

IkaDemisto: @yuvalbenshalom I meant following cases : ['tags'] =[""]
@liorblob added validation for it.

PR #266, opened 2020-03-17
Release notes validator will now run on: incident fields, indicator f…

bakatzir:
…ields, incident types, dashboard and reputations.

## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22689

## Description
minor changes in the file_validator where also needed to make sure we run the rn checks on necessary files.

## Screenshots
in the referenced issue

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/dashboard.py
+++ b/demisto_sdk/commands/common/hook_validations/dashboard.py
@@ -20,7 +20,7 @@ class DashboardValidator(BaseValidator):
             bool. Whether the dashboard is valid or not
         """
         is_dashboard_valid = [
-            super(DashboardValidator, self).is_valid_file(validate_rn),
+            super().is_valid_file(validate_rn),
             self.is_valid_version()
         ]
 
--- a/demisto_sdk/commands/common/hook_validations/dashboard.py
+++ b/demisto_sdk/commands/common/hook_validations/dashboard.py
@@ -12,7 +12,7 @@ class DashboardValidator(BaseValidator):
             widgets = [item.get('widget') for item in layout_of_dashboard]
         return widgets
 
-    def is_valid_dashboard(self):
+    def is_valid_dashboard(self, validate_rn=True):
         # type: () -> bool
         """Check whether the dashboard is valid or not.
 
@@ -20,6 +20,7 @@ class DashboardValidator(BaseValidator):
             bool. Whether the dashboard is valid or not
         """
         is_dashboard_valid = [
+            super(DashboardValidator, self).is_valid_file(validate_rn),
             self.is_valid_version()
         ]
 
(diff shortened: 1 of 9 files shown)


demisto_sdk/commands/common/hook_validations/dashboard.py
```diff
@@ -12,14 +12,15 @@ def get_widgets_from_dashboard(dashboard):
...
-    def is_valid_dashboard(self):
+    def is_valid_dashboard(self, validate_rn=True):
         # type: () -> bool
         """Check whether the dashboard is valid or not.
 
         Returns:
             bool. Whether the dashboard is valid or not
         """
         is_dashboard_valid = [
+            super(DashboardValidator, self).is_valid_file(validate_rn),
```
yaakovi: in python3 you don't need those args
```suggestion
            super().is_valid_file(validate_rn),
```

demisto_sdk/commands/validate/file_validator.py
```diff
@@ -366,7 +366,7 @@ def validate_modified_files(self, modified_files):  # noqa: C901
 
             elif checked_type(file_path, JSON_ALL_INCIDENT_TYPES_REGEXES):
                 incident_type_validator = IncidentTypeValidator(structure_validator)
-                if not incident_type_validator.is_valid_incident_type():
+                if not incident_type_validator.is_valid_incident_type(validate_rn=True):
```
yaakovi: why is that one different? why not `not file_type`?

bakatzir: in `modified_files` it should always be true
in `added_files` it is `not file_type`

demisto_sdk/commands/common/hook_validations/dashboard.py
bakatzir: @yaakovi seems like it is needed.
```
        return all(
            [
>               super.is_valid_file(validate_rn),
                self.is_valid_name(),
                self.is_valid_type(),
                self.is_valid_group(),
                self.is_valid_content_flag(),
                self.is_valid_system_flag(),
                self.is_valid_cliname(),
                self.is_valid_version(),
                self.is_current_valid_from_version(),
                self.is_valid_required()
            ]
        )
E       AttributeError: type object 'super' has no attribute 'is_valid_file'

demisto_sdk/commands/common/hook_validations/incident_field.py:182: AttributeError
```

bakatzir: same error occurs in the `dashboard.py` upon change.

yaakovi: `super()` not `super`

demisto_sdk/commands/validate/file_validator.py
yaakovi: right, my bad. sorry.

bakatzir:
@yaakovi correct regarding the super(). changed it in all the validators.

PR #269, opened 2020-03-17
Verify there are no hidden integration parameters

gal-berger:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/22814

## Description
Verify there are no hidden integration parameters


demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -562,3 +563,18 @@ def is_valid_display_name(self):
...
             return True
+
+    def is_all_params_not_hidden(self):
+        # type: () -> bool
+        """
+        Verify there are no hidden integration parameters.
+        Returns:
+            bool. True if there aren't hidden parameters False otherwise.
+        """
+        ans = True
```
yuvalbenshalom: there is no need in the var(know it is common in this code)
just return False if you found hidden and True if you got to the end.

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -562,3 +563,18 @@ def is_valid_display_name(self):
                 print_error(Errors.invalid_v2_integration_name(self.file_path))
                 return False
             return True
+
+    def is_all_params_not_hidden(self):
```
Itay4:
```suggestion
    def is_all_params_not_hidden(self):
```
```suggestion
    def is_all_params_not_hidden(self) -> bool:
```

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -562,3 +563,18 @@ def is_valid_display_name(self):
...
+        # type: () -> bool
```
Itay4: remove this line

demisto_sdk/commands/common/hook_validations/integration.py
gal-berger: I did it that way because I wanted to finish printing regarding all hidden parameters before exiting the function

demisto_sdk/commands/common/hook_validations/integration.py
gal-berger: done

demisto_sdk/commands/common/hook_validations/integration.py
gal-berger: done

gal-berger:
Screen Shot 2020-03-19 at 12 20 12

PR #252, opened 2020-03-10
Enhancement format command

orhovy:
## Status
~~In Hold ,Reason:Waiting for incidenttypes validator: https://github.com/demisto/demisto-sdk/pull/242/~~
WIP

## Related Issues
fixes: https://github.com/demisto/etc/issues/22489
fixes: https://github.com/demisto/etc/issues/22454
fixes: https://github.com/demisto/etc/issues/21573
~~also fixes: https://github.com/demisto/etc/issues/22816~~ @bakatzir 
## Description
1. Add new format options for :
- incidentfields
- incidenttypes
- indicator fields
- indicator types
- layouts
- dashboards
for each file , remove fields that are not set in schema of this type of file.
2. fixed tags of --source_file -> --input and --output_file -> --output
3. Using --input , if file, format will be activated on it, if directory (Pack or other) , will go through all files in directory and format all yml, json files , Using a generic function.
4. Removed flag --type-file , format command will decide on the type using generic function.
5. New --from-version flag, will set fromversion field to givven , if None,
 On modified files will set to 1.0.0 , On Added files will set to 5.0.0

PR #273, opened 2020-03-18
Fixed equal value

orhovy:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22349

## Description
replaced the string "simple: = " with "simple: '='"

depends on: https://github.com/demisto/demisto-sdk/pull/252


demisto_sdk/commands/format/tests/formatting_test.py
```diff
@@ -100,3 +104,23 @@ def test_playbook_sourceplaybookid(source_path):
...
     assert 'sourceplaybookid' not in base_yml.yml_data
+
+
+EQUAL_TEST = [
+    (EQUAL_VAL_FORMAT_PLAYBOOK_SOURCE, EQUAL_VAL_FORMAT_PLAYBOOK_DESTINATION, EQUAL_VAL_PATH, 'playbook'),
+]
+
+
+@pytest.mark.parametrize('input, output, path, file_type', EQUAL_TEST)
+def test_eqaul_value_in_file(input, output, path, file_type):
```
bakatzir:
```suggestion
def test_equal_value_in_file(input, output, path, file_type):
```

demisto_sdk/commands/format/update_generic_yml.py
```diff
@@ -67,6 +67,11 @@ def get_yml_data_as_dict(self):
         print(F'Reading YML data')
 
         with open(self.source_file) as f:
+            # yaml file can not load = with out a value after it -> changed to '='
+            read_file = f.read()
+            replaced = read_file.replace("simple: =", "simple: '='")
```
bakatzir: as yml can be also scripts and automations, can we do it only for playbooks?(use the `file_type` func)
@yuvalbenshalom WDYT?

PR #276, opened 2020-03-19
Generate docs backup

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22704

## Description
* Added the use_cases, command_permissions, limitations arguments to the generate-docs command.
* Removed the -t argument from the generate-docs command, the file type will be inferred.
* The -o argument in the generate-docs command is no longer mandatory, default value will be the input file directory.

## Screenshots
in related issue


demisto_sdk/commands/run_cmd/runner.py
```diff
@@ -31,6 +31,9 @@ def __init__(self, query: str, insecure: bool = False, debug: str = None, debug_
         if self.debug:
             self.query += ' debug-mode="true"'
 
+    def set_insecure(self, insecure):
```
yaakovi: why not pass it in init?

bakatzir: since it is a demisto_client object, so it will require more imports and internals changes in the `Runner` object that will occur in `common.py`. I though adding a set function is a cleaner approach. but - whatever you'll say.

demisto_sdk/commands/generate_docs/common.py
```diff
@@ -146,13 +166,15 @@ def stringEscapeMD(st, minimal_escaping=False, escape_multiline=False):
...
-def execute_command(command_example):
+def execute_command(command_example, insecure: bool):
     errors = []
     context = {}
     md_example = ''
     cmd = command_example
     try:
         runner = Runner('')
+        if insecure:
+            runner.set_insecure(insecure)
```
yaakovi: this is what i meant. no need for the set function.
```suggestion
        runner = Runner('', insecure=insecure)
```

init function of runner:
image

PR #279, opened 2020-03-19
Ignore Api Modules pack in metadata validation

liorblob:
## Status
Ready


demisto_sdk/commands/common/hook_validations/pack_unique_files.py
```diff
@@ -173,6 +173,9 @@ def validate_pack_unique_files(self):
         """Main Execution Method"""
         self.validate_secrets_file()
         self.validate_pack_ignore_file()
-        self.validate_pack_meta_file()
         self.validate_readme_file()
+        # We don't want to check the metadata file for this pack
+        if 'ApiModules' not in self.pack:
```
Itay4: lets use a const here

PR #280, opened 2020-03-19
add env var to allow skip version check

glicht:
## Status
Ready

## Description
When working off-line or with a slow internet the version check can really slow things down for simple tasks. Added an env var `DEMISTO_SDK_SKIP_VERSION_CHECK` if it is set then version check is skipped. Additionally if `CI` is set then we skip. 

## Screenshots


bakatzir:
@DeanArbel FYI

PR #290, opened 2020-03-23
general section was not being generated correctly

bakatzir:
## Status
Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22989

### DIFF
--- a/demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
+++ b/demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
@@ -193,20 +193,6 @@ def test_get_used_in():
 # integration tests
 
 
-def test_generate_integration_general_section():
-    docs = []  # type: list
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
```diff
@@ -193,6 +193,20 @@ def test_get_used_in():
 # integration tests
 
 
+def test_generate_integration_general_section():
```
yaakovi: this doesn't test the function (you are not even calling it...)

bakatzir: because there is not a separate function for the general part.

bakatzir:
@yuvalbenshalom well, the test I wrote does not use a function from the file so it is a bit redundant.
and to create a function for this one-liner in the `generate_integration_doc.py` seems a bit forced.

PR #296, opened 2020-03-24
[Fix] Added case insensitivity to playbook condition validation

DeanArbel:
## Status
Ready

## Description
Validation of condition branches compares the declared vs. used branches using a case sensitive comparison. After reviewing this with @idovandijk it was decided this behavior is not the intended validation, and as such, it was removed.

## Additional Info
Validated with playbook from https://github.com/demisto/content/pull/6132

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/playbook.py
+++ b/demisto_sdk/commands/common/hook_validations/playbook.py
@@ -103,22 +103,26 @@ class PlaybookValidator(BaseValidator):
             bool. if the task handles all condition branches correctly.
         """
         is_all_condition_branches_handled: bool = True
+        # ADD all possible conditions to task_condition_labels (UPPER)
         # #default# condition should always exist in a builtin condition
-        task_condition_labels = {'#default#'}
+        task_condition_labels = {'#DEFAULT#'}
         for condition in task.get('conditions', []):
             label = condition.get('label')
             if label:
-                task_condition_labels.add(label)
+                task_condition_labels.add(label.upper())
+
+        # REMOVE all used condition branches from task_condition_labels (UPPWE)
         next_tasks: Dict = task.get('nexttasks', {})
-        for next_task_branch, next_task_ids in next_tasks.items():
+        for next_task_branch in next_tasks.keys():
             try:
-                if next_task_ids:
-                    task_condition_labels.remove(next_task_branch)
+                if next_task_branch:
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/hook_validations/playbook.py
```diff
@@ -103,22 +103,26 @@ def is_builtin_condition_task_branches_handled(self, task: Dict) -> bool:
...
         # #default# condition should always exist in a builtin condition
-        task_condition_labels = {'#default#'}
+        task_condition_labels = {'#DEFAULT#'}
         for condition in task.get('conditions', []):
             label = condition.get('label')
             if label:
-                task_condition_labels.add(label)
+                task_condition_labels.add(label.upper())
+
+        # REMOVE all used condition branches from task_condition_labels (UPPWE)
```
Itay4:
```suggestion
        # REMOVE all used condition branches from task_condition_labels (UPPER)
```

demisto_sdk/commands/common/hook_validations/playbook.py
```diff
@@ -103,22 +103,26 @@ def is_builtin_condition_task_branches_handled(self, task: Dict) -> bool:
             bool. if the task handles all condition branches correctly.
         """
```
Itay4: add a comment in the docstring why we're applying `upper`

demisto_sdk/commands/common/hook_validations/playbook.py
```diff
@@ -103,22 +103,26 @@ def is_builtin_condition_task_branches_handled(self, task: Dict) -> bool:
...
+                task_condition_labels.add(label.upper())
+
+        # REMOVE all used condition branches from task_condition_labels (UPPWE)
         next_tasks: Dict = task.get('nexttasks', {})
-        for next_task_branch, next_task_ids in next_tasks.items():
+        for next_task_branch in next_tasks.keys():
             try:
-                if next_task_ids:
-                    task_condition_labels.remove(next_task_branch)
+                if next_task_branch:
```
Itay4: why did we change the condition here?

DeanArbel: While going over the code I noticed this bug. it makes no sense to check the `next_task_ids` since they should always have a value, and we're actually removing the `next_task_branch`

PR #301, opened 2020-03-26
Add functions from content

liorblob:
## Status
Ready

## Related Issues
https://github.com/demisto/etc/issues/22938

## Description
As part of the content repo cleanup, we're moving all the constants and tools to the SDK so there will be no inconsistencies.


demisto_sdk/commands/common/tools.py
```diff
@@ -648,3 +648,384 @@ def run_command_os(command: str, cwd: Path, env: dict = os.environ) -> Tuple[str
...
+    :return: tag
+    """
+    tags = run_command('git tag').split('\n')
+    tags = [tag for tag in tags if re.match(r'\d+\.\d+\.\d+', tag) is not None]
+    tags.sort(key=LooseVersion, reverse=True)
+
+    return tags[0]
+
+
+def collect_pack_content_items(pack_path):
```
IkaDemisto: You can remove it, will be moved to store_packs

demisto_sdk/commands/common/tools.py
```diff
@@ -648,3 +648,384 @@ def run_command_os(command: str, cwd: Path, env: dict = os.environ) -> Tuple[str
...
+
+            data[directory] = dir_data
+    except Exception as e:
+        task_status = False
+        print_error("Failed to collect pack content items at path :{}\n. Additional info {}".format(pack_path, e))
+    finally:
+        return task_status, data
+
+
+def input_to_list(input_data):
```
IkaDemisto: Same here

liorblob:
@yuvalbenshalom The only addition here is the `Docker` class, which is difficult to test because it uses SSH. We can choose to skip testing that for now or have a zoom about it with @IkaDemisto .

IkaDemisto:
@liorblob do we need the `Docker` class in demisto-sdk? In is for our internal use, mainly in `test_integration`. Are we planning to move it also to sdk?

liorblob:
@IkaDemisto Nope, should we move the class to that file?

IkaDemisto:
> @IkaDemisto Nope, should we move the class to that file?

Yes, I think so.

PR #303, opened 2020-03-28
Added new command for pack dependencies  

IkaDemisto:
## Status
Ready

## Related Issues
related to: https://github.com/demisto/etc/issues/22129

## Description
Added new `demisto-sdk` command that uses id set and searches for specific pack dependencies. Writes back the found result to `pack_metadata.json`.


IkaDemisto:
Id set file is for unit tests

demisto_sdk/__main__.py
```diff
@@ -589,10 +591,27 @@ def id_set_command(**kwargs):
     id_set_creator.create_id_set()
 
 
+# ====================== find-dependencies ====================== #
+@main.command(name="find-dependencies",
```
bakatzir: add `hidden=True

IkaDemisto: But it is not internal command.

bakatzir: I do not think contributors should run this command,
but that we should add it automatically to the their pack_metada.

IkaDemisto: Hmm it is mainly for contributors, they should define what dependencies are mandatory or not after running this command.
@yuvalbenshalom how do you think we should handle it?

yuvalbenshalom: @IkaDemisto I think @bakatzir is correct. we need a command that dose "update dependencies", and update the metadata.
we can consider to split it. lets talk about it.

IkaDemisto: we cannot add this automatically, when you init pack there are no dependencies

yuvalbenshalom: will create the updating command in a different issue and PR

PR #319, opened 2020-03-31
add fromversion when extracting pwsh

glicht:
## Status
Ready

## Related Issues
fixes: link to the issue

## Description
Add fromversion when dealing with pwsh extraction

## Screenshots


demisto_sdk/commands/split_yml/extractor.py
```diff
@@ -85,6 +85,13 @@ def extract_to_package_format(self) -> int:
             del yaml_obj['image']
             if 'detaileddescription' in yaml_obj:
                 del yaml_obj['detaileddescription']
+        script_obj['script'] = SingleQuotedScalarString('')
+        code_type = script_obj['type']
+        if code_type == TYPE_PWSH and not yaml_obj.get('fromversion'):
```
Itay4: i would also check for `fromVersion`
some time people add it instead of `fromversion` because they see it in other content entities

glicht: If they add it our schema validation will catch should catch it. I think it is ok.

glicht:
@bakatzir also added a validation hook for fromversion when using powershell:

image

glicht:
@bakatzir Good idea. I added format support for pwsh version

PR #320, opened 2020-03-31
Docker image tag validation login change

ohaim1008:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/22765ת https://github.com/demisto/etc/issues/23742

## Description
All docker images in all server versions, in new or modified files, should uphold:
1. Have a tag
2. The tag is not labeled "latest"
3. The tag is the most updated tag exists in docker-hub
4. Not the default one (demisto/python:1.3-alpine)


demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -51,17 +51,10 @@ def is_docker_image_latest_tag(self):
...
-            if self.docker_image_latest_tag != self.docker_image_tag and not \
-                    'demisto/python:1.3-alpine' == '{}:{}'.format(self.docker_image_name, self.docker_image_tag):
-                # If docker image name are different and if the docker image isn't the default one
-                self.is_latest_tag = False
-        # Case of an added file
-        elif not self.is_modified_file:
-            if self.docker_image_latest_tag != self.docker_image_tag:
-                self.is_latest_tag = False
+        if self.docker_image_latest_tag != self.docker_image_tag and not \
+                'demisto/python:1.3-alpine' == '{}:{}'.format(self.docker_image_name, self.docker_image_tag):
```
Itay4, line 55: we should allow using `demisto/python:1.3-alpine`

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -180,10 +174,11 @@ def find_latest_tag_by_date(tags):
         latest_tag_date = datetime.now() - timedelta(days=400000)
         for tag in tags:
             tag_date = datetime.strptime(tag.get('last_updated'), '%Y-%m-%dT%H:%M:%S.%fZ')
-            if tag_date >= latest_tag_date:
+            if tag_date >= latest_tag_date and tag.get('name') != 'latest':
                 latest_tag_date = tag_date
                 latest_tag_name = tag.get('name')
-
+        if latest_tag_name == 'latest':
```
Itay4, line 180: not sure can this condition be true, you're checking in the previous condition for that..

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -180,10 +174,11 @@ def find_latest_tag_by_date(tags):
         latest_tag_date = datetime.now() - timedelta(days=400000)
         for tag in tags:
             tag_date = datetime.strptime(tag.get('last_updated'), '%Y-%m-%dT%H:%M:%S.%fZ')
-            if tag_date >= latest_tag_date:
+            if tag_date >= latest_tag_date and tag.get('name') != 'latest':
```
Itay4, line 177: what happens to a docker image that `latest` is the only tag it got ?

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -169,6 +162,7 @@ def lexical_find_latest_tag(tags):
     @staticmethod
     def find_latest_tag_by_date(tags):
```
Itay4, line 163:
```suggestion
    def find_latest_tag_by_date(tags: list) -> str:
```

demisto_sdk/commands/common/hook_validations/docker.py
ohaim1008: It is allowing there is an "and not' before the statement, I think I'll separate it to 2 statement so it will be more readable

demisto_sdk/commands/common/hook_validations/docker.py
ohaim1008: It's true only in case there is only 1 docker image in the docker-hub and it labelled "latest" (I saw that currently there are couple of those)

demisto_sdk/commands/common/hook_validations/docker.py
ohaim1008: latest_tag_name will stay "latest" and we will get the error:
'The docker image tag is labeled "latest" please create a valid docker image tag'

Itay4:
still missing tests

ohaim1008:
Changed the logic and the tests,
I think there is a need in changing (or removing) the breaking_backwards_docker, currently there is no option to change the docker in the yml and pass the build, so in those cases merge is forced.
Do we want to keep it like that? @Itay4

CHANGELOG.md
```diff
@@ -1,17 +1,14 @@
 # Changelog
-
```
Itay4, line 2: update changelog

demisto_sdk/commands/common/tests/docker_test.py
```diff
@@ -5,7 +5,7 @@
 from demisto_sdk.commands.common.hook_validations.docker import \
     DockerImageValidator
 from demisto_sdk.commands.common.tools import get_yaml
-from mock import patch
+from unittest.mock import patch
```
Itay4, line 8: use pytest mock to patch

demisto_sdk/commands/common/tests/docker_test.py
```diff
@@ -132,52 +132,32 @@ def test_parse_docker_image():
 
 
 # disable-secrets-detection-end
-
-
 def test_is_docker_image_latest_tag():
```
Itay4, line 135: it is a bad practise to have 1 unit test for all cases, lets split this to a few tests which each tests a specific cases so it will be more readable

Itay4:
@ohaim1008 please update pr status in the comment to ready if it is

Itay4:
add to each test docstring according to `Given-When-Then` template

Itay4:
i think we didnt handle the default docker image correctly
in case a modified integration/script use it, they build should fail, as we shouldnt use it anymore

other than that, lets trigger a content build, using this branch as sdk reference, with the following:
1. modify an integration which does not have the most updated tag - check it fails
2. modify an integration which uses default docker image - check it fails
3. modify an integration which has `latest` tag - check it fails
4. modify an integration which has the most updated tag - check it passes

in addition, lets remove the validation of docker image for backward compatibility

ohaim1008:
@Itay4 are you sure? I think a lot of integrations are using the default one and if they won't we will have a lot of similar docker images in the docker hub.
which means a lot of docker PR's... what do you think?

and completely agree that we should remove it from bc.

Itay4:
> @Itay4 are you sure? I think a lot of integrations are using the default one and if they won't we will have a lot of similar docker images in the docker hub.
> which means a lot of docker PR's... what do you think?
> 
> and completely agree that we should remove it from bc.

thats exactly why we need to fail the build on that - we dont want to keep using that docker image, we want to replace it with new/updated one

ohaim1008:
> > @Itay4 are you sure? I think a lot of integrations are using the default one and if they won't we will have a lot of similar docker images in the docker hub.
> > which means a lot of docker PR's... what do you think?
> > and completely agree that we should remove it from bc.
> 
> thats exactly why we need to fail the build on that - we dont want to keep using that docker image, we want to replace it with new/updated one

OK, I'll make the changes and check the build.

ohaim1008:
> i think we didnt handle the default docker image correctly
> in case a modified integration/script use it, they build should fail, as we shouldnt use it anymore
> 
> other than that, lets trigger a content build, using this branch as sdk reference, with the following:
> 
> 1. modify an integration which does not have the most updated tag - check it fails
> 2. modify an integration which uses default docker image - check it fails
> 3. modify an integration which has `latest` tag - check it fails
> 4. modify an integration which has the most updated tag - check it passes
> 
> in addition, lets remove the validation of docker image for backward compatibility

Done, I used this brunch and checked for different integration and it did work as it should have.
also tried to change docker image and saw that it doesn't break bc.

Another change that was necessary is in "parse_docker_image" we gave to scripts or integrations that without docker image the default one, now we use the most updated one "demisto/python:2.7.17.6981"

Itay4:
@ohaim1008 please provide link to the content build

ohaim1008:
@Itay4 
https://circleci.com/gh/demisto/content/50824
The integrations I used:
GenericSQL - uses the latest docker image - valid docker 
Helloworld - uses an old version while a newer is available
Cherwell - using demisto/python:1.3-alpine
ActiveMQ - uses a “latest” docker image

Itay4:
> @Itay4
> https://circleci.com/gh/demisto/content/50824
> The integrations I used:
> GenericSQL - uses the latest docker image - valid docker
> Helloworld - uses an old version while a newer is available
> Cherwell - using demisto/python:1.3-alpine
> ActiveMQ - uses a “latest” docker image
> 
> not quite sure what you mean with adding integration test to validate - to add docker check in integration_test?

looks good overall
for `latest` it looks weird:
`Packs/ActiveMQ/Integrations/ActiveMQ/ActiveMQ.yml: You're not using latest docker for the file, please update to latest version. Current: latest, Latest: latest`
lets change it that if `latest` is given as tag, raise an exception `latest` tag is not allowed, but only versioned ones 

also, for old docker image, like for example `HelloWorld` here, lets change it from error to warning and not fail the build on that

ohaim1008:
@Itay4 Fixed and I think it looks better,
content build in - https://circleci.com/gh/demisto/content/51156
I removed the unused code lines, not sure why the coverage is still a little bit lower, the marked lines were fixed.

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -42,36 +40,44 @@ def is_docker_image_valid(self):
...
-            if self.docker_image_latest_tag != self.docker_image_tag and not \
-                    'demisto/python:1.3-alpine' == '{}:{}'.format(self.docker_image_name, self.docker_image_tag):
-                # If docker image name are different and if the docker image isn't the default one
-                self.is_latest_tag = False
-        # Case of an added file
-        elif not self.is_modified_file:
-            if self.docker_image_latest_tag != self.docker_image_tag:
-                self.is_latest_tag = False
+        if self.docker_image_latest_tag != self.docker_image_tag:
+            # If docker image tag is the most updated one that exists in docker-hub
```
Itay4, line 61: not sure i understand this comment

Itay4:
there's a weird char here

image

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -42,36 +40,44 @@ def is_docker_image_valid(self):
...
 
         if not self.is_latest_tag:
-            print_error('The docker image tag is not the latest, please update it.\n'
-                        'The docker image tag in the yml file is: {}\n'
-                        'The latest docker image tag in docker hub is: {}\n'
+            self.is_latest_tag = False
+            print_warning('The docker image tag is not the latest, please update it.\n'
+                          'The docker image tag in the yml file is: {}\n'
+                          'The latest docker image tag in docker hub is: {}\n'
+                          'You can check for the tags of {} here: https://hub.docker.com/r/{}/tags\n'
```
Itay4, line 69:
```suggestion
                          'You can check for the most updated version of {} here: https://hub.docker.com/r/{}/tags\n'
```

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -42,36 +40,44 @@ def is_docker_image_valid(self):
...
+                          .format(self.docker_image_tag, self.docker_image_latest_tag, self.docker_image_name,
+                                  self.docker_image_name))
+
+        # the most updated tag should be numeric and not labeled "latest"
+        if self.docker_image_latest_tag == "latest":
+            self.is_latest_tag = False
+            print_error('"￿Latest" tag is not allowed,\n'
```
Itay4, line 76:
```suggestion
            print_error('"latest" tag is not allowed,\n'
```

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -42,36 +40,44 @@ def is_docker_image_valid(self):
...
+                        'Please create a new updated versioned image\n'
```
Itay4, line 77:
```suggestion
                        'Please create or update to an updated versioned image\n'
```

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -42,36 +40,44 @@ def is_docker_image_valid(self):
...
                         'You can check for the tags of {} here: https://hub.docker.com/r/{}/tags\n'
```
Itay4, line 78:
```suggestion
                        'You can check for the most updated version of {} here: https://hub.docker.com/r/{}/tags\n'
```

demisto_sdk/commands/common/tools.py
```diff
@@ -456,17 +456,6 @@ def run_threads_list(threads_list):
         t.join()
 
 
-def get_dockerimage45(script_object):
```
Itay4, line 459: why did you remove this?

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -42,36 +40,44 @@ def is_docker_image_valid(self):
...
+        if self.docker_image_latest_tag != self.docker_image_tag:
+            # If docker image tag is the most updated one that exists in docker-hub
+            self.is_latest_tag = False
 
         if not self.is_latest_tag:
-            print_error('The docker image tag is not the latest, please update it.\n'
-                        'The docker image tag in the yml file is: {}\n'
-                        'The latest docker image tag in docker hub is: {}\n'
+            self.is_latest_tag = False
+            print_warning('The docker image tag is not the latest, please update it.\n'
```
Itay4, line 66: do we have a test that checks this warning is printed?

Itay4:
@ohaim1008 please have a look at the following and lets try to cover the red lines:
https://coveralls.io/builds/30538706/source?filename=demisto_sdk%2Fcommands%2Fcommon%2Fhook_validations%2Fdocker.py

demisto_sdk/commands/common/hook_validations/docker.py
ohaim1008: yes it's in test_negative__non_latest_docker_image in validatie_integration_test.py

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Missing release notes message are now showing the release notes file path to update.
 * Fixed an issue in the **validate** command in which unified YAML files were not ignored.
 * File format suggestions are now shown in the relevant file format (JSON or YAML).
+* Fixed docker image tag checking in validation hook.
```
Itay4, line 9:
```suggestion
* Changed Docker image validation to fail only on non-valid ones.
* Removed backward compatibility validation when Docker image is updated.
```

PR #335, opened 2020-04-07
in get remote file, check if json file or yml file, if not than do no…

orhovy:
…t try to load it , return empty dict.

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23339

## Description
Added a check for file type , whether not a yml nor a json we wont load it and return an empty dict.

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -141,6 +141,7 @@ def get_remote_file(full_file_path, tag='master'):
         details = json.loads(res.content)
     elif full_file_path.endswith('yml'):
         details = yaml.safe_load(res.content)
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -139,9 +139,10 @@ def get_remote_file(full_file_path, tag='master'):
 
     if full_file_path.endswith('json'):
         details = json.loads(res.content)
-    else:
+    elif full_file_path.endswith('yml'):
         details = yaml.safe_load(res.content)
-
+    else:
```
bakatzir, line 144: let's add a comment saying it is probably a readme/changelog

PR #340, opened 2020-04-12
changelog for fix get_remote_file for md files

orhovy:
## Status
- [x] Ready


orhovy:
@Shellyber 
also I added it to changelog of 0.4.9 because there is no changelog for 0.5.0

PR #359, opened 2020-04-21
Fix bug secrets escape

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Description
handling escape characters from whitelist.json.

## Must have
- [x] Tests

### DIFF
--- a/demisto_sdk/commands/secrets/tests/secrets_test.py
+++ b/demisto_sdk/commands/secrets/tests/secrets_test.py
@@ -8,7 +8,15 @@ from demisto_sdk.commands.common.git_tools import git_path
 from demisto_sdk.commands.secrets.secrets import SecretsValidator
 
 
-def create_whitelist_secrets_file(file_path, urls=[], ips=[], files=[], generic_strings=[]):
+def create_whitelist_secrets_file(file_path, urls=None, ips=None, files=None, generic_strings=None):
+    if files is None:
+        files = []
+    if urls is None:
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/secrets/tests/secrets_test.py
```diff
@@ -172,9 +180,17 @@ def test_remove_white_list_regex(self):
...
+        file_contents = self.validator.remove_whitelisted_items_from_file(file_contents, white_list)
         assert white_list not in file_contents
 
+    def test_remove_whitelisted_items_from_file_escaped_whitelist(self):
+        white_list = '***.url'
+        file_contents = '''
+        Random and unmeaningful file content
+        '''
+        file_contents = self.validator.remove_whitelisted_items_from_file(file_contents, white_list)
+        assert file_contents
```
Itay4, line 192: What are we checking here?

jochman: The secret contains `***.url` which is an invalid regex that causes the system to raise an exception.

Itay4:
And how exactly that test checks that?

jochman:
> And how exactly that test checks that?

I've found that this term failed when running secrets. So I fixed it and added a unittest to see that it's not failing again.

Itay4:
It is extremely unclear... let’s clarify what we’re testing here in the test docstring and let’s assert something that actually checks for something

PR #370, opened 2020-04-25
Add pyupgrade to pre commit hooks

Itay4:
## Status
- [x] Ready

## Description
Added 2 pre-commit hooks:
1. black - we have inconsistent code style in the repo cause of refactoring done in a few iterations, need `black` to try and align to one style
2. pyupgrade - we still have some leftovers of python 2 syntax in the code, need `pyupgrade` to update those to python 3


glicht:
@Itay4 I think we should exclude the test files. Looking at the build I see the hooks are running also on test files.

Itay4:
@glicht i've removed `black` as it doesn't run well with `flake8` and `autopep8`

glicht:
So, if you've removed `black` do we need all the formatting changes you've done?

Itay4:
@glicht done

PR #374, opened 2020-04-27
halpert Handling the case file_path is a tuple due to renaming or moving

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

### DIFF
--- a/demisto_sdk/commands/validate/file_validator.py
+++ b/demisto_sdk/commands/validate/file_validator.py
@@ -276,7 +276,9 @@ class FilesValidator:
             modified_files (set): A set of the modified files in the current branch.
         """
         for file_path in modified_files:
-            file_type = find_type(file_path)
+            # modified_files are returning from running git diff.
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/validate/file_validator.py
```diff
@@ -276,7 +276,8 @@ def validate_modified_files(self, modified_files):  # noqa: C901
             modified_files (set): A set of the modified files in the current branch.
         """
         for file_path in modified_files:
-            file_type = find_type(file_path)
+            # in case of moving files- file_path is a tuple in length of 2
```
Itay4, line 279: how can you know that it is of length 2?  why not check it? do we have a test for the case it is a tuple?

hod-alpert: Because when the file was renamed it shows the original name and the new name.
 I don't think it can be anything more than that

Itay4: lets clarify the comment and mention its returned from git diff

PR #377, opened 2020-04-27
Generate Docs Bug Fixes

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23824
fixes: https://github.com/demisto/etc/issues/23914
fixes: https://github.com/demisto/etc/issues/23931

## Description
1. Removes the `Source` column of the inputs table of the docs generated for playbooks by the `generate-docs` command.
2. The row for the inputs table generated for a playbook that accepted a `playbookInputQuery` as input was practically empty. Added default values for the `Name` and `Description` columns for inputs of this type to playbooks.
3. Quotation marks (single and double) were accidentally escaped if they appeared in the `Description` column. Fixed.
4. Updated the `Default Value` column to be the combined `root` and `accessor` (separated by a period) where applicable.

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
+++ b/demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
@@ -17,16 +17,16 @@ TEST_INTEGRATION_PATH = os.path.join(FILES_PATH, 'fake_integration/fake_integrat
 
 
 def test_stringEscapeMD():
-    from demisto_sdk.commands.generate_docs.common import stringEscapeMD
-    res = stringEscapeMD('First fetch timestamp (<number> <time unit>, e.g., 12 hours, 7 days)')
+    from demisto_sdk.commands.generate_docs.common import string_escape_md
+    res = string_escape_md('First fetch timestamp (<number> <time unit>, e.g., 12 hours, 7 days)')
     assert '<' not in res
     assert '>' not in res
-    res = stringEscapeMD("new line test \n", escape_multiline=True)
+    res = string_escape_md("new line test \n", escape_multiline=True)
     assert '\n' not in res
     assert '<br/>' in res
-    res = stringEscapeMD('Here are "Double Quotation" marks')
+    res = string_escape_md('Here are "Double Quotation" marks')
     assert '"' in res
-    res = stringEscapeMD("Here are 'Single Quotation' marks")
+    res = string_escape_md("Here are 'Single Quotation' marks")
     assert "'" in res
 
 
--- a/demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
+++ b/demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
@@ -24,6 +24,10 @@ def test_stringEscapeMD():
     res = stringEscapeMD("new line test \n", escape_multiline=True)
     assert '\n' not in res
     assert '<br/>' in res
+    res = stringEscapeMD('Here are "Double Quotation" marks')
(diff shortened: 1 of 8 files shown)


demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
```diff
@@ -103,10 +107,22 @@ def test_get_inputs():
...
-                        'Source': '', 'Required': 'Required'}]
+    expected_query = '(type:ip or type:file or type:Domain or type:URL) -tags:pending_review ' \
+        'and (tags:approved_black or tags:approved_white or tags:approved_watchlist)'
+    expected_inputs = [
+        {
+            'Name': 'InputA', 'Description': '', 'Default Value': 'File.Name', 'Required': 'Optional'
+        },
+        {
+            'Name': 'InputB', 'Description': 'This is input b',
+            'Default Value': 'johnnydepp@gmail.com', 'Required': 'Required'
```
Itay4, line 118: i like jonny

avidan-H:
> cool
>
> 1. for the two functions you updated, please add type hints in their signatures and convert the docstring to google-style
> 2. stringEscapeMD ?? seriously??? what pep is that exactly ?? please refactor func name
> 3. please add integration test for generate playbook docs under /demisto_sdk/tests/integration_tests/generate_docs_integration_test.py

"cool"

1. okiedokie will do.
2. I didn't name this - but okiedokie will do.
3. I don't understand. How come I should add a test to the integration stuff when I changed the playbook stuff?

demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
avidan-H: It wasn't my addition. It was there before my edits, so I can't take credit.

demisto_sdk/tests/integration_tests/generate_docs_integration_test.py
```diff
@@ -0,0 +1,95 @@
...
+from demisto_sdk.commands.common.git_tools import git_path
+from os.path import join
+from pathlib import Path
+
+from click.testing import CliRunner
+from demisto_sdk.__main__ import main
+
+GENERATE_DOCS_CMD = "generate-docs"
+DEMISTO_SDK_PATH = join(git_path(), "demisto_sdk")
+VALID_PLAYBOOK_WITH_IO = join(DEMISTO_SDK_PATH, "tests/test_files/playbook-Test_playbook.yml")
```
Itay4, line 11: these seem to be used only in specific tests, why do they need to be global consts?

demisto_sdk/tests/integration_tests/generate_docs_integration_test.py
```diff
@@ -0,0 +1,95 @@
...
+        - Ensure README.md has an outputs section.
+        """
+
+        runner = CliRunner(mix_stderr=False)
+        arguments = [
+            GENERATE_DOCS_CMD,
+            '-i', VALID_PLAYBOOK_WITH_IO,
+            '-o', tmpdir
+        ]
+        _ = runner.invoke(main, arguments)
```
Itay4, line 39: also assert status code, no stderr, stdout and no exception

avidan-H: I'm not checking `stdout` just cause it can output that you're not using the latest version, (if you're testing locally).

PR #379, opened 2020-04-29
flake8 plugin - direct dict key access

Itay4:
## Status
- [x] In Progress

## Description
Added a flake8 custom plugin to enforce the non-usage of direct dict key access

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,7 @@ jobs:
             name: pre-commit
             command: |
               . .tox/py37/bin/activate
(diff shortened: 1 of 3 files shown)


ronykoz:
This might cause a lot of noise in the repo, not sure it's worth the effort, and if enabling this on only changed + modified it will also cause plenty of noise because we will have to fix the entire module of all these errors, not sure it's worth the effort

anara123:
I also don't think it is worth it. Also I think if we add validation it should be for all the repo and not just for the new/modified ones. So it is a lot of work

PR #388, opened 2020-04-30
Common docs - copied to packs bundle

IkaDemisto:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold 

## Description
Added copy of docs files to packs bundle.


demisto_sdk/commands/create_artifacts/tests/content_creator_test.py
```diff
@@ -89,3 +89,35 @@ def test_copy_packs_content_to_packs_bundle(self):
...
+        mocker.patch('builtins.print')
+        mocker.patch('os.mkdir')
+
+        doc_files_paths = ('./Documentation/doc-CommonServer.json', './Documentation/doc-howto.json')
+        mocker.patch('os.path.exists', side_effect=lambda p: p in doc_files_paths)
+
+        copied_to_bundle = mocker.patch('shutil.copyfile')
+        copied_to_packs = mocker.patch('shutil.copy')
+
+        ContentCreator.copy_docs_files(content_bundle_path="dummy_bundle_path", packs_bundle_path="dummy_packs_path")
```
anara123, line 111: I prefer not to test specific methods (inner/private methods) but public methods. In this case I would prefer to test `create_content` and validate that doc files are in the content_new.zip.
But if it is too much of a problem because of the current code structure, please let me know

IkaDemisto: `content_creator` is not testable. Additionally common doc files aren't created in `create-content-artifacts` and running `create_content` takes ~6 minutes to run with actual data. Single unit test that runs 7 minutes is just wrong.

PR #392, opened 2020-05-03
print yaml formed error instead of json for integration files

jochman:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23586

## Screenshots
image

## Must have
- [x] Tests

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/integration.py
+++ b/demisto_sdk/commands/common/hook_validations/integration.py
@@ -579,12 +579,13 @@ class IntegrationValidator(BaseValidator):
             params = [_key for _key in self.current_file.get('configuration', [])]
             for param in FETCH_REQUIRED_PARAMS:
                 if param not in params:
-                    print_error(f'A required parameter'
-                                f'\n```\n{yaml.dump(param)}```\n'
-                                f' is missing or malformed in the YAML file.'
-                                'The correct format of the parameter should be as follows:'
-                                f'\n```\n{yaml.dump(param)}```\n'
-                                f'If another other arguments exists in the param remove them.')
+                    print_error(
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -579,9 +580,10 @@ def is_valid_fetch(self) -> bool:
             for param in FETCH_REQUIRED_PARAMS:
                 if param not in params:
                     print_error(f'Integration with fetch-incidents was detected '
```
Itay4, line 582: lets change the error message a bit:
```
A required parameter <PARAM-NAME> is missing or malformed in the YAML file.
The correct format of the parameter should be as follows:
<YAML-DUMP>
```

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -578,10 +579,12 @@ def is_valid_fetch(self) -> bool:
             params = [_key for _key in self.current_file.get('configuration', [])]
             for param in FETCH_REQUIRED_PARAMS:
                 if param not in params:
-                    print_error(f'Integration with fetch-incidents was detected '
-                                f'("isfetch:  true" was found in the YAML file).'
-                                f'\nA required parameter is missing or malformed in the file {self.file_path}, '
-                                f'the param is:\n{param}')
+                    print_error(f'A required parameter'
+                                f'\n```\n{yaml.dump(param)}```\n'
```
Itay4, line 583: just the param name, no need the whole dump here
lets remove the backticks
lets remove the line you added (If another other arguments exists in the param remove them.)

PR #393, opened 2020-05-03
fix mypy in commands-builder

jochman:
- [x] Ready 

image


demisto_sdk/commands/lint/commands_builder.py
```diff
@@ -19,7 +19,7 @@ def get_python_exec(py_num: float) -> str:
         str: python executable
     """
     if py_num < 3:
-        py_num = ""
+        py_num = 2
```
Itay4, line 22: id prefer we keep it as empty string, so lets change the assignment below to be `py_num = "3"`

PR #394, opened 2020-05-03
fix mypy: constants, tools

jochman:
- [x] Ready

image


demisto_sdk/commands/common/tools.py
```diff
@@ -683,36 +683,39 @@ def get_content_path() -> str:
...
+        return git_repo.working_dir
     except (git.InvalidGitRepositoryError, git.NoSuchPathError):
         print_error("Please run demisto-sdk in content repository - Aborting!")
-
-    return git_repo.working_dir
+    return ''
 
 
-def run_command_os(command: str, cwd: Path, env: dict = os.environ) -> Tuple[str, str, int]:
+def run_command_os(command: str, cwd: Union[Path, str], env=os.environ) -> Tuple[str, str, int]:
```
Itay4, line 697:
```suggestion
def run_command_os(command: str, cwd: Union[Path, str], env: os._Environ = os.environ) -> Tuple[str, str, int]:
```

demisto_sdk/commands/common/tools.py
```diff
@@ -683,36 +683,39 @@ def get_content_path() -> str:
...
         command(str): Command to be executed.
         cwd(Path): Path from pathlib object to be executed
-        env: Enviorment variables for the execution
+        env: Environment variables for the execution
     Returns:
         str: Stdout of the command
         str: Stderr of the command
         int: exit code of command
     """
+    if isinstance(cwd, str):
```
Itay4, line 708: mypy passes for me without this change, why is this needed?

PR #395, opened 2020-05-03
mypy fix helprers, initiator, linter

jochman:
## Status
- [x] Ready

image

fixes:
```python
demisto_sdk/commands/lint/helpers.py:100: error: Incompatible return value type (got "int", expected "bool")
demisto_sdk/commands/lint/helpers.py:266: error: Name 'container_obj' already defined on line 260
demisto_sdk/commands/lint/helpers.py:289: error: Incompatible return value type (got "str", expected "float")
demisto_sdk/commands/lint/helpers.py:309: error: Item "None" of "Optional[IO[bytes]]" has no attribute "read"
demisto_sdk/commands/lint/helpers.py:313: error: Incompatible return value type (got "Union[bytes, Any]", expected "str")
demisto_sdk/commands/init/initiator.py:258: error: Incompatible types in assignment (expression has type "int", variable has type "str")
demisto_sdk/commands/init/initiator.py:267: error: Unsupported operand types for - ("str" and "int")
```

### DIFF
--- a/demisto_sdk/commands/lint/helpers.py
+++ b/demisto_sdk/commands/lint/helpers.py
@@ -10,7 +10,7 @@ import textwrap
 from contextlib import contextmanager
 from functools import lru_cache
 from pathlib import Path
-from typing import Dict, Generator, List, Optional
+from typing import IO, Dict, Generator, List, Optional, Union
 
 # Third party packages
 import docker
@@ -54,7 +54,9 @@ logger = logging.getLogger('demisto-sdk')
 
 
 def validate_env() -> None:
-    """Packs which use python2 will need to be run inside virtual enviorment including python2 as main and the specified req"""
+    """Packs which use python2 will need to be run inside virtual environment including python2 as main
+    and the specified req
+    """
     wrn_msg = 'demisto-sdk lint not in virtual environment, Python2 lints will fail, use "source .hooks/bootstrap"' \
               ' to create the virtual environment'
     command = "python -c \"import sys; print('{}.{}'.format(sys.version_info[0], sys.version_info[1]))\""
@@ -69,7 +71,7 @@ def validate_env() -> None:
 
 
 def build_skipped_exit_code(no_flake8: bool, no_bandit: bool, no_mypy: bool, no_pylint: bool, no_vulture: bool,
-                            no_test: bool, no_pwsh_analyze: bool, no_pwsh_test: bool, docker_engine: bool) -> bool:
+                            no_test: bool, no_pwsh_analyze: bool, no_pwsh_test: bool, docker_engine: bool) -> int:
     """
     no_flake8(bool): Whether to skip flake8.
     no_bandit(bool): Whether to skip bandit.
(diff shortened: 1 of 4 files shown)


demisto_sdk/commands/lint/helpers.py
```diff
@@ -257,58 +259,64 @@ def get_python_version_from_image(image: str) -> float:
         float: Python version X.Y (3.7, 3.6, ..)
     """
     docker_client = docker.from_env()
-    container_obj: Container = None
-    py_num = ""
-    for trial1 in range(2):
+    containers = list()
```
Itay4, line 262: why do we need that list ?

demisto_sdk/commands/lint/helpers.py
```diff
@@ -257,58 +259,64 @@ def get_python_version_from_image(image: str) -> float:
...
             # Wait for container to finish
             container_obj.wait(condition="exited")
             # Get python version
             py_num = container_obj.logs()
             if isinstance(py_num, bytes):
                 py_num = float(py_num)
-                break
             else:
                 raise docker.errors.ContainerError
+            for _ in range(2):
```
Itay4, line 283: why did you move it here?

demisto_sdk/commands/lint/helpers.py
```diff
@@ -257,58 +259,64 @@ def get_python_version_from_image(image: str) -> float:
...
     """
+    data: Union[str, bytes] = b''
     archive, stat = container_obj.get_archive(container_path)
-    filelike = io.BytesIO(b"".join(b for b in archive))
-    tar = tarfile.open(fileobj=filelike)
-    data = tar.extractfile(stat['name']).read()
-    if encoding:
-        data = data.decode(encoding)
+    file_like = io.BytesIO(b"".join(b for b in archive))
+    tar: Optional[tarfile.TarFile] = tarfile.open(fileobj=file_like)
```
Itay4, line 313:
```suggestion
    tar: tarfile.TarFile = tarfile.open(fileobj=file_like)
```

demisto_sdk/commands/lint/helpers.py
```diff
@@ -257,58 +259,64 @@ def get_python_version_from_image(image: str) -> float:
...
+    if isinstance(tar, tarfile.TarFile):
```
Itay4, line 314: remove that condition

demisto_sdk/commands/lint/helpers.py
```diff
@@ -257,58 +259,64 @@ def get_python_version_from_image(image: str) -> float:
...
     """ Copy file from container.
 
     Args:
         container_obj(Container): Container ID to copy file from
         container_path(Path): Path in container image (file)
-        encoding(str): valide encoding e.g. utf-8
+        encoding(str): valid encoding e.g. utf-8
 
     Returns:
         str: file as string decode as utf-8
```
Itay4, line 305:
```suggestion
        str or bytes: File as string decoded in utf-8
```

demisto_sdk/commands/lint/helpers.py
jochman: because before that the code tried to use container_obj which can be not assigned

demisto_sdk/commands/lint/helpers.py
jochman: it's Optional, mypy will fail without the optional annotation

demisto_sdk/commands/lint/helpers.py
jochman: it must be because tafile.open can return None

PR #402, opened 2020-05-04
Download list files bug fix

guyfreund:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24107

## Description
The file was causing the error was of type `preprocessrule`, the `find_type` function doesn't handle files of this type. From now and so on files of this type will be ignored by the download command.

### DIFF
--- a/demisto_sdk/commands/download/downloader.py
+++ b/demisto_sdk/commands/download/downloader.py
@@ -33,6 +33,7 @@ from mergedeep import merge
 from ruamel.yaml import YAML
 from tabulate import tabulate
 from urllib3.exceptions import MaxRetryError
+from typing import List
 
 logging.disable(logging.CRITICAL)
 
@@ -85,11 +86,14 @@ class Downloader:
         if not self.verify_flags():
             return 1
         if not self.fetch_custom_content():
+            self.remove_traces()
             return 1
         if self.handle_list_files_flag():
+            self.remove_traces()
             return 0
         self.handle_all_custom_content_flag()
         if not self.verify_output_pack_is_pack():
+            self.remove_traces()
             return 1
         self.build_pack_content()
         self.build_custom_content()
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/download/downloader.py
```diff
@@ -161,7 +165,12 @@ def get_custom_content_objects(self) -> list:
         :return: The list of all custom content objects
         """
         custom_content_file_paths: list = get_child_files(self.custom_content_temp_dir)
-        return [self.build_custom_content_object(file_path) for file_path in custom_content_file_paths]
+        custom_content_objects: List = list()
+        for file_path in custom_content_file_paths:
+            custom_content_object = self.build_custom_content_object(file_path)
+            if custom_content_object:
```
Itay4, line 171: need a comment here

demisto_sdk/commands/download/downloader.py
```diff
@@ -355,7 +364,7 @@ def build_custom_content_object(file_path: str) -> dict:
             'entity': file_entity,
             'type': file_type,
             'file_ending': file_ending,
-        }
+        } if file_type else {}
```
Itay4, line 367: so we now ignore all content entities that are not the ones we know, right?

Itay4:
we need a test that unknown content entity, such as preprocess rule, is not caught

demisto_sdk/commands/download/downloader.py
guyfreund: yes

demisto_sdk/commands/download/downloader.py
guyfreund: sbba

ShahafBenYakir:
@guyfreund Follwong itay's comment do we state in the docs the supported entities?

Itay4:
@guyfreund update changelog as well

PR #416, opened 2020-05-07
fix validate for version in layout

moishce:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24168

## Description
Fixes that validate command checks in layout files that version is -1 in both version fields


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Changelog
+* Fixed the **validate** command on the layout files to check both version fields are -1
```
ShahafBenYakir, line 2:
```suggestion
* Fixed a bug in the **validate** command where layouts version was not checked correctly.
```

demisto_sdk/tests/test_files/layout-invalid3.json
```diff
@@ -0,0 +1,44 @@
...
+		"version": 55,
+		"modified": "2019-06-20T16:01:05.659574019-07:00",
+		"name": "",
+		"kind": "close",
+		"typeId": "ExtraHop Detection",
+		"system": false,
+		"sections": [
+			{
+				"id": "",
+				"version": 0,
```
ShahafBenYakir, line 16: Let's change this to -1. It will still fail but this way we are checking both cases (in this case it is valid on the inside one bit not on the outside one)

moishce: We check that in the other files

ronykoz:
@ShahafBenYakir @moishce any reason for not fixing conflicts and merging this?

ShahafBenYakir:
@ronykoz https://github.com/demisto/etc/issues/24168#issuecomment-628697784

PR #422, opened 2020-05-11
Image validatition will not consider image in the yml file as error

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 * Fixed a bug in **upload** command where created unified docker45 yml files were not deleted.
 * Added support for IndicatorTypes directory in packs (for `reputation` files, instead of Misc).
 * Fixed parsing playbook condition names as string instead of boolean in **validate** command
+* Improved image validation in YAML files.
 
 #### 1.0.3
 * Fixed a bug in the **format** command where comments were being removed from YAML files.
--- a/demisto_sdk/commands/common/hook_validations/image.py
+++ b/demisto_sdk/commands/common/hook_validations/image.py
@@ -80,16 +80,17 @@ class ImageValidator:
 
         if data_dictionary.get('image'):
             is_image_in_yml = True
-
         if not re.match(INTEGRATION_REGEX, self.file_path, re.IGNORECASE):
             package_path = os.path.dirname(self.file_path)
-            if is_image_in_yml:
-                print_error("You have added an image in the yml "
-                            "file, please update the package {}".format(package_path))
-                return False
             image_path = glob.glob(package_path + '/*.png')
             if image_path:
                 is_image_in_package = True
+        if is_image_in_package and is_image_in_yml:
+            print_error(
+                "The file {} has image in both yml and package, remove the 'image' key from the yml file".format(
+                    self.file_path))
+            self._is_valid = False
+            return False
 
         if not (is_image_in_package or is_image_in_yml):
             print_error("You have failed to add an image in the yml/package for {}".format(self.file_path))
--- a/demisto_sdk/commands/common/tests/image_test.py
+++ b/demisto_sdk/commands/common/tests/image_test.py
@@ -2,6 +2,9 @@ import os
 
 from demisto_sdk.commands.common.git_tools import git_path
 from demisto_sdk.commands.common.hook_validations import image
+from demisto_sdk.commands.common.hook_validations.integration import \
+    IntegrationValidator
+from demisto_sdk.commands.common.tests.integration_test import mock_structure
 
 
 def test_is_not_default_image():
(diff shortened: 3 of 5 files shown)


demisto_sdk/commands/common/hook_validations/image.py
```diff
@@ -83,10 +83,6 @@ def is_existing_image(self):
 
         if not re.match(INTEGRATION_REGEX, self.file_path, re.IGNORECASE):
             package_path = os.path.dirname(self.file_path)
```
Itay4, line 85: need to check the image is exactly in one of: yml or dir

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Changelog
 * Added support for IndicatorTypes directory in packs (for `reputation` files, instead of Misc).
+* Image in the yml file will not be considered as error.
```
Itay4, line 3:
```suggestion
* Improved image validation in YAML files.
```

demisto_sdk/commands/common/tests/image_test.py
```diff
@@ -28,3 +31,26 @@ def test_is_not_default_image():
     image.INTEGRATION_REGXES.append(int_path)
     image_validator = image.ImageValidator(int_path)
     assert image_validator.is_not_default_image() is False
+
+
+def test_is_valid_image_positive(monkeypatch):
```
Itay4, line 36: add `Given-When-Then` docstring

demisto_sdk/commands/common/tests/image_test.py
```diff
@@ -28,3 +31,26 @@ def test_is_not_default_image():
...
+    structure = mock_structure(file_path=integration_path)
+    monkeypatch.setattr('demisto_sdk.commands.common.hook_validations.image.INTEGRATION_REGXES', [integration_path])
+    # Adding monkey patching this will make image validator behave like this is an integration outside of
+    # pack context and ignore the image that's in the same folder as the file
+    monkeypatch.setattr('demisto_sdk.commands.common.hook_validations.image.INTEGRATION_REGEX', integration_path)
+    validator = IntegrationValidator(structure)
+    assert validator.is_valid_image() is True
+
+
+def test_image_in_both_yml_and_directory(monkeypatch):
```
Itay4, line 49: add `Given-When-Then` docstring

demisto_sdk/commands/common/hook_validations/image.py
```diff
@@ -80,16 +80,17 @@ def is_existing_image(self):
...
-                            "file, please update the package {}".format(package_path))
-                return False
             image_path = glob.glob(package_path + '/*.png')
             if image_path:
                 is_image_in_package = True
+        if is_image_in_package and is_image_in_yml:
+            print_error(
+                "The file {} has image in both yml and package, remove the 'image' key from the yml file".format(
+                    self.file_path))
+            self._is_valid = False
```
Itay4, line 92: do we need to fill this in? not just return false?

demisto_sdk/commands/common/hook_validations/image.py
```diff
@@ -80,16 +80,17 @@ def is_existing_image(self):
...
         if not re.match(INTEGRATION_REGEX, self.file_path, re.IGNORECASE):
             package_path = os.path.dirname(self.file_path)
-            if is_image_in_yml:
-                print_error("You have added an image in the yml "
-                            "file, please update the package {}".format(package_path))
-                return False
             image_path = glob.glob(package_path + '/*.png')
             if image_path:
                 is_image_in_package = True
+        if is_image_in_package and is_image_in_yml:
```
Itay4, line 88: i think the condition should be that it is exactly in one, not in both

hod-alpert: I'm returning error if it's in both and if it's in none of the two, if it's in one- no exception will be raised, is that what you mean?

demisto_sdk/commands/common/hook_validations/image.py
hod-alpert: How will the user know what went wrong if we don't do this?

Itay4: im asking if by returning `False` does it not set `_is_valid` to `False` ?

Itay4: yea, missed second condition
looks good

demisto_sdk/commands/common/hook_validations/image.py
Itay4: yea, missed second condition
looks good

demisto_sdk/commands/common/hook_validations/image.py
hod-alpert: It doesn't set _is_valid to False

PR #423, opened 2020-05-11
Playbook validation fixes

hod-alpert:
* Removed condition branches check from test playbooks
* Fixed parsing playbook condition names as string instead of boolean
* Added valid fields to the schema

## Status
- [ ] In Progress
- [x] Ready

## Must have
- [x] Tests
- [x] Documentation


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
-
+* Removed condition branches check from test playbooks
+* Fixed parsing playbook condition names as string instead of boolean
+* Added valid fields to the schema
```
Itay4, line 4:
```suggestion
```

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
-
+* Removed condition branches check from test playbooks
```
Itay4, line 2:
```suggestion
```

demisto_sdk/commands/common/hook_validations/playbook.py
```diff
@@ -90,6 +90,8 @@ def is_condition_branches_handled(self):  # type: () -> bool
         Return:
             bool. if the Playbook handles all condition branches correctly.
         """
+        if 'TestPlaybooks' in self.file_path:
```
Itay4, line 93: verify we have test that cover this condition

demisto_sdk/commands/common/hook_validations/playbook.py
```diff
@@ -125,14 +127,14 @@ def is_builtin_condition_task_branches_handled(self, task: Dict) -> bool:
         for condition in task.get('conditions', []):
             label = condition.get('label')
             if label:
-                task_condition_labels.add(label.upper())
+                task_condition_labels.add(str(label).upper())
```
Itay4, line 130: verify that it works as expected, and if it does add comment why the str cast is needed

hod-alpert: Works when consistent- meaning either fixed to be **'yes'** instead of **yes** in all places or not fixed in all places.
I think we don't have a reason to think it will be partially fixed nowhere because before this fix it would have been failed.

To sum it all- I think it's fine.

demisto_sdk/commands/common/hook_validations/playbook.py
Itay4: @hod-alpert did you verify that?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Changelog
 * Added support for IndicatorTypes directory in packs (for `reputation` files, instead of Misc).
+* Fixed parsing playbook condition names as string instead of boolean
```
Itay4, line 3: mention it is in the validate command

demisto_sdk/commands/common/hook_validations/playbook.py
hod-alpert: Yes, I added a test for that ` test_unhandled_condition_skipping`

PR #427, opened 2020-05-11
fix bug in format indicatorfield

moishce:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: demisto/etc#24357

## Description
added schemas for indicatorfield


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Changelog
+* Fixed a bug in the **format** command where indicatorField field.
```
ShahafBenYakir, line 2: What was the bug we fixed? Was it any fix or we just added tests to validate indicator fields?

ShahafBenYakir, line 2: I think we can remove this because we dod not fix anything here, the fix was already mentioned in this file right?

moishce:
The bug was because there was a missing indicatorfield file in schemas,
But it has already been added by @hod-alpert ,
So I just arranged the tests that would really test what was needed

CHANGELOG.md
ShahafBenYakir: Fixed a bug in the **format** command where `indicatorField` wasn't part of the SDK schemas.

ShahafBenYakir:
Change the changelog to this https://github.com/demisto/demisto-sdk/pull/427/files/15cca8f07b394a427349281f69f39a1cfd53c02c..7948da98f8b091d78a3a8b95c60dda2069245e70#r424309615

PR #432, opened 2020-05-12
Change RNValidator to use new RN format and old format

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

related docs PR: https://github.com/demisto/content-docs/pull/121


demisto_sdk/commands/common/tests/release_notes_test.py
```diff
@@ -1,91 +1,60 @@
...
+    - Not filled out release notes.
+    - Valid release notes
 
+    When
+    - Running validation on release notes.
 
-@pytest.mark.parametrize('release_notes_diff, expected_result', diff_package)
-def test_rn_master_diff(release_notes_diff, expected_result, mocker):
+    Then
+    - Ensure validation correctly identifies valid release notes.
```
yuvalbenshalom, line 38: you need to writhe here what dose the validation expects to find/print in this case

PR #435, opened 2020-05-12
support private repositories

anara123:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
related to https://github.com/demisto/etc/issues/24216

## Description
Update demisto-sdk to support private repositories


demisto_sdk/commands/validate/file_validator.py
```diff
@@ -674,18 +680,43 @@ def run_all_validations_on_file(self, file_path: str, file_type: str = None) ->
...
+
         packs = {os.path.basename(pack) for pack in glob(f'{PACKS_DIR}/*')}
         self.validate_pack_unique_files(packs)
         all_files_to_validate = set()
-        for directory in [PACKS_DIR, BETA_INTEGRATIONS_DIR, TEST_PLAYBOOKS_DIR]:
-            all_files_to_validate |= {file for file in glob(fr'{directory}/**', recursive=True) if
-                                      not should_file_skip_validation(file)}
+
+        # go over packs
+        for pack_name in os.listdir(PACKS_DIR):
```
ronykoz, line 697: what about TEST_PLAYBOOKS_DIR ?

demisto_sdk/commands/validate/file_validator.py
```diff
@@ -674,18 +680,43 @@ def run_all_validations_on_file(self, file_path: str, file_type: str = None) ->
...
-        print('Validating conf.json')
-        conf_json_validator = ConfJsonValidator()
-        if not conf_json_validator.is_valid_conf_json():
-            self._is_valid = False
+
+        if not skip_conf_json:
+            print('Validating conf.json')
+            conf_json_validator = ConfJsonValidator()
+            if not conf_json_validator.is_valid_conf_json():
+                self._is_valid = False
```
ronykoz, line 690: we have this on line 804, do we need this here too?

demisto_sdk/commands/validate/file_validator.py
```diff
@@ -101,16 +101,22 @@ def __init__(self, is_backward_check=True, prev_ver=None, use_git=False, is_circ
...
         self.validate_id_set = validate_id_set
         self.file_path = file_path
         self.skip_pack_rn_validation = skip_pack_rn_validation
         self.changed_pack_data = set()
 
-        if self.validate_conf_json:
+        self.is_private_repo = is_private_repo
+        if is_private_repo:
+            print('Running in a private repository')
+            self.skip_conf_json = True  # private repository don't have conf.json file
```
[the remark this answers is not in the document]
anara123: This is their private repository, they can decide what to run, how to run it.
The partner will have to figure out how to run test playbooks  (probably it will not be part of the build - my guess). So there is no need to have `conf.json`

demisto_sdk/commands/validate/file_validator.py
anara123: Will not exist next week. All test playbooks should be in packs

demisto_sdk/commands/validate/file_validator.py
anara123: This is not the same piece of code/use case. I don't see a reason to have them both, but I will not change this in this PR.
`validate_all_files ` != `validate_all_files_schema `

PR #445, opened 2020-05-17
[Fix] demisto-sdk validate should not fail on missing "else" for conditions

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24451

## Description
Removed validation for else path in playbook condition tasks.
Tested with the file mentioned in the issue - following the fix the validate passed.

## Must have
- [x] Tests


DeanArbel:
@Itay4 please note that both conditions are tested - with an else path and without one. I added a valid "no else path" to the test data, to assert the new behavior.

PR #448, opened 2020-05-17
Validate fail on empty/missing args field - fix

dantavori:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24534

## Description
Fixed an issue where validate fails on empty/missing "args" field in yml.


yaakovi:
add a test

demisto_sdk/commands/common/hook_validations/script.py
```diff
@@ -125,16 +126,23 @@ def is_added_required_args(self):
     def is_there_duplicates_args(self):
         # type: () -> bool
         """Check if there are duplicated arguments."""
-        args = [arg['name'] for arg in self.current_file.get('args', [])]
-        if len(args) != len(set(args)):
-            return True
+        args = self.current_file.get('args', [])
+        if args:
+            args = [arg['name'] for arg in args]
```
teizenman, line 131: Maybe it would be better to use `arg.get('name') for arg in args if arg.get('name') is not None`
So we wont fail on the bad args?

dantavori: The problem is that although in line 129 we defined `args` to be an empty list by default, for some reason its value is eventually `None` and then we get a null pointer exception (you can see in the issue).
Anyway after consulting with @ronykoz I'm closing this PR for now - the failure was simply fixed by itself somehow. If it appears again we'll handle it.

PR #464, opened 2020-05-19
Added validation stage to the sdk build

reutshal:
## Status
- [] In Progress
- [X] Ready


.circleci/config.yml
```diff
@@ -32,6 +32,40 @@ jobs:
...
               coveralls
+        - run:
+            name: test validate files and yaml
+            command: |
+              . .tox/py37/bin/activate
+
+              git clone https://github.com/demisto/content.git
+              cd content
+
+              if [[ $CIRCLE_BRANCH != master && $FAKE_MASTER != true ]]; then
```
yaakovi, line 43: what's `FAKE_MASTER` supposed to mean?

reutshal:
> so,
> regular builds will run validate of a single pack
> and master builds will run validate of content's master, create id set, and create artifacts?

Exactly.

.circleci/config.yml
reutshal: Like we talked today, it is a way to trigger the sdk build as it would run on master

yaakovi: in that case, you can switch the condition to depend only on that variable (when not defined, will enter the condition scope):
```suggestion
              if [[ -z "$FAKE_MASTER" ]]; then
```

PR #466, opened 2020-05-20
fixing an issue where unifier randomly selected code files in a package

ronykoz:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23526

## Description
fixing an issue where unifier randomly selected code files in a package

Will update the code to use an error code once @roysagi will merge his PR. apart from that, we are ready


demisto_sdk/commands/unify/unifier.py
```diff
@@ -216,8 +216,17 @@ def get_code_file(self, script_type):
         if self.package_path.endswith('ApiModule'):
             return os.path.join(self.package_path, os.path.basename(os.path.normpath(self.package_path)) + '.py')
 
+        yml_name = os.path.basename(self.yml_path).split('.')[0]
```
yaakovi, line 219: use `os.path.splitext` in case the integration name contains `.` (like we had `RSA Netwitness 11.1` and was renames due to code not supporting it)

demisto_sdk/commands/unify/unifier.py
```diff
@@ -216,8 +216,17 @@ def get_code_file(self, script_type):
...
+
         script_path = list(filter(lambda x: not re.search(ignore_regex, x),
-                                  glob.glob(os.path.join(self.package_path, '*' + script_type))))[0]
+                                  glob.glob(os.path.join(self.package_path, f'{yml_name}' + script_type))))
```
yaakovi, line 222: why do we need `glob` if you search for a file with a specific name?
just use `script_file_path = os.path.splitent(self.yml_path)[0] + '.py'`

demisto_sdk/commands/unify/unifier.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, input: str, dir_name=INTEGRATIONS_DIR, output: str = '',
             # Also,
             # we don't take the unified file by default because
             # there might be packages that were not created by the plugin.
-            if 'unified' not in path:
+            if 'unified' not in path and os.path.basename(os.path.dirname(path)) not in [SCRIPTS_DIR, INTEGRATIONS_DIR]:
```
yaakovi, line 53: what if I'm running the script on a package unrelated to content repo?
I don't think that we should make decisions based on file path in this flow.

PR #475, opened 2020-05-24
adding CHANGELOG to error-codes change

roysagi:
Adding Changelog for the previous change


ronykoz:
@roysagi i think we can close this one, right?

PR #479, opened 2020-05-25
Validation fix for Excluded files in release notes

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: Internal bug


demisto_sdk/commands/validate/file_validator.py
```diff
@@ -307,7 +307,12 @@ def validate_modified_files(self, modified_files):  # noqa: C901
         Args:
             modified_files (set): A set of the modified files in the current branch.
         """
-        changed_packs = self.get_packs(modified_files)
+        _modified_files = set()
+        for mod_file in modified_files:
+            if not any(non_permitted_type in mod_file.lower() for non_permitted_type in ALL_FILES_VALIDATION_IGNORE_WHITELIST):
+                if 'ReleaseNotes' not in mod_file.lower():
+                    modified_files.add(mod_file)
+        changed_packs = self.get_packs(_modified_files)
```
yuvalbenshalom, line 315: is this complicated code nesseary? why not just go over modified_files and filter 'ReleaseNotes'?

amshamah419: One check is for any white listed item, and the other check is for the release notes. I need to exclude not just release notes from the changed list.

PR #483, opened 2020-05-26
ReleaseNotes object in TestSuite

dantavori:
## Status
- [x] In Progress
- [ ] Ready

## Description
This feature makes using SDK's TestSuite in testing release notes generator in content repo possible.


TestSuite/pack.py
```diff
@@ -76,6 +81,7 @@ def __init__(self, packs_dir: Path, name: str, repo):
         self.readme = TextBased(self._pack_path, 'README.md')
 
         self.pack_metadata = JSONBased(self._pack_path, 'pack_metadata.json', '')
+        self.pack_metadata.write_json(self.create_metadata_content())
```
jochman, line 84: use create_metadata_content() to write the json and return the metadata.

TestSuite/pack.py
```diff
@@ -126,6 +135,9 @@ def create_script(
             image
         )
         self.scripts.append(script)
+
+        script_filename = name + '.yml'
+        self.added_files.add(self._scripts_path / name / script_filename)
```
jochman, line 140: why do we need the added_files?

TestSuite/release_notes.py
```diff
@@ -0,0 +1,31 @@
+import json
+
+from pathlib import Path
+from demisto_sdk.commands.update_release_notes.update_rn import UpdateRN
+
+
+class ReleaseNotes(UpdateRN):
```
jochman, line 7: it's odd that ReleaseNoted is inherited from UpdateRN.
Let's get rid of the inheritance add all of the functionality to ReleaseNotes without inheritance.

demisto_sdk/tests/integration_tests/update_rn_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+    When
+    - Updating release notes.
+
+    Then
+    - Ensure release notes update is valid.
+    """
+    pack = repo.create_pack('FakePack')
+    with ChangeCWD(pack.repo_path):
+        create_integration_and_release_notes(pack, 'revision', idx=0)
```
jochman, line 40: tear apart to 3 different tests

TestSuite/pack.py
dantavori: the UpdateRN class expects this set in order to create the release notes. when you use create_integration() or create_pack(), it will be added to this set (can use it for other types, but for start those are enough), and then when you call update_release_notes() it uses this set to indicate created files and then "flushes" its content for the next use of the function.

demisto_sdk/tests/integration_tests/update_rn_test.py
dantavori: I splitted them, however they must use the same Pack object and therefore cannot be separated into different *test* functions.

TestSuite/pack.py
jochman: so why don't you create a function that will calculate those?
With the necessity to add each object to this list manually, the next developer will forget to do so. Create a function that returns all file paths insteads

demisto_sdk/tests/integration_tests/update_release_notes_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Then
+    - Ensure release notes update is NOT valid.
+    """
+    create_integration_and_release_notes(pack, 'major', idx=2, update_rn_content=False)
+    assert not is_release_note_valid(mocker, pack, idx=2)
+    assert '2_0_0.md' in os.listdir(os.path.join(pack.path, 'ReleaseNotes'))
+
+
+def test_update_release_notes(mocker, pack):
```
jochman, line 75: I meant to run each test in a different test and not under the same function. Just remove this function and rename all other function to `test_<name>`
don't forget to add a `with ChangeCWD(pack.repo_path)` to each of them.
This will make pytest run a new session for each test and if one test fail it'll tell you exactly where instead searching what failed in all those asserts.

demisto_sdk/tests/integration_tests/update_rn_test.py
jochman: 👍

TestSuite/pack.py
dantavori: I don't add it manually, it's added in create_integration() and create_script(). In what other way can this be done?

PR #488, opened 2020-05-31
Uplpoad integration yml

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25102

## Description
First commit is a failing unit-test.
Added the unified yml files that were in scripts and integration to directory upload flow.

## Screenshots
We can see the difference between the expected list of uploaded entities to the ones that were actually uploaded.
Comparison Failure (test_upload_pack) 2020-05-31 21-49-41

## Must have
- [x] Tests


demisto_sdk/commands/upload/uploader.py
```diff
@@ -130,11 +130,25 @@ def directory_uploader(self, path: str):
             path (str): Path for directory to upload.
         """
         if is_path_of_integration_directory(path):
+            # Upload unified integration files
+            list_unified_integrations = get_child_files(path)
```
Itay4, line 134: the var name is a bit misleading here, its not really a list of unified integrations right?

demisto_sdk/commands/upload/uploader.py
```diff
@@ -130,11 +130,25 @@ def directory_uploader(self, path: str):
...
             for integration in list_integrations:
                 self.integration_uploader(integration)
 
         elif is_path_of_script_directory(path):
+            # Upload unified scripts files
+            list_unified_scripts = get_child_files(path)
+            for unified_script in list_unified_scripts:
+                file_type = find_type(unified_script)
+                if file_type == 'script':
+                    self.integration_uploader(unified_script)
```
Itay4, line 150: shouldnt this be script_uploader ?

demisto_sdk/commands/upload/tests/uploader_test.py
```diff
@@ -350,14 +349,23 @@ def test_upload_pack(demisto_client_configure, mocker):
...
     status_code = uploader.upload()
+    expected_entities = ['DummyIntegration.yml', 'integration-UploadTest.yml', 'DummyScriptUnified.yml',
+                         'script-DummyScript.yml', 'DummyPlaybook.yml', 'incidenttype-Hello_World_Alert.json',
+                         'incidentfield-Hello_World_ID.json', 'incidentfield-Hello_World_Type.json',
+                         'incidentfield-Hello_World_Status.json', 'classifier-aws_sns_test_classifier.json',
+                         'widget-ActiveIncidentsByRole.json', 'layout-details-test_bla-V2.json',
+                         'upload_test_dashboard.json']
     assert status_code == 0
-    assert len(print.call_args_list) == 19
+    assert len(print.call_args_list) == 21
```
Itay4, line 366: can we avoid this number? can we actually calculate it? it is unclear what is `21`

demisto_sdk/commands/upload/uploader.py
ShahafBenYakir: True

demisto_sdk/commands/upload/uploader.py
ShahafBenYakir: It is. Could be a list of them.

demisto_sdk/commands/upload/tests/uploader_test.py
ShahafBenYakir: I can remove it.

demisto_sdk/commands/upload/uploader.py
Itay4: that func just returns all the files that are under the given dir path, how do you know they are all unified integrations? how can there be more than one?

demisto_sdk/commands/upload/tests/uploader_test.py
Itay4: im ok with having it, just would be better if we can sum up all entities and getting to the number rather than having it hard coded

demisto_sdk/commands/upload/uploader.py
ShahafBenYakir: Consider this:
demisto-sdk  ~:dev:demisto:demisto-sdk  -  :demisto_sdk:tests:test_files:Packs:DummyPack:Scripts:DummyScriptUnified2 yml 2020-06-01 16-05-59

Both `DummyScriptUnified` and `DummyScriptUnified2` could be unified scripts under the `Scripts` directory. I don't know they are unified I can only check if they are integrations and try to upload them. Same goes for integrations.

demisto_sdk/commands/upload/tests/uploader_test.py
ShahafBenYakir: mmm I see what you are saying but I am not sure it is the best thing here since some of the prints here relate to the unification process and stuff like this. so the sum of messages != sum of entities.

demisto_sdk/commands/upload/uploader.py
Itay4: ok but the var name is still misleading, it is not a list of unified entities, but a list of all the files under that dir

demisto_sdk/commands/upload/tests/uploader_test.py
Itay4: ok so lets add comment explaining what that number is

demisto_sdk/commands/upload/uploader.py
ShahafBenYakir: Which I assume to be valid unified integrations/scripts. I can change it to `files` but then IMO it is even more misleading as it is not clear why am I iterating over random files in a directory as opposed to iterating over unified integrations/scripts. What ever you think

PR #489, opened 2020-05-31
Add doc files dir

ronykoz:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24681

## Description
Add doc files dir
Seems that incidentTypes is already there


demisto_sdk/commands/common/constants.py
```diff
@@ -23,6 +23,7 @@
 TOOLS_DIR = 'Tools'
 RELEASE_NOTES_DIR = 'ReleaseNotes'
 TESTS_DIR = 'Tests'
+DOC_FILES_DIR = 'doc_files'
```
reutshal, line 26: Shouldn't this be uppercased?
And why not DocFiles?

ronykoz: @reutshal this is the convention in the repo

PR #494, opened 2020-06-01
Run lint on commonserverpython

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25125

## Must have
- [x] Tests


demisto_sdk/commands/lint/tests/test_linter/gather_facts_test.py
```diff
@@ -38,6 +38,33 @@ def test_not_valid_yaml(self, demisto_content: Callable, create_integration: Cal
...
+        - Repo with CommonServerPython script
+
+        When
+        - Running lint on a repo with a change to CommonServerPython
+
+        Then
+        - Validate that the CommonServerPython is in the file list to check
+        """
+        from demisto_sdk.commands.lint import linter
+        from wcmatch.pathlib import Path
```
Itay4, line 54: is there a reason this is imported in every test?

jochman: That's how it worked before. I can move it up and change this file.

Itay4:
not sure when we stopped running flake8 on commonserver, need to verify it passes so we wont break

demisto_sdk/commands/lint/tests/test_linter/gather_facts_test.py
Itay4: lets move it, it doesnt make sense

jochman:
> not sure when we stopped running flake8 on commonserver, need to verify it passes so we wont break

It will break as commonmsp is full of errors:
image

Itay4:
thats why im saying lets fix content before merging this

PR #502, opened 2020-06-02
Validate - several improvements 

roysagi:
…RP102 to allowed ignored errors, adding some additional prints

## Status
- [x] In Progress
- [x] Ready

## Description
Updating reports schema and adding to schema check, adding IF107 and RP102 to allowed ignored errors, adding some additional prints


roysagi:
> Who verified the report scheme?

Not yet verified. On it.

roysagi:
@bakatzir @yaakovi - If you could please check that my Reports schema is correct.

PR #512, opened 2020-06-04
adding reports to schema check and adding SC100 to ignorable errors

roysagi:
Adding reports to schema check and adding SC100 to ignorable errors.


ronykoz:
@roysagi let's see what happened to the coverage

PR #522, opened 2020-06-08
validate- additional fixes and additions

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Description
 - fixing issue where betaintegration schema was not checked correctly
 - Added schema check for Connections.
 - Fixed an issue where docker was checked on master branch.


demisto_sdk/commands/validate/file_validator.py
```diff
@@ -978,6 +980,7 @@ def is_valid_structure(self):
         """
         if self.validate_all:
             self.validate_all_files(self.skip_conf_json)
+            self.validate_all_files_schema()
```
ronykoz, line 983: let's do scheme first

demisto_sdk/commands/common/hook_validations/incident_field.py
```diff
@@ -232,9 +232,7 @@ def is_valid_name(self):
             for word in name.split():
                 if word.lower() in bad_words:
                     error_message, error_code = Errors.invalid_incident_field_name(word)
-                    formatted_error = self.handle_error(error_message, error_code, file_path=self.file_path,
-                                                        should_print=False)
-                    click.secho(formatted_error, fg="yellow")
+                    click.secho(f"{self.file_path}: [{error_code}] - {error_message}".rstrip("\n") + "\n", fg="yellow")
```
ronykoz, line 235: why? we should fail her

roysagi: This is IF100 which we said should not fail. but if it goes to handle_error it will add it to the report at the end - which it shouldn't. This way it will only print an error.

demisto_sdk/commands/validate/file_validator.py
roysagi: done

PR #523, opened 2020-06-08
non-latest tag should fail the validation

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related issues
fixes: https://github.com/demisto/etc/issues/25365

## Description
non-latest tag should fail the validation


demisto_sdk/commands/common/errors.py
```diff
@@ -421,6 +422,15 @@ def no_docker_tag(docker_image):
     def docker_not_formatted_correctly(docker_image):
         return f'The docker image: {docker_image} is not of format - demisto/image_name:X.X'
 
+    @staticmethod
+    @error_code_decorator
+    def docker_not_on_the_latest_tag(docker_image_tag, docker_image_latest_tag, docker_image_name):
+        return f'The docker image tag is not the latest, please update it.\n' \
```
ronykoz, line 428: please specify it should be the numeric value and not the tag latest

demisto_sdk/commands/common/hook_validations/description.py
```diff
@@ -67,7 +67,7 @@ def is_duplicate_description(self):
         if not re.match(PACKS_INTEGRATION_YML_REGEX, self.file_path, re.IGNORECASE):
             package_path = os.path.dirname(self.file_path)
             try:
-                md_file_path = glob.glob(os.path.join(os.path.dirname(self.file_path), '*_description.md'))[0]
+                md_file_path = glob.glob(os.path.join(self.file_path.split('.')[0], '_description.md'))[0]
```
ronykoz, line 70: why not 
os.path.basename(self.file_path)

demisto_sdk/commands/common/hook_validations/structure.py
```diff
@@ -114,8 +114,9 @@ def is_valid_scheme(self):
             # disabling massages of level INFO and beneath of pykwalify such as: INFO:pykwalify.core:validation.valid
             log = logging.getLogger('pykwalify.core')
             log.setLevel(logging.WARNING)
+            scheme_file_name = 'integration' if self.scheme_name == 'betaintegration' else self.scheme_name
```
ronykoz, line 117: can't we point betaintegation to integration scheme ?

roysagi: this is exactly what is done here. we are pointing the beta to the integration scheme.

demisto_sdk/commands/common/errors.py
roysagi: added, though the lines below tell what should be the changed tag.

demisto_sdk/commands/common/hook_validations/description.py
roysagi: I'm looking for the full path including file name without the extension  - I'll use: os.path.splitext instead

demisto_sdk/commands/validate/file_validator.py
```diff
@@ -100,12 +100,16 @@ def __init__(self, is_backward_check=True, prev_ver=None, use_git=False, only_co
...
+        if self.validate_all:
+            self.skip_docker_checks = True
+
         self.branch_name = ''
         self.use_git = use_git
         self.skip_pack_rn_validation = skip_pack_rn_validation
         if self.use_git:
             print('Using git')
-            self.branch_name = self.get_current_working_branch()
+            self.branch_name = ''
```
ronykoz, line 112: why do we need it then?

demisto_sdk/commands/validate/file_validator.py
```diff
@@ -100,12 +100,16 @@ def __init__(self, is_backward_check=True, prev_ver=None, use_git=False, only_co
                  validate_all=False, is_private_repo=False, skip_pack_rn_validation=False, print_ignored_errors=False,
                  configuration=Configuration()):
         self.validate_all = validate_all
+        self.skip_docker_checks = False
+        if self.validate_all:
+            self.skip_docker_checks = True
```
ronykoz, line 105:
```suggestion
        self.skip_docker_checks = True if self.validate_all else False
```

demisto_sdk/commands/validate/file_validator.py
roysagi: wrong change - fixing

PR #527, opened 2020-06-09
Release 1-1-2

reutshal:
## Status
- [] In Progress
- [X] Ready

## Description
- Fixed an issue with the server version compare

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -485,6 +485,7 @@ def checked_type(file_path, compared_regexes=None, return_regex=False):
 
 def format_version(version):
     """format server version to form X.X.X
+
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -483,6 +483,25 @@ def checked_type(file_path, compared_regexes=None, return_regex=False):
...
+
+    Returns:
+        string.
+        The formatted server version.
+    """
+    formatted_version = version
+    if len(version) == 1:
+        formatted_version = f'{version}.0.0'
+    elif len(version) == 3:
+        formatted_version = f'{version}.0'
```
ronykoz, line 500: we should do split(.)
and do the len on that

PR #546, opened 2020-06-16
fixing bug - testplaybook checked on added files

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Description
TestPlaybooks are being checked on added files

### DIFF
--- a/demisto_sdk/commands/validate/file_validator.py
+++ b/demisto_sdk/commands/validate/file_validator.py
@@ -540,7 +540,7 @@ class FilesValidator:
             print('\nValidating {}'.format(file_path))
             self.check_for_spaces_in_file_name(file_path)
 
-            if re.search(TEST_PLAYBOOK_REGEX, file_path, re.IGNORECASE) and not file_type:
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/validate/file_validator.py
```diff
@@ -540,7 +540,7 @@ def validate_added_files(self, added_files, received_file_type: str = None, modi
             print('\nValidating {}'.format(file_path))
             self.check_for_spaces_in_file_name(file_path)
 
-            if re.search(TEST_PLAYBOOK_REGEX, file_path, re.IGNORECASE) and not file_type:
+            if re.search(TEST_PLAYBOOK_REGEX, file_path, re.IGNORECASE) or not file_type:
```
reutshal, line 543: Actually, do we want to skip every file that we don't find it's file type?

reutshal: Why not just remove the file_type? why do we need it?

roysagi: you are right - made it work the same as in modified files

PR #547, opened 2020-06-16
Tag feed validate

orhovy:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25719

## Description
Added validation of Tag param in feed integrations.

## Must have
- [ ] Tests
- [ ] Documentation


demisto_sdk/commands/common/constants.py
```diff
@@ -876,6 +876,14 @@ class PB_Status:
     }
 ]
 
+FEED_TAG_PARAM = {
+  'display': 'Tags',
+  'hidden': 'false',
+  'name': 'feedTags',
+  'required': False,
+  'type': 0
+}
```
bakatzir, line 885: please ask alex for an `additionalinfo` and add it.

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -673,6 +675,22 @@ def is_valid_fetch(self) -> bool:
 
         return fetch_params_exist
 
+    def is_valid_feed_tag(self) -> bool:
```
bakatzir, line 678: I think this function is not needed, and it can sit under `all_feed_params_exist`

PR #584, opened 2020-07-05
fix path of git sha1 in config.yml

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Must have
- [x] Tests


demisto_sdk/commands/validate/file_validator.py
```diff
@@ -1129,7 +1129,7 @@ def get_content_release_identifier(self):
         except Exception:
             return
         else:
-            return file_content.get('jobs').get('build').get('environment').get('GIT_SHA1')
+            return file_content.get('references').get('environment').get('environment').get('GIT_SHA1')
```
Itay4, line 1132:
```suggestion
            return file_content.get('references', {}).get('environment', {}).get('environment', {}).get('GIT_SHA1')
```

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -954,7 +954,7 @@ def get_content_release_identifier(self):
         except Exception:
             return
         else:
-            return file_content.get('jobs').get('build').get('environment').get('GIT_SHA1')
+            return file_content.get('references').get('environment').get('environment').get('GIT_SHA1')
```
Itay4, line 957: we shouldnt have 2 funcs that do the exact same thing.. lets use the same one

demisto_sdk/commands/validate/file_validator.py
jochman: Don't you think it'll be better if it actually fails if it fails in this step?

Itay4: on `NoneType has no get`? no..

jochman: yeah, but it means that it got the circle.yml as it needs to then the key is wrong.

jochman: more likely is that we will encounter it before it'll reach any customer

Itay4: doesnt matter.. if something is wrong, need to have an indicative error message

PR #586, opened 2020-07-05
Parallelize Circle Build

avidan-H:
## Status
- [x] Ready

## Description
Parallelizes the demisto-sdk circleci build

## Screenshots
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -198,6 +198,7 @@ workflows:
       - deploy:
           <<: *tag_filter
           requires:
+            - tox3-7
             - tox3-8
             - precommit-checks
             - coveralls-check
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,11 @@ jobs:
               tox -e py37 -v -- --cov=demisto_sdk --cov-report=html
         - store_artifacts:
             path: coverage_html_report
+        - persist_to_workspace:
+            root: ~/project
+            paths:
+              - coverage_html_report
+              - .coverage
         - save_cache:
             paths:
               - .tox
@@ -72,13 +77,13 @@ jobs:
               pre-commit --version
               pre-commit run -a
               deactivate
-        - store_artifacts:
-            path: coverage_html_report
-  coveralls-check:
+  coveralls-upload:
       docker:
         - image: python:3.8
       steps:
         - checkout
+        - attach_workspace:
+            at: ~/project
         - run:
             name: Coveralls upload
             command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -93,16 +158,50 @@ jobs:
...
+      - create-id-set:
+          <<: *tag_filter
+          <<: *reqs_tox
+      - create-content-artifacts:
+          <<: *tag_filter
+          <<: *reqs_tox
+      - deploy:
+          <<: *tag_filter
+          requires:
+            - tox3-7
```
Itay4, line 201: remove

Itay4: bring it back !!!

avidan-H:
@Itay4 fixed the coveralls issue. Need to change the branch requirements in the repo's settings so that they don't require the `toxify` job to pass (since that isn't a job defined in the parallelized CircleCI config).

avidan-H:
### Updated Parallel Workflow
image

avidan-H:
One thing that I haven't looked at is the daily report that is sent to the content team channel after the `demisto-sdk`'s nightly is run. I need to look and see if this affects that - I imagine it might.

.circleci/config.yml
```diff
@@ -2,36 +2,88 @@
...
         - run:
             name: Pre-commit
             command: |
               . .tox/py37/bin/activate
               pre-commit --version
               pre-commit run -a
               deactivate
-        - store_artifacts:
-            path: coverage_html_report
+  coveralls-upload:
```
ronykoz, line 80: why not to upload the coveralls in the tox-3-7

.circleci/config.yml
```diff
@@ -41,6 +93,12 @@ jobs:
               else
                 echo "Skipping coveralls"
               fi
+  validate-files:
+      docker:
+        - image: python:3.8
+      steps:
+        - checkout
+        - *restore_cache
```
ronykoz, line 101: why the `*`

avidan-H: It injects the alias - https://circleci.com/docs/2.0/writing-yaml/.

.circleci/config.yml
avidan-H: That is an option that I considered last night when I was wrestling with it a bit... I did spend an inordinate amount of time last night getting it to work in a separate job though :P - but more seriously, since there are 4 concurrent jobs that are waiting on the `tox3-7` step to complete, I don't see a reason not to have that job complete sooner, and have the `coveralls` as a job that will run concurrently with those others.

PR #590, opened 2020-07-06
fixing yml script write bug

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Description
fixing the way yml are being rewritten by content creator

### DIFF
--- a/demisto_sdk/commands/create_artifacts/content_creator.py
+++ b/demisto_sdk/commands/create_artifacts/content_creator.py
@@ -543,7 +543,7 @@ class ContentCreator:
                                 continue
                             unifier = Unifier(package_dir, dir_name, dest_dir)
 
-                            if parse_version(unifier.yml_data.get('fromversion', '0.0.0')) <= parse_version('6.0.0'):
+                            if parse_version(unifier.yml_data.get('toversion', '99.99.99')) >= parse_version('6.0.0'):
                                 new_file_paths = unifier.merge_script_package_to_yml(
                                     file_name_suffix=self.file_name_suffix)
                                 for new_file_path in new_file_paths:
--- a/demisto_sdk/commands/create_artifacts/content_creator.py
+++ b/demisto_sdk/commands/create_artifacts/content_creator.py
@@ -122,8 +122,12 @@ class ContentCreator:
         if self.no_fromversion:
             return {}
 
+        ryaml = YAML()
+        ryaml.preserve_quotes = True
+        ryaml.width = 50000  # make sure long lines will not break (relevant for code section)
         if not yml_content:
-            yml_content = get_yaml(file_path)
+            with open(file_path, 'r') as yml_file:
+                yml_content = ryaml.load(yml_file)
 
         if parse_version(yml_content.get('toversion', '99.99.99')) > parse_version(
                 LATEST_SUPPORTED_VERSION) > parse_version(yml_content.get('fromversion', '0.0.0')):
@@ -135,9 +139,6 @@ class ContentCreator:
             self.fix_script_in_unified_yml(yml_content)
 
             if save_yml:
-                ryaml = YAML()
-                ryaml.preserve_quotes = True
-                ryaml.width = 50000  # make sure long lines will not break (relevant for code section)
                 with open(file_path, mode='w', encoding='utf-8') as f:
                     ryaml.dump(yml_content, f)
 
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/create_artifacts/content_creator.py
```diff
@@ -119,9 +129,11 @@ def add_from_version_to_yml(self, file_path=None, yml_content=None, save_yml=Tru
                 LATEST_SUPPORTED_VERSION) > parse_version(yml_content.get('fromversion', '0.0.0')):
             yml_content['fromversion'] = LATEST_SUPPORTED_VERSION
 
+            self.fix_script_in_unified_yml(yml_content)
+
             if save_yml:
-                with open(file_path, 'w') as f:
-                    yaml.dump(yml_content, f)
+                with open(file_path, mode='w', encoding='utf-8') as f:
+                    YAML().dump(yml_content, f)
```
ronykoz, line 136: why do we need this change

roysagi: According to @yaakovi this is a better way to write a yaml since it works alongside the FoldedScalarString function that prevents the change in the script.

demisto_sdk/commands/create_artifacts/content_creator.py
```diff
@@ -432,7 +455,8 @@ def copy_test_files(self, test_playbooks_dir=TEST_PLAYBOOKS_DIR):
                         new_file_path = self.add_suffix_to_file_path(os.path.join(self.test_bundle,
                                                                                   os.path.basename(new_path)))
                         shutil.copyfile(new_path, new_file_path)
-                        self.add_from_version_to_yml(new_file_path)
+                        if new_file_path.endswith('yml'):
```
yaakovi, line 458: good safeguard.
did you encounter a file in `TestPlaybook` folder which wasn't a YML file?
if so, was it script related?

demisto_sdk/commands/create_artifacts/content_creator.py
```diff
@@ -518,7 +543,7 @@ def copy_packs_content_to_packs_bundle(self, packs):
                                 continue
                             unifier = Unifier(package_dir, dir_name, dest_dir)
 
-                            if parse_version(unifier.yml_data.get('fromversion', '0.0.0')) >= parse_version('6.0.0'):
+                            if parse_version(unifier.yml_data.get('fromversion', '0.0.0')) <= parse_version('6.0.0'):
```
yaakovi, line 546: shouldn't the filter for packs be related to `toversion` instead of `fromversion`?

demisto_sdk/commands/create_artifacts/content_creator.py
```diff
@@ -108,35 +109,57 @@ def run(self):
...
+            json_content = ujson.load(f)
 
         if parse_version(json_content.get('toVersion', '99.99.99')) > parse_version(
                 LATEST_SUPPORTED_VERSION) > parse_version(json_content.get('fromVersion', '0.0.0')):
             json_content['fromVersion'] = LATEST_SUPPORTED_VERSION
 
+            # ujson lets you keep html chars as unicode like "<" should be "\u003c"
             with open(file_path, 'w') as f:
-                json.dump(json_content, f, indent=4)
+                ujson.dump(json_content, f, indent=4, encode_html_chars=True, escape_forward_slashes=False,
```
yaakovi, line 161: this change is also relevant to format command, please open an issue.

roysagi: done https://github.com/demisto/etc/issues/26423

demisto_sdk/commands/create_artifacts/content_creator.py
roysagi: I am not sure if I encountered any other files, I mostly wrote this to be sure of the functionality of this method.

demisto_sdk/commands/create_artifacts/content_creator.py
roysagi: Correct me if I am wrong but the idea is that content entities added in 6.0.0 should no be added to content_new.zip as the are "self-releasing" hence it should be that they are :`fromversion: 6.0.0`

yaakovi: but what about content entities that are already exists?
for example, Integration with `fromversion: 5.0.0` should be included in the packs.

roysagi: done

PR #595, opened 2020-07-08
set integration dependency of indicator types as optional

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
all integrations dependencies of indicator type are marked as optional.
need to add tests

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/find_dependencies/find_dependencies.py
+++ b/demisto_sdk/commands/find_dependencies/find_dependencies.py
@@ -238,7 +238,7 @@ class PackDependencies:
     @staticmethod
     def _label_as_optional(pack_ids):
         """
-        Sets pack as mandatory.
+        Sets pack as optional.
 
         Args:
             pack_ids (set): collection of pack ids to set as optional.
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -235,6 +235,20 @@ def _label_as_mandatory(pack_ids):
         """
         return [(p, True) for p in pack_ids]
 
+    @staticmethod
+    def _label_as_optional(pack_ids):
+        """
+        Sets pack as mandatory.
```
yuvalbenshalom, line 241:
```suggestion
        Sets pack as optional.
```

yaakovi:
> I am not sure her. dose it make sense to set indicators dependencies as optional?
> and also, dose this resolve the problems we see in the dependencies? what about the ip/url/domain commands?

1) this change makes only the integration-based dependencies of indicator type as optional. the script-based dependencies remain mandatory.
2) run it on CommonTypes pack, as the result was as expected. I can run it on more packs that include `Indicator Type`.
3) if a command is implemented by several integrations, the integrations are marked as an optional dependency. see `_detect_generic_commands_dependencies` function.

PR #600, opened 2020-07-15
Updated playbook dependencies optional branch

yaakovi:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26065

## Description
when a task is marked with `skipunavailable`, mark rest playbook branch as an optional dependency.


demisto_sdk/commands/common/constants.py
```diff
@@ -841,7 +841,7 @@ class PB_Status:
 BASE_PACK = "Base"
 NON_SUPPORTED_PACK = "NonSupported"
 DEPRECATED_CONTENT_PACK = "DeprecatedContent"
-IGNORED_DEPENDENCY_CALCULATION = {BASE_PACK, NON_SUPPORTED_PACK, DEPRECATED_CONTENT_PACK}
+IGNORED_DEPENDENCY_CALCULATION = {BASE_PACK, NON_SUPPORTED_PACK, }
```
yaakovi, line 844: Temporary for local testing - will be reverted

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -136,6 +137,60 @@ def get_changed_files(files_string):
     return added_files_list, modified_files_list, added_script_list, modified_script_list
 
 
+def build_dependency_graph(playbook_data):
```
IkaDemisto, line 140: Function name can be confusing. I think tasks graph is more suitable here

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -136,6 +137,60 @@ def get_changed_files(files_string):
...
+        for leaf in leaf_nodes:
+            leaf_task = tasks.get(leaf)
+            leaf_mandatory = graph.nodes[leaf]['mandatory']
+
+            # In this case the playbook is invalid, one of the "nexttasks" fields contain invalid task id.
+            if not leaf_task:
+                print_error(f'{playbook_data.get("id")}: No such task {leaf} in playbook')
+                continue
+                # raise ValueError(f'No such task {leaf} in playbook {playbook_data.get("id")}')
+            leaf_next_tasks = sum(leaf_task.get('nexttasks', {}).values(), [])
```
IkaDemisto, line 170: Not sure why you need to use sum on `nexttasks`

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -136,6 +137,60 @@ def get_changed_files(files_string):
...
+                    # raise ValueError(f'No such task {leaf} in playbook {playbook_data.get("id")}')
+
+                # If task can't be skipped and predecessor task is mandatory - set as mandatory.
+                mandatory = leaf_mandatory and not task.get('skipunavailable', False)
+                if task_id not in graph.nodes():
+                    graph.add_node(task_id, mandatory=mandatory)
+                else:
+                    # If task already in graph, update mandatory field.
+                    # If one of the paths to the task is mandatory - set as mandatory.
+                    graph.nodes[task_id]['mandatory'] = graph.nodes[task_id]['mandatory'] or mandatory
```
IkaDemisto, line 186: Should not it be just:
```
graph.nodes[task_id]['mandatory'] = mandatory
```
?

demisto_sdk/commands/common/update_id_set.py
yaakovi: in conditional tasks, there might be several lists:
image

demisto_sdk/commands/common/update_id_set.py
yaakovi: just to be on the safe side,
if I first recognized it as mandatory and then I found another task that is optional, I don't want to override the mandatory mark. this way I can maintain the `mandatory: true` once I found it.

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -137,6 +138,60 @@ def get_changed_files(files_string):
...
+
+        for leaf in leaf_nodes:
+            leaf_task = tasks.get(leaf)
+            leaf_mandatory = graph.nodes[leaf]['mandatory']
+
+            # In this case the playbook is invalid, one of the "nexttasks" fields contain invalid task id.
+            if not leaf_task:
+                print_error(f'{playbook_data.get("id")}: No such task {leaf} in playbook')
+                continue
+                # raise ValueError(f'No such task {leaf} in playbook {playbook_data.get("id")}')
```
reutshal, line 170: Remove this comment if not needed

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -137,6 +138,60 @@ def get_changed_files(files_string):
...
+            leaf_next_tasks = sum(leaf_task.get('nexttasks', {}).values(), [])
+
+            for task_id in leaf_next_tasks:
+                task = tasks.get(task_id)
+                if not task:
+                    print_error(f'{playbook_data.get("id")}: No such task {leaf} in playbook')
+                    continue
+                    # raise ValueError(f'No such task {leaf} in playbook {playbook_data.get("id")}')
```
reutshal, line 178: Same

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -250,22 +270,26 @@ def _label_as_optional(pack_ids):
...
         Returns:
             set: dependencies data that includes pack id and whether is mandatory or not.
 
         """
         dependencies_packs = set()
+        verbose_file.write('\n### Scripts')
 
         for script_mapping in pack_scripts:
             script = next(iter(script_mapping.values()))
+            script_dep = set()
```
reutshal, line 291:
```suggestion
            script_dependencies = set()
```

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -275,99 +299,153 @@ def _collect_scripts_dependencies(pack_scripts, id_set):
 
                 if pack_name:  # found script dependency implementing pack name
                     pack_dependencies_data = PackDependencies._label_as_mandatory(pack_name)
-                    dependencies_packs.update(pack_dependencies_data)  # set found script as mandatory
+                    script_dep.update(pack_dependencies_data)  # set found script as mandatory
```
reutshal, line 302:
```suggestion
                    script_dependencies.update(pack_dependencies_data)  # set found script as mandatory
```

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -275,99 +299,153 @@ def _collect_scripts_dependencies(pack_scripts, id_set):
...
                     continue  # found dependency in script section, skipping to next depends on element
 
                 # try to search dependency by integration integration
                 pack_names = PackDependencies._search_packs_by_integration_command(command, id_set)
 
                 if pack_names:  # found integration dependency implementing pack name
                     pack_dependencies_data = PackDependencies._detect_generic_commands_dependencies(pack_names)
-                    dependencies_packs.update(pack_dependencies_data)
+                    script_dep.update(pack_dependencies_data)
```
reutshal, line 310:
```suggestion
                    script_dependencies.update(pack_dependencies_data)
```

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -275,99 +299,153 @@ def _collect_scripts_dependencies(pack_scripts, id_set):
...
+
+            verbose_file.write(
+                f'{os.path.basename(script.get("file_path", ""))} depends on: '
+                f'{script_dep}  '
```
reutshal, line 314:
```suggestion
                f'{script_dependencies}  '
```

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -275,99 +299,153 @@ def _collect_scripts_dependencies(pack_scripts, id_set):
...
+            )
+            dependencies_packs.update(script_dep)
```
reutshal, line 316:
```suggestion
            dependencies_packs.update(script_dependencies)
```

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -385,36 +463,45 @@ def _collect_layouts_dependencies(pack_layouts, id_set):
...
 
-            related_incident_types = incident_field_data.get('incident_types', [])
-            packs_found_from_incident_types = PackDependencies._search_packs_by_items_names(
-                related_incident_types, id_set['IncidentTypes'])
-
-            if packs_found_from_incident_types:
-                pack_dependencies_data = PackDependencies. \
-                    _label_as_mandatory(packs_found_from_incident_types)
-                dependencies_packs.update(pack_dependencies_data)
+            # related_incident_types = incident_field_data.get('incident_types', [])
```
reutshal, line 497: Why did you comment this section?

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -717,22 +891,24 @@ def build_dependency_graph(pack_id, id_set):
...
     @staticmethod
-    def find_dependencies(pack_name, id_set_path=None):
+    def find_dependencies(pack_name, id_set_path=None, debug_file_path=None):
         """
         Main function for dependencies search and pack metadata update.
 
         Args:
             pack_name (str): pack id, currently pack folder name is in use.
             id_set_path (str): id set json.
+            debug_file_path (str): path to dependency explanations file.
```
reutshal, line 901: Doesn't mypy return error here?
id_set_path, debug_file_path are of type `str` while their default value is None.
Consider changing their default to `''`

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -717,22 +891,24 @@ def build_dependency_graph(pack_id, id_set):
...
         """
         if not id_set_path:
-            id_set = IDSetCreator(output=None, print_logs=False).create_id_set()
+            id_set = IDSetCreator(print_logs=False).create_id_set()
         else:
             with open(id_set_path, 'r') as id_set_file:
                 id_set = json.load(id_set_file)
-
-        dependency_graph = PackDependencies.build_dependency_graph(pack_id=pack_name, id_set=id_set)
+        with VerboseFile(debug_file_path) as verbose_file:
```
reutshal, line 909: What happens if the `debug_file_path` does not exist?

demisto_sdk/commands/find_dependencies/find_dependencies.py
IkaDemisto: +1

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -275,99 +299,153 @@ def _collect_scripts_dependencies(pack_scripts, id_set):
...
             pack_playbooks (list): collection of pack playbooks data.
             id_set (dict): id set json.
+            verbose_file (VerboseFile): path to dependency explanations file.
 
         Returns:
             set: dependencies data that includes pack id and whether is mandatory or not.
 
         """
         dependencies_packs = set()
+        verbose_file.write('\n### Playbooks')
```
IkaDemisto, line 367: Good idea. I think in later step the log should be extended and contain the reason and additional data for found dependency.

demisto_sdk/commands/find_dependencies/find_dependencies.py
yaakovi: after speaking with @michalgold we agreed that this dependency is not correct.
fields should be based on types as they can exist without it.

demisto_sdk/commands/find_dependencies/find_dependencies.py
yaakovi: the object ignores all the prints, see the implementation at the top of the file

demisto_sdk/commands/find_dependencies/find_dependencies.py
yaakovi: what do you mean later step?
this is output example:

image

IkaDemisto: I mean that if layout from pack A uses specific field from pack B then better to log it somewhere. But it is not urgent.
For example: https://panw-global.slack.com/archives/G011E63JXPB/p1594907099182700

PR #602, opened 2020-07-15
fixing content creator logic

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Description
there are 3 bundles tests, content and packs
file with fromversion below 6.0.0  should be in both content and packs
file with fromversion above 6.0.0  should only be in packs
file with toversion 5.9.9 or below should only be in content
content and tests bundle act the same

### DIFF
--- a/demisto_sdk/commands/create_artifacts/content_creator.py
+++ b/demisto_sdk/commands/create_artifacts/content_creator.py
@@ -174,12 +174,13 @@ class ContentCreator:
             bundle (str): the bundle being created
 
         Returns:
-            bool. True if the file should be added to the bundle (above 6.0.0 and in packs bundle or exists below 6.0.0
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/create_artifacts/content_creator.py
```diff
@@ -166,31 +166,55 @@ def add_from_version_to_json(self, file_path):
...
+    def should_process_file_to_bundle(self, file_path, bundle):
+        """
+
+        Args:
+            file_path (str): the file_path being processed
+            bundle (str): the bundle being created
+
+        Returns:
+            bool. True if the file should be added to the bundle (above 6.0.0 and in packs bundle or exists below 6.0.0
+            and in content bundle or in test bundle both above and below 6.0.0)
```
ronykoz, line 178: separate to bullets, super confusing

PR #608, opened 2020-07-19
Validate command release branch fix

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Description
Validate should always pass when running using git on release branch.


demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -716,6 +722,10 @@ def setup_git_params(self):
             self.compare_type = '..'
             self.prev_ver = get_content_release_identifier(self.branch_name)
 
+            # when running against git while on release branch - show errors but don't fail the validation
+            if self.branch_name.startswith('19.') or self.branch_name.startswith('20.'):
```
yaakovi, line 726:
```suggestion
            if self.branch_name.startswith('20.'):
```

PR #623, opened 2020-07-26
handle wrong pack name in find dependencies

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/26678

## Description
print an error message in case of misspelled pack name

## Screenshots
image

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -869,8 +869,15 @@ def find_dependencies_command(**kwargs):
     id_set_path = kwargs.get('id_set_path')
     verbose = kwargs.get('verbose')
     update_pack_metadata = not kwargs.get('no_update')
-    PackDependencies.find_dependencies(pack_name=pack_name, id_set_path=id_set_path, debug_file_path=verbose,
-                                       update_pack_metadata=update_pack_metadata)
+
+    try:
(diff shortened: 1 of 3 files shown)


demisto_sdk/__main__.py
```diff
@@ -869,8 +869,15 @@ def find_dependencies_command(**kwargs):
     id_set_path = kwargs.get('id_set_path')
     verbose = kwargs.get('verbose')
     update_pack_metadata = not kwargs.get('no_update')
-    PackDependencies.find_dependencies(pack_name=pack_name, id_set_path=id_set_path, debug_file_path=verbose,
-                                       update_pack_metadata=update_pack_metadata)
+
+    try:
+        PackDependencies.find_dependencies(pack_name=pack_name,
```
IkaDemisto, line 874: The root cause for it is not `glob.glob` in find_pack_path function?

yaakovi: when passing invalid pack name, the function `_collect_pack_items` returns empty lists.
then `_find_pack_dependencies` doesn't find any dependencies, causing the graph to contain only the original (invalid) pack name.
therefore the original response was empty dict.

PR #637, opened 2020-08-03
Merge id sets

anara123:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/24216

## Description
Merge two id-sets. We will use it to merge public and private id-sets and find duplicates.

## Screenshots

## Must have
- [x] Tests
- [ ] Documentation


demisto_sdk/commands/common/update_id_set.py
```diff
@@ -1242,6 +1244,94 @@ def get_general_paths(path):
...
+    duplicates = []
+    united_id_set = IDSet(copy.deepcopy(first_id_set_dict))
+
+    first_id_set = IDSet(first_id_set_dict)
+    second_id_set = IDSet(second_id_set_dict)
+
+    for object_type, object_list in second_id_set.get_dict().items():
+        for obj in object_list:
+            obj_id = list(obj.keys())[0]
+            is_duplicate = has_duplicate(first_id_set.get_list(object_type), obj_id, object_type, print_logs,
```
yaakovi, line 1321: calculate the `first_id_set.get_list(object_type)` expression outside the inner loop

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -1242,6 +1244,94 @@ def get_general_paths(path):
...
+    unified_id_set, has_duplicates, duplicates = merge_id_sets(first_id_set, second_id_set, print_logs)
+
+    if unified_id_set:
+        with open(output_id_set_path, mode='w', encoding='utf-8') as f:
+            json.dump(unified_id_set.get_dict(), f, indent=4)
+
+    return unified_id_set, has_duplicates, duplicates
+
+
+def merge_id_sets(first_id_set_dict: dict, second_id_set_dict: dict, print_logs: bool = True):
```
yaakovi, line 1307: why not join both id sets and then use the original `has_duplicate` function?

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -1445,12 +1535,25 @@ def find_duplicates(id_set, print_logs):
...
+
+    if external_object and len(duplicates) == 0:
+        return False
 
-    if len(duplicates) < 2:
+    if not external_object and len(duplicates) == 1:
         return False
 
+    if external_object:
+        duplicates.append(external_object)
```
yaakovi, line 1555: you can move this right after the list-comprehension and keep the original condition (instead of the above two)

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -1242,6 +1244,94 @@ def get_general_paths(path):
...
+            is_duplicate = has_duplicate(first_id_set.get_list(object_type), obj_id, object_type, print_logs,
+                                         external_object=obj)
+            if is_duplicate:
+                has_duplicates = True
+                duplicates.append(obj_id)
+            else:
+                united_id_set.add_to_list(object_type, obj)
+
+    if has_duplicates:
+        return None, has_duplicates, duplicates
```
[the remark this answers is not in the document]
anara123: correct, I can delete the boolean.

demisto_sdk/commands/common/update_id_set.py
anara123: I remember we had some duplicates in our id_sets. So I wanted to avoid the noice. To represent only the duplicates that were added from one of them. But if we don't have these duplicates anymore, then I can merge them

yaakovi: ok, then leave it as is

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -1242,41 +1242,41 @@ def merge_id_sets_from_files(first_id_set_path, second_id_set_path, output_id_se
...
                                          external_object=obj)
             if is_duplicate:
-                has_duplicates = True
                 duplicates.append(obj_id)
             else:
                 united_id_set.add_to_list(object_type, obj)
 
-    if has_duplicates:
-        return None, has_duplicates, duplicates
+    if len(duplicates) > 0:
```
yaakovi, line 1276:
```suggestion
    if duplicates:
```

PR #643, opened 2020-08-04
Feed fetch default true

orhovy:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
regarding:
https://github.com/demisto/etc/issues/27176

## Description
edited the required params so feed integrations will require fetch true in default value.

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


demisto_sdk/commands/common/constants.py
```diff
@@ -854,6 +854,7 @@ class PB_Status:
 
 FEED_REQUIRED_PARAMS = [
     {
+        'defaultvalue': 'true',
```
yaakovi, line 857: shouldn't this be a boolean?

orhovy:
@yaakovi 
ran validate already on all repo and updated the failures.
PR: https://github.com/demisto/content/pull/8367

demisto_sdk/commands/common/constants.py
orhovy: In some integration it
defaultvalue: true
and in some its
defaultvalue: 'true'
Both work.
I checked how the majority looks and its that way. defaultvalue: 'true'

yaakovi: seems odd that the default value is string when the type of the parameter is boolean but ok...
if that's the majority then let's keep it that way.

orhovy:
image
ran on master and validated all files

PR #651, opened 2020-08-09
Init Templates and bug fix

Bargenish:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: demisto/etc/issues/27060
fixes: demisto/etc/issues/24011

## Description
Fixed an issue where the **Init** command failed on scripts.
Changed the init templates source to the remote repo.
init will use local templates only if the remote source is not reachable.

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/common/tests/tools_test.py
+++ b/demisto_sdk/commands/common/tests/tools_test.py
@@ -152,6 +152,15 @@ class TestGetRemoteFile:
                                                return_content=True)
         assert hello_world_py
 
+    def test_get_remote_file_content(self):
+        hello_world_py = tools.get_remote_file('Packs/HelloWorld/Integrations/HelloWorld/HelloWorld.py',
+                                               return_content=True)
+        hello_world_text = hello_world_py.decode()
+        assert isinstance(hello_world_py, bytes)
+        assert hello_world_py
+        assert 'main()' in hello_world_text
+        assert hello_world_text.startswith('"""HelloWorld Integration for Cortex XSOAR (aka Demisto)')
+
     def test_get_remote_file_origin(self):
         hello_world_yml = tools.get_remote_file('Packs/HelloWorld/Integrations/HelloWorld/HelloWorld.yml', 'master')
         assert hello_world_yml
--- a/demisto_sdk/commands/common/tests/tools_test.py
+++ b/demisto_sdk/commands/common/tests/tools_test.py
@@ -147,6 +147,11 @@ class TestGetRemoteFile:
         assert hello_world_yml
         assert hello_world_yml['commonfields']['id'] == 'HelloWorld'
 
+    def test_get_remote_file_content_sanity(self):
+        hello_world_py = tools.get_remote_file('Packs/HelloWorld/Integrations/HelloWorld/HelloWorld.py',
+                                               return_content=True)
+        assert hello_world_py
+
     def test_get_remote_file_origin(self):
         hello_world_yml = tools.get_remote_file('Packs/HelloWorld/Integrations/HelloWorld/HelloWorld.yml', 'master')
         assert hello_world_yml
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -145,6 +145,15 @@ def run_command(command, is_silenced=True, exit_on_error=True, cwd=None):
 
 
 def get_remote_file(full_file_path, tag='master', return_content=False):
+    """
+    Args:
+        full_file_path (string):The full path of the file.
+        tag (string): The branch name. default is 'master'
+        return_content (bool): Determines whether to return the file's raw content or the dict representation of it.
+    Returns:
+        The file content in the required format.
+
+    """
     # 'origin/' prefix is used to compared with remote branches but it is not a part of the github url.
     tag = tag.lstrip('origin/')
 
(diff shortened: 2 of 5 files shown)


demisto_sdk/commands/init/initiator.py
```diff
@@ -95,6 +96,13 @@ def __init__(self, output: str, name: str = '', id: str = '', integration: bool
 
     HELLO_WORLD_INTEGRATION = 'HelloWorld'
     HELLO_WORLD_SCRIPT = 'HelloWorldScript'
+    HELLO_WORLD_SCRIPT_FILES = ['HelloWorldScript.py', 'HelloWorldScript.yml', 'HelloWorldScript_test.py']
+    HELLO_WORLD_INTEGRATION_FILES = ['HelloWorld.py', 'HelloWorld.yml', 'HelloWorld_description.md',
+                                     'HelloWorld_image.png', 'HelloWorld_test.py', 'Pipfile', 'Pipfile.lock',
+                                     'test_data/domain_reputation.json', 'test_data/get_alert.json',
+                                     'test_data/ip_reputation.json', 'test_data/scan_results.json',
+                                     'test_data/search_alerts.json', 'test_data/update_alert_status.json']
```
yaakovi, line 104: use `os.path.join` for all the files under `test_data`. not sure if this will work on windows machines (you can ask someone from the security team for help testing it).

demisto_sdk/commands/init/initiator.py
```diff
@@ -627,3 +635,28 @@ def copy_demistotmock(self):
...
+        else:
+            path = os.path.join('Packs', 'HelloWorld', 'Scripts', 'HelloWorldScript')
+
+        for file in files_list:
+            try:
+                file_content = tools.get_remote_file(os.path.join(path, file), return_content=True)
+                with open(os.path.join(self.full_output_path, file), 'wb') as f:
+                    f.write(file_content)
+            except Exception:
+                print(f"Could not fetch remote template - {file}. Using local templates instead.")
```
yaakovi, line 659: also, specify the error message
```suggestion
                print_warning(f"Could not fetch remote template - {file}. Using local templates instead.")
```

demisto_sdk/commands/init/initiator.py
```diff
@@ -627,3 +635,28 @@ def copy_demistotmock(self):
...
+        Args:
+            files_list: List of files to download.
+        Returns:
+            bool. True if the files were downloaded and saved successfully, False otherwise.
+        """
+        if self.is_integration:
+            path = os.path.join('Packs', 'HelloWorld', 'Integrations', 'HelloWorld')
+            os.mkdir(os.path.join(self.full_output_path, self.TEST_DATA_DIR))
+        else:
+            path = os.path.join('Packs', 'HelloWorld', 'Scripts', 'HelloWorldScript')
```
bakatzir, line 651: don't we want to create a test dir for scripts?

bakatzir, line 651: don't we want to create a test dir for scripts?

demisto_sdk/commands/init/tests/initiator_test.py
```diff
@@ -305,3 +306,54 @@ def test_format_pack_dir_name(initiator, input_name, expected_output_name):
...
+    with open(file_path, 'r') as f:
+        file_content = f.read()
+
+    assert res
+    assert "Test im in file" in file_content
+
+    os.remove(os.path.join(PACK_NAME, 'Test.py'))
+    os.rmdir(PACK_NAME)
+
+
```
bakatzir, line 338: let's add a test case for the creation of *_test.py files for integration and scripts.

demisto_sdk/commands/common/tools.py
```diff
@@ -144,7 +144,7 @@ def run_command(command, is_silenced=True, exit_on_error=True, cwd=None):
     return output
 
 
-def get_remote_file(full_file_path, tag='master'):
+def get_remote_file(full_file_path, tag='master', return_content=False):
```
bakatzir, line 147: `return_content` returns the raw content of the file and not the dict representation of it?
let's add docstring to this func.

demisto_sdk/commands/common/tests/tools_test.py
```diff
@@ -147,6 +147,11 @@ def test_get_remote_file_sanity(self):
         assert hello_world_yml
         assert hello_world_yml['commonfields']['id'] == 'HelloWorld'
 
+    def test_get_remote_file_content_sanity(self):
```
bakatzir, line 150: can we add 1 more test case(non sanity) with an example of how the `file.content` looks like?

demisto_sdk/commands/init/initiator.py
yaakovi: the current template doesn't include one and most scripts don't have test files so I don't think it is necessary

demisto_sdk/commands/init/initiator.py
Bargenish: the relevant exception is being printed as part of `get_remote_file` function.
`get_remote_file` returns an empty dict if an error has occurred so the error that the above function raises would be `a bytes-like object is required, not 'dict'`.
Should I add it anyway

demisto_sdk/commands/init/initiator.py
Bargenish: it is not needed. `HelloWorld` pack doesnt have a test_data dir inside the HelloWorldScript directory. Not sure we want to create an empty folder.

PR #659, opened 2020-08-12
Remove Escape Chars

Bargenish:
## Status
- [ ] In Progress
- [X] Ready

## Related Issues
fixes: demisto/etc/issues/27200
## Description
removed `_` and `-` from special markdown chars regex to avoid adding a `\` beforehand what could break urls.

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


Bargenish:
We use this function for almost every section of the README (for all entities).
The cases in which Markdown looks at those chars as special ones are when a line begins with a dash or if an underscore surrounds an entire word.
It doesn't happen very often that the yaml file contains such cases. Url's are more common.

yaakovi:
> We use this function for almost every section of the README (for all entities).
> The cases in which Markdown looks at those chars as special ones are when a line begins with a dash or if an underscore surrounds an entire word.
> It doesn't happen very often that the yaml file contains such cases. Url's are more common.

but it does have a use...
need to support both.

Bargenish:
@yaakovi Added support for both cases.

demisto_sdk/commands/generate_docs/common.py
```diff
@@ -170,7 +170,20 @@ def string_escape_md(st, minimal_escaping=False, escape_multiline=False, escape_
         for c in '|':
             st = st.replace(c, '\\' + c)
     else:
-        st = "".join(["\\" + str(c) if c in r"\`*_{}[]()#+-!" else str(c) for c in st])
+        st = "".join(f'\\{str(c)}' if c in r"\`*{}[]()#+!" else str(c) for c in st)
+
+        # The following code adds an escape character for '-' and '_' following cases:
+        # 1. The string begins with a dash. e.g: - This input specifies the entry id
+        # 1. The string has a word which wrapped with an underscore in it . e.g: This input _specifies_ the entry id
```
yaakovi, line 177:
```suggestion
        # 2. The string has a word which wrapped with an underscore in it. e.g: This input _specifies_ the entry id
```

yaakovi (approved):
update changelog as well.

PR #664, opened 2020-08-13
Init Contribution Changes

avidan-H:
## Status
- [x] Ready

## Description
Changes init for contribution zip conversion. Namely,
- Author value can be passed when initializing the `Initiator` class to be set in the `pack_metadata.json` file.
- `support` field in `pack_metadata.json` file is set to `community`.
- `email` field in `pack_metadata.json` file is set to the empty string.
- Differentiate between the name set in `pack_metadata.json` file (display name and therefore can contain spaces) and the pack directory name which must be appropriately formatted.

### DIFF
--- a/demisto_sdk/commands/init/initiator.py
+++ b/demisto_sdk/commands/init/initiator.py
@@ -86,7 +86,7 @@ class Initiator:
         self.full_output_path = ''
 
         self.name = name
-        if name is not None and len(name) != 0 and not self.contribution:
+        if name and not self.contribution:
             while ' ' in name:
                 name = str(input("The directory and file name cannot have spaces in it, Enter a different name: "))
 
--- a/demisto_sdk/commands/init/initiator.py
+++ b/demisto_sdk/commands/init/initiator.py
@@ -59,11 +59,12 @@ class Initiator:
            full_output_path (str): The full path to the newly created pack/integration/script
            contribution (str|Nonetype): The path to a contribution zip file
            description (str): Description to attach to a converted contribution pack
+           author (str): Author to ascribe to a pack converted from a contribution zip
     """
 
     def __init__(self, output: str, name: str = '', id: str = '', integration: bool = False, script: bool = False,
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/init/initiator.py
```diff
@@ -84,14 +86,14 @@ def __init__(self, output: str, name: str = '', id: str = '', integration: bool
         self.full_output_path = ''
 
         self.name = name
-        if name is not None and len(name) != 0:
-            if self.contribution:
-                self.name = self.format_pack_dir_name(name)
-            else:
-                while ' ' in name:
-                    name = str(input("The directory and file name cannot have spaces in it, Enter a different name: "))
+        if name is not None and len(name) != 0 and not self.contribution:
```
Itay4, line 89:
```suggestion
        if name and not self.contribution:
```

demisto_sdk/commands/init/initiator.py
```diff
@@ -192,15 +194,22 @@ def convert_contribution_to_pack(self):
...
                     # from contribution metadata
                     metadata_dict['description'] = self.description or metadata.get('description')
-                    metadata_dict['name'] = pack_name
-                    metadata_dict['author'] = metadata.get('author', '')
-                    metadata_dict['support'] = metadata.get('support', '')
+                    metadata_dict['name'] = pack_display_name
+                    metadata_dict['author'] = self.author or metadata.get('author', '')
+                    metadata_dict['support'] = 'community'
                     metadata_dict['url'] = metadata.get('supportDetails', {}).get('url', '')
-                    metadata_dict['email'] = metadata.get('supportDetails', {}).get('email', '')
```
Itay4, line 203: lets not remove the `email` field completely, just initialize it with empty string

avidan-H: The `email` field is initialized with the empty string later in the `create_metadata` method.

orhovy:
@avidan-H Can you please add to CHANGELOG.md

PR #693, opened 2020-08-20
Verbose format

orlichter1:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23412

## Description
Added a verbose option for `demisto-sdk format`. Now if the `-v` flag is not inputed, only the update message shows up and the summary at the end.
In the process also added the following changes (went through @roysagi ):
1. If a file is changed due to format and is saved onto the same path, no `Saving...` message is printed
2. Added a `suppress_print` parameter to validate classes in order to fulfill this
3. Test scripts were not formatted before and would fail the command, added them to be formatted as regular scripts.
4. Noticed that beta integrations were not formatted as well so added them to `UNFORMATTED_FILES` so they won't fail the command and opened a dev task to add a formatter for them.

## Must have
- [x] Tests
- [ ] Documentation


roysagi:
@orlichter1 - 3% coverage decrease seems not likely - maybe try pulling from master

demisto_sdk/commands/format/format_module.py
```diff
@@ -54,20 +55,23 @@
                      'changelog',
                      'image',
                      'javascriptfile',
-                     'powershellfile']
+                     'powershellfile',
+                     'betaintegration']
```
yaakovi, line 59:
```suggestion
                     'betaintegration',
                     ]
```

demisto_sdk/commands/format/update_incidentfields.py
```diff
@@ -17,8 +17,8 @@ class IncidentFieldJSONFormat(BaseUpdateJSON):
             output (str): the desired file name to save the updated version of the JSON to.
     """
 
-    def __init__(self, input: str = '', output: str = '', path: str = '', from_version: str = '', no_validate: bool = False):
-        super().__init__(input, output, path, from_version, no_validate)
+    def __init__(self, input: str = '', output: str = '', path: str = '', from_version: str = '', no_validate: bool = False, verbose: bool = False):
+        super().__init__(input, output, path, from_version, no_validate, verbose)
```
yaakovi, line 21: keyword argument shouldn't be passed as positional argument (relevant for all appearances)

yaakovi (approved):
nice, please attach a screenshot with and without the verbose option for a pack with 3 modified files

demisto_sdk/commands/format/update_pythonfile.py
```diff
@@ -22,18 +22,19 @@ class PythonFileFormat(BaseUpdate):
...
+    def format_py_using_autopep(py_file_path, verbose: bool = False):
         """Run autopep8 formatter on python file.
         Args:
             py_file_path (str): The python file path.
         Returns:
             bool. True if succeed to run autopep8 on file, False otherwise.
         """
-        print(f"\nRunning autopep8 on file: {py_file_path}\n")
+        if verbose:
+            print(f"\nRunning autopep8 on file: {py_file_path}\n")
```
bakatzir, line 37: didn't we want to shy away from the 'print` for click.secho?
https://github.com/demisto/etc/issues/23929
@yaakovi ?
relevant for all occurrences.

yaakovi: wasn't aware of it.
@orlichter1, please update to `click`

PR #694, opened 2020-08-20
update the error message with how to rerun update-rn

roysagi:
## Description
error messages in validate still point to -p flag for update rn command - this fixes that.


amshamah419 (approved):
Looks good. Just make sure to update the changelog.

PR #700, opened 2020-08-24
Catch meta_data join error when path is not found in RN command

amshamah419:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27849

## Description
Error was caused when giving an incomplete path as an input.


roysagi (approved):
looks good

PR #720, opened 2020-08-31
format: support for updating docker image

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
reference: https://github.com/demisto/etc/issues/25903

## Description
Add the option to update the docker image as part of the format command.

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fixed an issue where **update-release-notes** would add duplicate release notes when two packs with the same substring existed in the modified files.
 * Fixed an issue where a non-descriptive error would be returned when giving the **update-release-notes** command a pack which can not be found.
 * Added dependencies check for *widgets* in **find-dependencies** command.
+* Added support for *update-docker* in **format** command.
 
 # 1.1.10
 * Updated the **init** command. Relevant only when passing the *--contribution* argument.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 * Fixed an issue where **update-release-notes** would fail to bump new versions if the feature branch was out of sync with the master branch.
 * Fixed an issue where a non-descriptive error would be returned when giving the **update-release-notes** command a pack which can not be found.
 * Added dependencies check for *widgets* in **find-dependencies** command.
-* Added support for *update-docker* in **format** command.
+* Added a `update-docker` flag to **format** command.
 * Added a `json-to-outputs` flag to the **run** command.
 * Added a verbose (`-v`) flag to **format** command.
 * Fixed an issue where **download** added the prefix "playbook-" to the name of playbooks.
--- a/demisto_sdk/commands/format/README.md
+++ b/demisto_sdk/commands/format/README.md
@@ -27,7 +27,7 @@ When done formatting, the **validate** command will run, to let you know of thin
 
    When no validation on file is needed.
 
-* **-up ,--update-docker**
+* **-ud ,--update-docker**
 
    Set if you want to update the docker image of the integration/script to the newest available tag.
 
--- a/demisto_sdk/commands/format/README.md
+++ b/demisto_sdk/commands/format/README.md
@@ -27,6 +27,10 @@ When done formatting, the **validate** command will run, to let you know of thin
 
    When no validation on file is needed.
 
+* **-up ,--update-docker**
+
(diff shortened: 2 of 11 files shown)


bakatzir (commented):
Please edit also the `latest_docker_error` in errors.py and recommend running `demisto-sdk format -ud`

glicht:
> Please edit also the `latest_docker_error` in errors.py and recommend running `demisto-sdk format -ud`

This is a different error. Running update docker won't help as the newest tag is `latest`. It requires creating a new docker image.

demisto_sdk/commands/format/update_script.py
```diff
@@ -17,16 +20,52 @@ class ScriptYMLFormat(BaseUpdateYML):
...
-    def __init__(self, input: str = '', output: str = '', path: str = '', from_version: str = '', no_validate: bool = False):
+    def __init__(self, input: str = '', output: str = '', path: str = '', from_version: str = '', no_validate: bool = False,
+                 update_docker: bool = False):
         super().__init__(input, output, path, from_version, no_validate)
+        self.update_docker = update_docker
         if not from_version and self.data.get("type") == TYPE_PWSH:
             self.from_version = '5.5.0'
 
+    @staticmethod
+    def update_docker_image_in_script(script_obj: dict, from_version: Optional[str] = None):
```
orhovy, line 31: why making it a static function in scriptYML? I think that adding it to BaseUpdateYML is better as ScriptYMLFormat and IntegrationYMLFormat both inherent from it.

glicht: It's mostly a matter of style regarding when to use static or not. In this case there are also other yml formatters that extend BaseUpdateYML, so it is less fitting.

demisto_sdk/commands/format/update_integration.py
```diff
@@ -11,6 +11,7 @@
                                                           SKIP_RETURN_CODE,
                                                           SUCCESS_RETURN_CODE)
 from demisto_sdk.commands.format.update_generic_yml import BaseUpdateYML
+from demisto_sdk.commands.format.update_script import ScriptYMLFormat
```
bakatzir, line 14: As both the `update_integration` and `update_script` extend the update_generic_yml(BaseUpdateYML)
I think it is better to move the implementation of update_docker_image_in_script there to not create this dependency.

glicht: I felt that it is less fitting as other yml formatters also extend BaseUpdateYML and the logic there should be generic to yml stuff and not scripts/integrations.

bakatzir: I think otherwise, but understand it is not a decisive matter.

bakatzir (approved):

demisto_sdk/commands/format/README.md
```diff
@@ -27,6 +27,10 @@ When done formatting, the **validate** command will run, to let you know of thin
 
    When no validation on file is needed.
 
+* **-up ,--update-docker**
```
roysagi, line 30:
```suggestion
* **-ud ,--update-docker**
```

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Fixed an issue where **update-release-notes** would fail to bump new versions if the feature branch was out of sync with the master branch.
 * Fixed an issue where a non-descriptive error would be returned when giving the **update-release-notes** command a pack which can not be found.
 * Added dependencies check for *widgets* in **find-dependencies** command.
+* Added support for *update-docker* in **format** command.
```
roysagi, line 15:
```suggestion
* Added a `update-docker` flag to **format** command.
```

for changelog consistency

demisto_sdk/commands/format/update_integration.py
roysagi: I agree with @bakatzir here - we have some functions in the BaseUpdateYML that are used only by part of the classes inheriting from it (i.e. not part of the` update_yml` generic function).

For instance `update_tests` - is used by playbook formatter but not the test playbook formatter.

I believe this is the same case here as it might be confusing why should the docker be checked through the unrelated script formatter.

glicht: @roysagi not sure I am fully following. I'll merge and feel free to do later on any design changes as you see fitting.

PR #723, opened 2020-08-31
format cmd add missing & remove extra fields

darkushin:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27794
fixes: https://github.com/demisto/etc/issues/28042

## Description
- Remove excessive fields in files that were downloaded from server 6.0 and add missing fields that are required.
- Fix indentation error in json files when running format command.

## Must have
- [x] Tests
- [x] Documentation


guyfreund:
@darkushin regarding https://github.com/demisto/etc/issues/28042, I suggest to work here with dynamic json indentation and not a constant number, see `demisto-sdk download` for reference.

demisto_sdk/commands/format/update_playbook.py
```diff
@@ -22,7 +22,19 @@ def __init__(self, input: str = '', output: str = '', path: str = '', from_versi
...
         """Add empty description to playbook and tasks."""
         if self.verbose:
-            click.echo('Adding empty descriptions to relevant tasks')
+            click.echo('Adding descriptions for the playbook and to relevant tasks')
+        if 'description' not in set(self.data.keys()):
+            click.secho('No description is specified for this playbook, would you like to add a description? [Y/n]',
+                        fg='red')
+            user_answer = input()
+            if user_answer in ['n', 'N', 'no', 'No']:
+                print_error('Adding empty description for the playbook.')
```
yuvalbenshalom, line 31: if the user decided not to add a description, I think it should not print as an error. it just did what the user asked.

demisto_sdk/commands/format/update_generic.py
```diff
@@ -84,11 +84,21 @@ def set_version_to_default(self, location=None):
...
             print('Removing Unnecessary fields from file')
-        for key in arguments_to_remove:
-            self.data.pop(key, None)
+        self.recursive_remove_unnecessary_keys(schema.get('mapping'), self.data)
+
+    def recursive_remove_unnecessary_keys(self, schema, data):
+        """Recursively removes all the unnecessary fields in the file"""
+        data_keys = set(data.keys())
+        for key in data_keys:
+            if key not in schema.keys():
```
yuvalbenshalom, line 97: if we are in verbose mode, maybe it is worth noting what keys we are removing in the format.
@roysagi WDYT?

yuvalbenshalom (commented):
looks good.
see my comments.

demisto_sdk/commands/format/update_playbook.py
```diff
@@ -22,7 +22,19 @@ def __init__(self, input: str = '', output: str = '', path: str = '', from_versi
...
+                user_description = ''
+            else:
+                click.secho('Please enter the description', fg='yellow')
+                user_description = input()
```
roysagi, line 35:
```suggestion
                user_description = input("Please enter the description")
```

Don't think we need color here, this would do.

Also if a user simply clicks enter does it get an empty string or None? we should check that.

demisto_sdk/commands/format/update_playbook.py
```diff
@@ -22,7 +22,19 @@ def __init__(self, input: str = '', output: str = '', path: str = '', from_versi
     def add_description(self):
         """Add empty description to playbook and tasks."""
         if self.verbose:
-            click.echo('Adding empty descriptions to relevant tasks')
+            click.echo('Adding descriptions for the playbook and to relevant tasks')
+        if 'description' not in set(self.data.keys()):
+            click.secho('No description is specified for this playbook, would you like to add a description? [Y/n]',
+                        fg='red')
```
roysagi, line 28: I suggest using `bright-red` for better readability - test it out see which looks better to you.

demisto_sdk/commands/format/update_generic_json.py
```diff
@@ -34,7 +34,7 @@ def save_json_to_destination_file(self):
         if self.source_file != self.output_file:
             click.secho(f'Saving output JSON file to {self.output_file}', fg='white')
         with open(self.output_file, 'w') as file:
-            ujson.dump(self.data, file, indent=4)
```
roysagi, line 37: unfortunately, we should use ujson but it is missing some keys here - it should be:
```
ujson.dump(self.data, file, indent=4, encode_html_chars=True, escape_forward_slashes=False,
                   ensure_ascii=False)
```
the reason being `json` package can change the json file from what we currently have

roysagi (changes requested):

demisto_sdk/commands/format/update_generic.py
roysagi: Could be a good feature for debugging on our side.

demisto_sdk/commands/format/update_playbook.py
darkushin: If the user clicks enter, it will add the description key with an empty string.

demisto_sdk/commands/format/update_generic.py
darkushin: Added (line 105)

roysagi (approved):
looks good, great job

PR #731, opened 2020-09-02
find_type fix

darkushin:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28065
fixes: https://github.com/demisto/etc/issues/28337
fixes: https://github.com/demisto/etc/issues/28341
fixes: https://github.com/demisto/etc/issues/28359

## Description
Fixing an issue in find_type function, where the function doesn't recognize the type correctly.

## Must have
- [x] Tests
- [x] Documentation


yaakovi (changes requested):
the whole purpose of `find_type` is to find the type of the file based on the *content* of the file and not his path.
the command `demisto-sdk validate -i ~/CortexXDR.yml` should figure the correct type and run the proper validations.
same for `demisto-sdk validate -i C:\CortexXDR.yml`

the example is for `validate` but relevant for all commands.

demisto_sdk/commands/common/tools.py
```diff
@@ -769,59 +769,60 @@ def find_type(path: str = '', _dict=None, file_type: Optional[str] = None, ignor
...
-        if 'transformer' in _dict and 'keyTypeMap' in _dict:
-            return FileType.CLASSIFIER
+            if 'keyTypeMap' in _dict and 'transformer' in _dict:
+                return FileType.CLASSIFIER
 
-        if 'canvasContextConnections' in _dict:
+        if CONNECTIONS_DIR in path:
             return FileType.CONNECTION
 
         if 'mapping' in _dict:
```
roysagi, line 815: I think mappers are also under the Classifiers directory

demisto_sdk/commands/common/tools.py
```diff
@@ -769,59 +769,60 @@ def find_type(path: str = '', _dict=None, file_type: Optional[str] = None, ignor
...
+        if DASHBOARDS_DIR in path:
             return FileType.DASHBOARD
 
-        if 'group' in _dict and LAYOUT_CONTAINER_FIELDS.intersection(_dict):
-            return FileType.LAYOUTS_CONTAINER
+        if LAYOUTS_DIR in path:
+            if 'group' in _dict and LAYOUT_CONTAINER_FIELDS.intersection(_dict):
+                return FileType.LAYOUTS_CONTAINER
+
+            return FileType.LAYOUT
```
bakatzir, line 825: As yaakovi said, please do not use the abs paths as it should work from non content repos as well.
for deferring layouts pre and post 6.0 you can also look at https://github.com/demisto/demisto-sdk/pull/742

demisto_sdk/commands/format/update_generic_json.py
```diff
@@ -67,8 +67,9 @@ def remove_null_fields(self):
             a = yaml.safe_load(file_obj)
         schema_fields = a.get('mapping').keys()
         for field in schema_fields:
-            # We want to keep 'false' and 0 values.
-            if field in self.data and self.data[field] in (None, '', [], {}):
+            # We want to keep 'false' and 0 values, and avoid removing fields that are required in the schema.
+            if field in self.data and self.data[field] in (None, '', [], {}) and \
+                    not a.get('mapping').get(field).get('required'):
```
yaakovi, line 72:
```suggestion
                    not a.get('mapping', {}).get(field, {}).get('required'):
```

yaakovi (approved):

demisto_sdk/commands/common/tools.py
```diff
@@ -822,15 +822,15 @@ def find_type(path: str = '', _dict=None, file_type: Optional[str] = None, ignor
         if 'brandName' in _dict and 'transformer' in _dict:
             return FileType.OLD_CLASSIFIER
 
-        if 'transformer' in _dict and 'keyTypeMap' in _dict:
-            return FileType.CLASSIFIER
+        if ('transformer' in _dict and 'keyTypeMap' in _dict) or 'mapping' in _dict:
+            if _dict.get('type') == 'classification':
```
roysagi, line 826:
```suggestion
            if _dict.get('type') and _dict.get('type') == 'classification':
```

let's be defensive here

demisto_sdk/commands/format/update_generic_json.py
```diff
@@ -67,8 +67,9 @@ def remove_null_fields(self):
             a = yaml.safe_load(file_obj)
```
roysagi, line 67:
```suggestion
            schema_data = yaml.safe_load(file_obj)
```


A good opportunity to change a bad name to a more useful one

demisto_sdk/commands/common/tools.py
```diff
@@ -822,15 +822,15 @@ def find_type(path: str = '', _dict=None, file_type: Optional[str] = None, ignor
...
+                return FileType.CLASSIFIER
+            else:
+                return FileType.MAPPER
```
roysagi, line 829:
```suggestion
            elif _dict.get('type') and 'mapping' in _dict.get('type'):
                return FileType.MAPPER
            else:
               return None
```

might give better results - here is my thought:
 - if the type is `classification` it is a classifier.
 - if the type has `mapping` in it - it is a mapper.
 - if not both then we cannot identify the file type and we should get the "file not identified error" hence return None. 
We can also print a warning message here to direct the user to the type field. but we need to make sure it looks good.

demisto_sdk/tests/integration_tests/validate_integration_test.py
```diff
@@ -411,28 +411,28 @@ def test_missing_fromversion_field_in_new_classifier(self, mocker, repo):
...
-        new_classifier_copy = NEW_CLASSIFIER.copy()
-        new_classifier_copy['type'] = 'test'
-        classifier = pack.create_classifier('new_classifier', new_classifier_copy)
-        with ChangeCWD(pack.repo_path):
-            runner = CliRunner(mix_stderr=False)
-            result = runner.invoke(main, [VALIDATE_CMD, '-i', classifier.path], catch_exceptions=False)
-        assert f"Validating {classifier.path} as classifier" in result.stdout
-        assert 'Classifiers type must be classification' in result.stdout
-        assert result.exit_code == 1
+    # def test_invalid_type_in_new_classifier(self, mocker, repo):
```
roysagi, line 414: we can either delete this or, by implementing my suggested changes make sure the non identified file error is raised.

roysagi (changes requested):

demisto_sdk/commands/common/tools.py
yaakovi: not really necessary as `None != 'classification'`

roysagi (approved):
looks good -

Maybe we can use this branch to check: https://github.com/demisto/etc/issues/28359 before merging? @bakatzir @yuvalbenshalom what do you think?

darkushin:
> looks good -
> 
> Maybe we can use this branch to check: [demisto/etc#28359](https://github.com/demisto/etc/issues/28359) before merging? @bakatzir @yuvalbenshalom what do you think?

@roysagi I checked it for all file types (except `canvas` which I couldn't download). The validate command recognized the type correctly before and after formatting the file.

bakatzir (approved):
Well done.
please add a line or two in CHANGELOG.md

PR #732, opened 2020-09-03
[Fix] remove sourcemoduleid

DeanArbel:
## Status
- [x] Ready

## Description
sourcemodule id was over indented. Moved it to root.


bakatzir (commented):
why are we allowing that?
is not it generated when copying an integration?

DeanArbel:
@bakatzir you're 100% right. I'll change the PR to remove it.

bakatzir (approved):

PR #734, opened 2020-09-03
Validate - add verify max fetch and first fetch param

ohaim1008:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/27412

## Description
Adding checks in validate - makes sure that an integration is using fetch it has the `fetch_time` (or `first_fetch`) and `max_fetch` params.
`max_fetch` param should also have a `defaultvalue`.

## Screenshots

## Must have
- [x] Tests
- [x] Documentation


demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -94,6 +95,7 @@ def __init__(self, is_backward_check=True, prev_ver=None, use_git=False, only_co
                                    FileType.TEST_PLAYBOOK,
                                    FileType.TEST_SCRIPT,
                                    )
+        self.support_level = None
```
bakatzir, line 98: is this needed?

demisto_sdk/commands/common/constants.py
```diff
@@ -45,6 +45,11 @@
 BETA_INTEGRATION = 'betaintegration'
 OLD_REPUTATION = 'reputations.json'
 
+XSOR = 'xsor'
```
bakatzir, line 48: redundant, see `XSOAR_SUPPORT`

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -87,10 +90,13 @@ def is_valid_file(self, validate_rn: bool = True, skip_test_conf: bool = False)
             self.is_valid_description(beta_integration=False),
         ]
 
+        answer3 = []
+        if self.support_level in LEVEL_SUPPORT_OPTIONS_3:
+            answers3 = [self.is_valid_max_fetch_and_first_fetch()]
```
bakatzir, line 95: you would have to comment to explain, also let's try to think on better var names

bakatzir, line 95:
```suggestion
        answer3 = []
        if self.support_level in LEVEL_SUPPORT_OPTIONS_3:
            answers = answers.append(self.is_valid_max_fetch_and_first_fetch())
```

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -87,10 +90,13 @@ def is_valid_file(self, validate_rn: bool = True, skip_test_conf: bool = False)
...
+
         if not skip_test_conf:
             answers.append(self.are_tests_configured())
-
-        return all(answers)
+        return all(answers + answer3)
```
bakatzir, line 99:
```suggestion
        return all(answers)
```

demisto_sdk/commands/common/constants.py
```diff
@@ -933,6 +942,21 @@ class PB_Status:
...
 
+MAX_FETCH_PARAM = {
+        'name': 'max_fetch',
+        'required': False,
+        'type': 0,
+        'defaultvalue': '50'
+    }
+
+FIRST_FETCH_PARAM = {
+    'defaultvalue': '7 days',
```
bakatzir, line 953: we do not enforce this defaultvalue right? so I do not think it is needed.

ohaim1008: we don't enforce it, I was debating wether to keep it or not but I think that as an example we should keep it, it might help to prevent future bugs, wdut?

bakatzir: better to keep and comment that it is nor validate, just for reference/consideration.

demisto_sdk/commands/common/hook_validations/integration.py
ohaim1008: I thought to keep the current structure so it will be more understood that there are support levels and matching "answers".

bakatzir: I see, maybe we can rename with the other validations PR? it seems (to me) less clear this way now. @roysagi @amshamah419 WDYT.

amshamah419: I'm in agreeance with @bakatzir. It's not really clear, but we could do something like `LEVEL_3_SUPPORT_ANSWERS` or something similar.

demisto_sdk/commands/common/constants.py
```diff
@@ -841,11 +845,15 @@ class PB_Status:
...
-PACK_INITIAL_VERSION = '1.0.0'
-PACK_SUPPORT_OPTIONS = ['xsoar', 'partner', 'developer', 'community']
 XSOAR_SUPPORT = "xsoar"
 XSOAR_AUTHOR = "Cortex XSOAR"
+PACK_INITIAL_VERSION = '1.0.0'
+PACK_SUPPORT_OPTIONS = ['xsoar', 'partner', 'developer', 'community']
+LEVEL_SUPPORT_OPTIONS_1 = [XSOAR_SUPPORT]
+LEVEL_SUPPORT_OPTIONS_2 = [XSOAR_SUPPORT, CERTIFIED_PARTNER]
+LEVEL_SUPPORT_OPTIONS_3 = [XSOAR_SUPPORT, CERTIFIED_PARTNER, PARTNER]
+LEVEL_SUPPORT_OPTIONS_4 = [XSOAR_SUPPORT, CERTIFIED_PARTNER, PARTNER, COMMUNITY]
```
roysagi, line 855: There might be an easier way to do this, using the preset error_ignore flags.
```
PRESET_ERROR_TO_IGNORE = {
    'community': ['BC', 'CJ', 'DS', 'PA117'],
    'non-certified-partner': ['CJ']
}
```
see in errors.py

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -87,10 +94,13 @@ def is_valid_file(self, validate_rn: bool = True, skip_test_conf: bool = False)
             self.is_valid_description(beta_integration=False),
         ]
 
+        answers3 = []
+        if support_level in LEVEL_SUPPORT_OPTIONS_3:
+            answers3 = [self.is_valid_max_fetch_and_first_fetch()]
```
roysagi, line 99: why not append to `answers`? - your life might be made easier by y previous comment

demisto_sdk/commands/common/tools.py
```diff
@@ -1275,3 +1275,26 @@ def camel_to_snake(camel: str) -> str:
     camel_to_snake_pattern = re.compile(r'(?<!^)(?=[A-Z][a-z])')
     snake = camel_to_snake_pattern.sub('_', camel).lower()
     return snake
+
+
+def get_support_level_from_file_path(file_path: str) -> str:
```
roysagi, line 1280: seems like a good tool which is similar to what is used in `base_validator.py` in the `update_checked_flags_by_support_level` method - we might want to think about merging the two ideas

roysagi (changes requested):

demisto_sdk/commands/common/hook_validations/integration.py
ohaim1008: wanted to make it ready for later changes but I thing I will append it for now

ohaim1008:
Content changes were merge - ignoring the error in relevant integrations.
Run content with the validation - looks ok.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * Fixed an issue where image and description files were not handled correctly by **validate** and **update-release-notes** commands.
 * Fixed an issue where the **format** command didn't remove all extra fields in a file.
 * Added an error in case an invalid id_set.json file is found while running the **validate** command.
+* Added fetch params check to **validate** command.
```
bakatzir, line 16:
```suggestion
* Added fetch params checks to the **validate** command.
```

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -85,11 +90,11 @@ def is_valid_file(self, validate_rn: bool = True, skip_test_conf: bool = False)
             self.is_valid_pwsh(),
             self.is_valid_image(),
             self.is_valid_description(beta_integration=False),
+            self.is_valid_max_fetch_and_first_fetch()
```
bakatzir, line 93: so a line change won't pop in the next validation that will be added.
```suggestion
            self.is_valid_max_fetch_and_first_fetch(),
```

demisto_sdk/commands/common/tests/dashboard_test.py
```diff
@@ -17,6 +17,7 @@ def mock_structure(file_path=None, current_file=None, old_file=None):
         structure.file_path = file_path
         structure.current_file = current_file
         structure.old_file = old_file
+        structure.support_level = 'xsor'
```
bakatzir, line 20:
```suggestion
        structure.support_level = 'xsoar'
```

demisto_sdk/commands/common/tests/integration_test.py
```diff
@@ -514,6 +516,46 @@ def test_not_fetch(self, capsys):
         assert self.validator.is_valid_fetch(), 'is_valid_fetch() returns False instead True'
 
 
+class TestIsValidMaxFetchAndFirstFetch:
```
bakatzir, line 519: please use the when/given/then convention.

bakatzir (commented):
Looks good overall.

demisto_sdk/commands/common/tests/dashboard_test.py
ohaim1008: I actually can remove it.

roysagi (approved):
looks good

PR #737, opened 2020-09-06
printing warning for non-supported file types in format command

roysagi:
## Description
printing warning for non-supported file types in format command.

Also errors in case an unidentified file is given


bakatzir (approved):
legit. please add changelog.

PR #740, opened 2020-09-06
playbook schema fix

guyfreund:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28230

## Description
playbook schema fix

### DIFF
--- a/demisto_sdk/commands/common/schemas/playbook.yml
+++ b/demisto_sdk/commands/common/schemas/playbook.yml
@@ -123,7 +123,10 @@ mapping:
                 allowempty: True
               occurred:
                 required: false
-                type: text
+                type: map
+                mapping:
+                  simple:
+                    type: text
               tags:
                 required: false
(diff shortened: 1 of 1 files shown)


roysagi (approved):
please run validate -a to make sure this passes.

Also this seems weird why would it pass till now.

roysagi:
@guyfreund - Seems it passed till now because it only appears in one playbook and it is null there.

demisto_sdk/commands/common/schemas/playbook.yml
```diff
@@ -123,7 +123,10 @@ mapping:
                 allowempty: True
               occurred:
                 required: false
-                type: text
+                type: map
+                mapping:
+                  simple:
+                    type: text
```
roysagi, line 129: might need to add:
```
allowempty: true
```

PR #742, opened 2020-09-06
Enhance layout path format

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28225

## Description
Added filename enforcement for `format` command when ran against `layout` or `layoutscontainer` file - 
* **layout-** prefix will be enforced for `layout`
* **layoutscontainer-** prefix will be enforced for `layoutscontainer`

`validate` command updated with this logic as well.

## Must have
- [x] Tests
- [ ] Documentation


demisto_sdk/commands/common/hook_validations/layout.py
```diff
@@ -24,7 +25,8 @@ def is_valid_layout(self, validate_rn=True) -> bool:
                 self.is_valid_version() and
                 self.is_valid_from_version() and
                 self.is_valid_to_version() and
-                self.is_to_version_higher_than_from_version())
+                self.is_to_version_higher_than_from_version() and
+                self.is_valid_file_path())
```
bakatzir, line 29: maybe we can. do an `all` on array as we do in validate? I think it is nicer.

demisto_sdk/commands/format/update_layout.py
```diff
@@ -29,6 +32,7 @@ def __init__(self, input: str = '', output: str = '', path: str = '', from_versi
                  no_validate: bool = False, verbose: bool = False):
         super().__init__(input=input, output=output, path=path, from_version=from_version, no_validate=no_validate,
                          verbose=verbose)
+        self.is_container = any(self.data.get(kind) for kind in LAYOUTS_CONTAINER_KINDS)
```
bakatzir, line 35: can you add a comment to explain why is that the case?

demisto_sdk/commands/format/update_layout.py
```diff
@@ -38,10 +42,64 @@ def format_file(self) -> Tuple[int, int]:
...
+            Tuple -
+                Set of keys that should be deleted from file
+                Dict with layout kinds as keys and set of keys that should
+                be deleted as values.
+        """
+        if self.is_container:
+            return self.__layoutscontainer__arguments_to_remove()
+        return self.__layout__arguments_to_remove()
+
+    def __layout__run_format(self):
```
bakatzir, line 70: this is the till 6.0 layout correct? please add comments to explain.

demisto_sdk/commands/format/update_layout.py
```diff
@@ -38,10 +42,64 @@ def format_file(self) -> Tuple[int, int]:
...
+            new_output_path = self.output_file.replace(output_basename, new_output_basename)
+
+            # rename file if source and output are the same
+            if self.output_file == self.source_file:
+                os.rename(self.source_file, new_output_path)
+                self.source_file = new_output_path
+
+            self.output_file = new_output_path
+
+    def __layoutscontainer__run_format(self):
```
bakatzir, line 88: 6.0 onwards? please add docstring.

bakatzir (commented):
Nice.
let's think together how we can test this flow.

bakatzir (approved):

PR #747, opened 2020-09-08
Invalid id set error

roysagi:
## Status
- [x] In Progress
- [ ] Ready

## Description
when running validate on invalid id_set we currently get python ValueError which is not helpful - this will give a better error message.


bakatzir (approved):
please add to changelog.

demisto_sdk/commands/common/hook_validations/pack_unique_files.py
```diff
@@ -292,27 +292,36 @@ def validate_pack_unique_files(self):
...
 
-        if (non_supported_pack.get('mandatory')) or (deprecated_pack.get('mandatory')):
-            error_message, error_code = Errors.invalid_package_dependencies(self.pack)
-            if self._add_error((error_message, error_code), file_path=self.pack_path):
-                return False
-        return True
+            if (non_supported_pack.get('mandatory')) or (deprecated_pack.get('mandatory')):
+                error_message, error_code = Errors.invalid_package_dependencies(self.pack)
+                if self._add_error((error_message, error_code), file_path=self.pack_path):
+                    return False
```
bakatzir, line 318: this case seems to not be covered in ut, please consider doing so.

demisto_sdk/commands/common/hook_validations/pack_unique_files.py
```diff
@@ -292,27 +292,36 @@ def validate_pack_unique_files(self):
...
+            return True
+        except ValueError as e:
+            if "Couldn't find any items for pack" in str(e):
+                error_message, error_code = Errors.invalid_id_set()
+                if self._add_error((error_message, error_code), file_path=self.pack_path):
+                    return False
+                return True
+            else:
+                raise e
```
bakatzir, line 327: this case seems to not be covered in ut, please consider doing so, especially as it is the addition.

yaakovi, line 327:
```suggestion
                raise
```

PR #754, opened 2020-09-12
Fix packs flag

GalRabin:
FYI - @yaakovi 

## Status
- [ ] In Progress
- [x] Ready

## Description
 - Fix flag to true/false value.
 - Fix summary to print only content_packs if flag is on.

I tested and validated functionality.

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -311,7 +311,7 @@ def validate(config, **kwargs):
 @click.option('-a', '--artifacts_path', help='Destination directory to create the artifacts.',
               type=click.Path(file_okay=False, resolve_path=True))
 @click.option('--zip/--no-zip', help='Zip content artifacts folders', default=True)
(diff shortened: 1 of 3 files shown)


demisto_sdk/__main__.py
```diff
@@ -311,7 +311,7 @@ def validate(config, **kwargs):
 @click.option('-a', '--artifacts_path', help='Destination directory to create the artifacts.',
               type=click.Path(file_okay=False, resolve_path=True))
 @click.option('--zip/--no-zip', help='Zip content artifacts folders', default=True)
-@click.option('--content-packs', help='Create only content_packs artifacts.', default=False)
+@click.option('--packs', help='Create only content_packs artifacts.', is_flag=True)
```
yaakovi, line 314: why are you changing the argument name?

yaakovi (approved):

demisto_sdk/__main__.py
GalRabin: This was the default argument name in the previous version.

PR #767, opened 2020-09-17
Fix lint duplicate log message

GalRabin:
FYI - @yaakovi @bakatzir @orhovy @roysagi 

## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28546

## Description
Fixing duplicate lint log messages, I produced this bug when I rewrite the `content-create-artifacts` command, However I believe that we should develop a better mechanism for logging in our code (Singleton or any other solution) so it will not happen again.

## Screenshots
After fix:
image


bakatzir (approved):
please add a changelog.

PR #768, opened 2020-09-17
[Fix] layoutscontainer (layout in 6.0 format) wrong layouts dir

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27878

## Description
Added support for layoutscontainer in **init** contribution flow - layoutscontainer should now be created as part of the `Layouts` folder.

## Must have
- [x] Tests


avidan-H (changes requested):
Can you please also add to the unit test a normal (pre-6.0 layout file) layout file, and assert that it also exists in the directory after the conversion? I want to make sure that we don't accidentally write over the existing directory and contents.

TestSuite/contribution.py
```diff
@@ -75,6 +77,9 @@ def __init__(self, tmpdir: Path, name: str, repo):
         self._reputations_path = self.target_dir / 'reputation'
         self._reputations_path.mkdir()
 
+        self._layoutscontainer_path = self.target_dir / 'layoutscontainer'
+        self._layoutscontainer_path.mkdir()
+
```
avidan-H, line 82:
```suggestion
		self._layouts_path = self.target_dir / 'layout'
		self._layouts_path.mkdir()

        self._layoutscontainer_path = self.target_dir / 'layoutscontainer'
        self._layoutscontainer_path.mkdir()

```

TestSuite/contribution.py
```diff
@@ -162,6 +167,24 @@ def create_dashboard(
...
         return dashboard
 
+    def create_layout(
+            self,
+            name,
+            content: dict = None
+    ):
+        prefix = LAYOUT
+        layout = self._create_json_based(name, prefix, content, dir_path=self._layoutscontainer_path)
+        self.layouts_containers.append(layout)
```
avidan-H, line 177:
```suggestion
        layout = self._create_json_based(name, prefix, content, dir_path=self._layouts_path)
        self.layouts.append(layout)
        return layout
```

TestSuite/contribution.py
```diff
@@ -162,6 +167,24 @@ def create_dashboard(
...
+
+    def create_layoutscontainer(
+            self,
+            name,
+            content: dict = None
+    ):
+        prefix = LAYOUTS_CONTAINER
+        layoutscontainer = self._create_json_based(name, prefix, content, dir_path=self._layoutscontainer_path)
+        self.layouts_containers.append(layoutscontainer)
```
avidan-H, line 186:
```suggestion
        self.layouts_containers.append(layoutscontainer)
        return layoutscontainer
```

demisto_sdk/commands/init/initiator.py
```diff
@@ -222,7 +222,15 @@ def convert_contribution_to_pack(self):
                     dst_name = ENTITY_TYPE_TO_DIR.get(basename)
                     src_path = os.path.join(pack_dir, basename)
                     dst_path = os.path.join(pack_dir, dst_name)
-                    content_item_dir = shutil.move(src_path, dst_path)
+                    if os.path.exists(dst_path):
```
avidan-H, line 225: Perhaps add a comment here about the case where this applies if you think it would be helpful. Up to you - otherwise, it looks good.

avidan-H (approved):

demisto_sdk/commands/init/initiator.py
```diff
@@ -222,7 +222,15 @@ def convert_contribution_to_pack(self):
...
+                        # move src folder files to dst folder
+                        for _, _, files in os.walk(src_path, topdown=False):
+                            for name in files:
+                                src_file_path = os.path.join(src_path, name)
+                                shutil.move(src_file_path, dst_path)
+                    else:
+                        # replace dst folder with src folder
+                        content_item_dir = shutil.move(src_path, dst_path)
                     if basename in {SCRIPT, AUTOMATION, INTEGRATION}:
                         self.content_item_to_package_format(content_item_dir, del_unified=True)
```
avidan-H, line 235: This won't happen in the particular case you addressed with these changes, but something you might want to think about is that now there is the possibility that the variable `content_item_dir` will not be assigned to even though it is used here.

demisto_sdk/commands/init/initiator.py
DeanArbel: I thought about adding in an explanation about why we do this, but I feel the added comments explain what's being done, and though it doesn't explain the full reason why, it does enough to warrant the user (while the UT add another layer of warranty)

demisto_sdk/commands/init/initiator.py
DeanArbel: Agreed, added handling for that

PR #771, opened 2020-09-21
Added a validation for removal of integration parameters in **validat…

bakatzir:
…e** command

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/27137

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/integration.py
+++ b/demisto_sdk/commands/common/hook_validations/integration.py
@@ -50,6 +50,7 @@ class IntegrationValidator(ContentEntityValidator):
 
         answers = [
             self.is_changed_context_path(),
+            self.is_removed_integration_parameters(),
             self.is_added_required_fields(),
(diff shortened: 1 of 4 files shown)


demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -547,6 +548,23 @@ def is_changed_context_path(self):
 
         return False
 
+    def is_removed_integration_parameters(self):
```
yuvalbenshalom, line 551: shouldn't this be already covered by the case that that a required parameter was changed? do we really need a new function for this?

bakatzir: `is_added_required_fields` Is for adding new required params.
I thought of putting it in the same func, but IMO it is best suited in a different one.

yuvalbenshalom (approved):

PR #779, opened 2020-09-23
Gen docs small improvements

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Small improvements to generate docs:

* Clarification on -e arg
* Use json formatting for context data
* Remove `Deprecated` logic in Script docs as this is handled in the the docs site.
* Script "Used In" to include max 10 entries
* Ignore 0.0.0 version (probably less relevant after move for version branch in content repo)
* Human Readable in script to be quoted as we do in integration docs
* Add _DEMISTO_SDK_ID_SET_REFRESH_INTERVAL_ (see changelog)
 

Generate doc can be seen here: 
https://github.com/demisto/content/blob/e94f5d29972183d33458e8743ef5f7ef619a6796/Packs/CommonScripts/Scripts/script-Set_README.md

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/common/update_id_set.py
+++ b/demisto_sdk/commands/common/update_id_set.py
@@ -1059,10 +1059,10 @@ def re_create_id_set(id_set_path: Optional[str] = DEFAULT_ID_SET_PATH, objects_t
         except ValueError:
             refresh_interval = -1
             print_color(
-                f"Re-creating id_set.\n"
-                f"DEMISTO_SDK_ID_SET_REFRESH_INTERVAL env var is set with value: "
+                "Re-creating id_set.\n"
+                "DEMISTO_SDK_ID_SET_REFRESH_INTERVAL env var is set with value: "
                 f"{os.getenv('DEMISTO_SDK_ID_SET_REFRESH_INTERVAL')} which is an illegal integer."
-                f"\nPlease modify or unset env var.", LOG_COLORS.YELLOW
+                "\nPlease modify or unset env var.", LOG_COLORS.YELLOW
             )
         if refresh_interval > 0:  # if file is newer than refersh interval use it as is
             mtime = os.path.getmtime(id_set_path)
--- a/demisto_sdk/commands/common/update_id_set.py
+++ b/demisto_sdk/commands/common/update_id_set.py
@@ -5,10 +5,11 @@ import os
 import re
 import time
 from collections import OrderedDict
+from datetime import datetime
 from distutils.version import LooseVersion
 from functools import partial
 from multiprocessing import Pool, cpu_count
-from typing import Callable, Tuple
+from typing import Callable, Optional, Tuple
 
 import click
 import networkx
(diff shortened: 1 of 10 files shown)


demisto_sdk/commands/generate_docs/generate_script_doc.py
```diff
@@ -74,8 +71,17 @@ def generate_script_doc(input, examples, output: str = None, permissions: str =
...
-                                             text='This script is used in the following playbooks and scripts.'))
+            if len(used_in) <= 10:
+                doc.extend(generate_list_section('Used In', used_in, True,
+                                                 text='This script is used in the following playbooks and scripts.'))
+            else:  # if we have more than 10 use a sample
+                print_warning(f'"Used In" section found too many scripts/playbooks ({len(used_in)}). Will use a sample of 10.'
+                              ' Full list is available as a comment in the README file.')
+                sample_used_in = random.sample(used_in, 10)
+                doc.extend(generate_list_section('Used In', sorted(sample_used_in), True,
+                                                 text='Sample usage of this script can be found in the following playbooks and scripts.'))
```
yaakovi, line 82: why random?

glicht:
Note that I did a change in id set creation. If an empty string is passed as path it will use the default path. If None is explicitly passed then the file will not be created. I am not sure if there is even a scenario that we need to create the id set without saving, but I left the option.

demisto_sdk/commands/create_id_set/create_id_set.py
```diff
@@ -3,6 +3,13 @@
 
 class IDSetCreator:
     def __init__(self, output: str = '', print_logs: bool = True):
+        """IDSetCreator
+
+        Args:
+            output (str, optional): The output path. Set to None to avoid creation of a file. '' means the default path.
+             Defaults to ''.
```
yaakovi, line 10:
```suggestion
             Defaults to 'Tests/id_set.json'.
```

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -1036,8 +1037,46 @@ def get_general_paths(path):
...
+                            "doesn't require a refresh. Will use current id set. "
+                            "If you want to force an id set referesh unset DEMISTO_SDK_ID_SET_REFRESH_INTERVAL or set to -1.", LOG_COLORS.GREEN)
+                with open(id_set_path, mode="r") as f:
+                    return json.load(f)
+            else:
+                print_color(f"DEMISTO_SDK_ID_SET_REFRESH_INTERVAL env var is set but current id_set: {id_set_path} "
+                            f"modify time: {mtime_dt} is older than refresh interval. "
+                            "Will re-generate the id set.", LOG_COLORS.GREEN)
+        else:
+            print_color("Note: DEMISTO_SDK_ID_SET_REFRESH_INTERVAL env var is not enabled. "
```
yaakovi, line 1074: why green and not a warning?

yaakovi (approved):

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -1036,8 +1037,46 @@ def get_general_paths(path):
...
+        id_set_path (str, optional): If passed an empty string will use default path. Pass in None to avoid saving the id set.
+            Defaults to DEFAULT_ID_SET_PATH.
+        objects_to_create (list, optional): [description]. Defaults to None.
+
+    Returns: id set object
+    """
+    if id_set_path == "":
+        id_set_path = DEFAULT_ID_SET_PATH
+    if id_set_path and os.path.exists(id_set_path):
+        refresh_interval = int(os.getenv('DEMISTO_SDK_ID_SET_REFRESH_INTERVAL', -1))
```
teizenman, line 1057:
```suggestion
        try:
            refresh_interval = int(os.getenv('DEMISTO_SDK_ID_SET_REFRESH_INTERVAL', -1))
        except ValueError:
            refresh_interval = -1
            print_color(
                f"Re-creating id_set.\n"
                f"DEMISTO_SDK_ID_SET_REFRESH_INTERVAL env var is set with value: "
                f"{os.getenv('DEMISTO_SDK_ID_SET_REFRESH_INTERVAL')} which is an illegal integer."
                f"\nPlease modify or unset env var.", LOG_COLORS.YELLOW
            )
```

demisto_sdk/commands/generate_docs/generate_script_doc.py
glicht: The idea is that when "Used In" is too big a list we limit to 10. I don't know which are the best 10 so just choose randomly.

demisto_sdk/commands/common/update_id_set.py
glicht: I think it's ok. By default this variable is not set. It is for advanced usage so don't want it to be a warning.

PR #787, opened 2020-09-29
Validate Contribution Certification in pack_metadata

darkushin:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/27412

## Description
Add validation for the pack_metadata.json file. The validation should verify that if the field 'certification' is in the pack_metadata file, its value is 'certified' or 'verified'.


demisto_sdk/commands/common/hook_validations/pack_unique_files.py
```diff
@@ -247,6 +247,13 @@ def _is_pack_meta_file_structure_valid(self):
                                            self.pack_meta_file):
                             return False
 
+            # if the field 'certification' exists, check that its value is set to 'certified' or 'verified'
+            certification = metadata.get(PACK_METADATA_CERTIFICATION)
+            if certification and certification not in ['certified', 'verified']:
```
reutshal, line 252: You can create a constant - "ALLOWED_CERTIFICATION_VALUES" = ['certified', 'verified']
So if want to add/remove one of them in the future it will be easier.
And then ask:
`if certification and certification not in ALLOWED_CERTIFICATION_VALUES:`

reutshal (approved):
Looks good!
Please also add a note about this in the sdk CHANGELOG

PR #809, opened 2020-10-08
remove uneeded term from download

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Description
There's really no need to validate that the dir is in content, just we have packs.


guyfreund:
@jochman why is that?

jochman:
> @jochman why is that?

It's the only piece of code that check we're in content dir and not checking only that we're using Packs. 
And if forces me to do ugly things instead of just delete this term.

guyfreund (approved):
add changelog

PR #821, opened 2020-10-12
Build improvements

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Changed order of build steps to reduce build time. Build is down to under 20 minutes (before was over 25).

Main change is that tox-3-7 is used as dependency for the content repo tests instead of tox-3-8. tox 3-8 which is with coverage is run in paralel.

See the build in circleci for visual flow of the jobs.

## Additional
Added a small fix so unit test update rn doesn't modify files in the repo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,14 +8,15 @@ references:
       filters:
         tags:
           only: /^v\d+\.\d+\.\d+$/
-    
+
     reqs_tox: &reqs_tox
       requires:
-        - tox3-8
+        - tox3-7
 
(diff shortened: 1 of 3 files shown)


glicht:
Note: tried running parallel unit tests here: https://github.com/demisto/demisto-sdk/pull/821/commits/486134f304f43f95cab79de7ac6d25bb2e5d583b . But it looks like we have some unit tests that can't run in parallel.

anara123 (approved):

PR #824, opened 2020-10-14
Unifier - add Integration doc link to detailed description

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28927

## Description
Added to unifier to add link to integration doc in the integration detailed description

## Screenshots
Screen Shot 2020-10-14 at 15 40 42

## Must have
- [x] Tests


demisto_sdk/commands/unify/unifier.py
```diff
@@ -433,3 +444,41 @@ def add_contributors_support(self, unified_yml: Dict, contributor_type: str, con
...
+            unified_yml (Dict): The integration YAML dictionary object
+
+        Returns:
+            str: The integration doc markdown link to add to the detailed description (if reachable)
+        """
+        normalized_integration_id = self.normalize_integration_id(unified_yml['commonfields']['id'])
+        integration_doc_link = INTEGRATIONS_DOCS_REFERENCE + normalized_integration_id
+
+        try:
+            res = requests.get(integration_doc_link, verify=False, timeout=10)
```
glicht, line 461: Few issues with doing this:
1. it will slow down the creation of the content pack
2. We only publish the doc after it is in master. So we are not running this code while working on the branch (if the integration was not yet merged to master with docs).

I think it will be better to simply check if there is a README.md file in the directory and if so just add the link.

glicht:
@Itay4 Looks good. 

If the detailed description is non-empty let's add a hr line. So there is a clear separation from the tips. If the detailed description is empty then no need for the hr line. 

So it will be like this:


---
[View Integration Documentation](https://xsoar.pan.dev)

glicht (approved):
Let's also add a test that calls `get_integration_doc_link` and checks that the url actually exists. This will add a failure if we ever do a change on xsoar.pan.dev for some reason. Use both an integration that has dashes and spaces in its name.

yaakovi (approved):

demisto_sdk/commands/unify/tests/unifier_test.py
```diff
@@ -151,10 +153,82 @@ def test_insert_description_to_yml():
...
+    """
+    readme = tmp_path / 'README.md'
+    readme.write_text('README')
+    unifier = Unifier(str(tmp_path))
+    integration_doc_link = unifier.get_integration_doc_link({'commonfields': {'id': 'Cortex XDR - IOC'}})
+    assert integration_doc_link == \
+        '[View Integration Documentation](https://xsoar.pan.dev/docs/reference/integrations/cortex-xdr---ioc)'
+    link = re.findall(r'\(([^)]+)\)', integration_doc_link)[0]
+    try:
+        requests.get(link, verify=False, timeout=10)
```
glicht, line 207: add `raise_for_status` as you will get a 404 if the page is missing

glicht (approved):

PR #850, opened 2020-10-28
demisto-sdk validate --staged flag

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28828

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -18,7 +18,7 @@ from demisto_sdk.commands.common.tools import (find_type,
                                                get_last_remote_release_version,
                                                get_pack_name,
                                                pack_name_to_path, print_error,
-                                               print_warning)
+                                               print_warning, run_command)
 from demisto_sdk.commands.create_artifacts.content_artifacts_creator import (
     ArtifactsManager, create_content_artifacts)
 from demisto_sdk.commands.create_id_set.create_id_set import IDSetCreator
@@ -217,9 +217,10 @@ def unify(**kwargs):
 
 
 # ====================== validate ====================== #
-@main.command(name="validate",
-              short_help='Validate your content files. If no additional flags are given, will validated only '
-                         'committed files')
+@main.command(
+    short_help='Validate your content files. If no additional flags are given, will validated only '
+    'committed files'
+)
 @click.help_option(
     '-h', '--help'
 )
@@ -230,12 +231,10 @@ def unify(**kwargs):
     '-s', '--id-set', is_flag=True,
     default=False, show_default=True, help='Validate the id_set file.')
 @click.option(
-    '--prev-ver', help='Previous branch or SHA1 commit to run checks against.')
+    "-idp", "--id-set-path", help="The path of the id-set.json used for validations.",
+    type=click.Path(resolve_path=True))
 @click.option(
-    '--post-commit',
-    is_flag=True,
-    help='Whether the validation should run only on the current branch\'s committed changed files. '
-         'This applies only when the -g flag is supplied.')
+    '--prev-ver', help='Previous branch or SHA1 commit to run checks against.')
 @click.option(
     '--no-backward-comp', is_flag=True, show_default=True,
     help='Whether to check backward compatibility or not.')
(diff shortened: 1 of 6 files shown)


demisto_sdk/__main__.py
```diff
@@ -217,9 +217,10 @@ def unify(**kwargs):
 
 
 # ====================== validate ====================== #
-@main.command(name="validate",
-              short_help='Validate your content files. If no additional flags are given, will validated only '
-                         'committed files')
+@main.command(
```
yaakovi, line 220: where is the command name?

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -811,6 +814,14 @@ def get_modified_and_added_files(self, compare_type, prev_ver):
         """
         if not self.no_configuration_prints:
             click.echo("Collecting all committed files")
```
roysagi, line 816: move this print to line 814.
This serves as the header for the committed files check.

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -811,6 +814,14 @@ def get_modified_and_added_files(self, compare_type, prev_ver):
...
+        if self.staged:
```
roysagi, line 817: when running in staged we want to run on all committed and staged files. If I get this correctly you are running on only staged files - unify with the results of the committed files section (actually a good opportunity to separate that section and this section to smaller commands to make the git handling more moduler.

demisto_sdk/tests/integration_tests/validate_integration_test.py
```diff
@@ -357,9 +357,8 @@ def test_integration_validate_invalid_pack_path(self):
         """
         runner = CliRunner(mix_stderr=False)
         result = runner.invoke(main, [VALIDATE_CMD, "-i", AZURE_FEED_INVALID_PACK_PATH])
-        assert result.exit_code == 1
-        assert f'{AZURE_FEED_INVALID_PACK_PATH} was not found' in result.output
-        assert result.stderr == ""
+        assert 'does not exist' in result.stderr
+        assert result.exit_code == 2
```
roysagi, line 361: when are we returning exit code 2? this should be 1.

demisto_sdk/tests/integration_tests/validate_integration_test.py
```diff
@@ -1813,9 +1812,9 @@ def test_validation_non_content_path(self):
         runner = CliRunner(mix_stderr=False)
         result = runner.invoke(main, [VALIDATE_CMD, '-i', join('Users', 'MyPacks', 'VMware'), '--no-docker-checks',
                                       '--no-conf-json', '--skip-pack-release-notes'], catch_exceptions=False)
-        assert result.exit_code == 1
+        assert result.exit_code == 2
```
roysagi, line 1815: same here

roysagi (changes requested):

demisto_sdk/tests/integration_tests/validate_integration_test.py
jochman: It fails on click checking if this path exists - automatically exit code 0 (with type=click.Path(exists=True))

demisto_sdk/tests/integration_tests/validate_integration_test.py
jochman: same here

demisto_sdk/commands/validate/validate_manager.py
jochman: It's for the hooks - Need to run on the staged files only, not the committed. it's a pre-commit.

demisto_sdk/__main__.py
jochman: You don't need it. it uses the command name below as the command name. (it replaces `_` with `-`)

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -845,11 +859,14 @@ def get_modified_and_added_files(self, compare_type, prev_ver):
...
                     click.echo("Collecting all local changed files against the content master")
-
-                # only changes against prev_ver (without local changes)
-                all_changed_files_string = run_command('git diff --name-status {}'.format(prev_ver))
+                if self.staged:
+                    click.echo("Collecting staged files only")
+                    all_changed_files_string = run_command('git diff --name-status --staged master')
+                else:
+                    # only changes against prev_ver (without local changes)
+                    all_changed_files_string = run_command('git diff --name-status {}'.format(prev_ver))
```
roysagi, line 869: If `staged` is set then it would return in the changes you made above - this code is unreachable.

roysagi (changes requested):

demisto_sdk/commands/validate/validate_manager.py
jochman: if staged is set, those the only things that should come back.

roysagi (approved):

PR #853, opened 2020-10-29
Readme validate fix

glicht:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
fix readme validation to apply same mdx fix logic as in content-docs. Will detect now the following file https://github.com/demisto/content/blob/b63f7408b7b7afbc4db0c369955b8a9003232539/Packs/PhishTank/Integrations/PhishTankV2/README.md which failing to deploy in content-docs

## Additional
change in workflow_test.py to lazy initialise ContentGitRepo so it isn't done in test discovery. @jochman FYI. 

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/readme.py
+++ b/demisto_sdk/commands/common/hook_validations/readme.py
@@ -3,6 +3,7 @@ import json
 import os
 import re
 import subprocess
+import tempfile
 from functools import lru_cache
 from pathlib import Path
 from threading import Lock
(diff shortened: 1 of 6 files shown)


demisto_sdk/commands/common/hook_validations/readme.py
```diff
@@ -55,9 +56,15 @@ def is_valid_file(self) -> bool:
...
     def mdx_verify(self) -> bool:
         mdx_parse = Path(__file__).parent.parent / 'mdx-parse.js'
-        # run the java script mdx parse validator
-        _, stderr, is_not_valid = run_command_os(f'node {mdx_parse} -f {self.file_path}', cwd=self.content_path,
-                                                 env=os.environ)
+        with open(self.file_path, 'r') as f:
+            readme_content = f.read()
+        readme_content = self.fix_mdx(readme_content)
+        with tempfile.NamedTemporaryFile('w+t') as fp:
+            fp.write(readme_content)
```
anara123, line 63: why do you read the original file and then write to temporary file?

glicht: As we modify the original content with the function `fix_mdx` and then we need to write it back as the script works with a file path.

jochman: also, we got lots of problems with tests writing to the test file itself. If we're using files - we should only read from them,.

anara123 (approved):

PR #855, opened 2020-11-03
Speed up pack dependencies calculations

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
Related: https://github.com/demisto/etc/issues/29255
Fixes: https://github.com/demisto/etc/issues/30052

## Description
Improved the timing from ~10-13 minutes to ~15 seconds.

1. Instead of generating a new dependencies graph for each pack from scratch and calculate its dependencies - A full dependencies graph of all packs will be generated once, saved into memory, and from it the dependency sub-graph of each pack will be extracted and used for the calculations
2. Once the full graph is ready, the dependency calculation of each pack will be executed concurrently using multiprocessing.

This is not a breaking change as I only added new methods so this can be merged regardless to the content PR.

This PR is about section 1, section 2 is in the matching content PR [here](https://github.com/demisto/content/pull/9793)
### Before:
image
### After:
image

## Must have
- [x] Tests

### DIFF
--- a/demisto_sdk/commands/find_dependencies/find_dependencies.py
+++ b/demisto_sdk/commands/find_dependencies/find_dependencies.py
@@ -2,7 +2,9 @@ import glob
 import json
 import os
 import sys
+from copy import deepcopy
 from distutils.version import LooseVersion
+from typing import Union
 
 import click
 import networkx as nx
@@ -33,7 +35,7 @@ class VerboseFile:
         self.fd = None
 
 
-def parse_for_pack_metadata(dependency_graph, graph_root):
+def parse_for_pack_metadata(dependency_graph: nx.DiGraph, graph_root: str) -> tuple:
     """
     Parses calculated dependency graph and returns first and all level parsed dependency.
     Additionally returns list of displayed pack images of all graph levels.
@@ -60,7 +62,7 @@ def parse_for_pack_metadata(dependency_graph, graph_root):
     return first_level_dependencies, all_level_dependencies
 
 
-def find_pack_path(pack_folder_name):
+def find_pack_path(pack_folder_name: str) -> list:
     """
     Find pack path matching from content repo root directory.
 
@@ -77,7 +79,7 @@ def find_pack_path(pack_folder_name):
     return found_path_results
 
 
-def find_pack_display_name(pack_folder_name):
+def find_pack_display_name(pack_folder_name: str) -> str:
     """
     Returns pack display name from pack_metadata.json file.
 
@@ -103,7 +105,7 @@ def find_pack_display_name(pack_folder_name):
     return pack_display_name
 
 
-def update_pack_metadata_with_dependencies(pack_folder_name, first_level_dependencies):
+def update_pack_metadata_with_dependencies(pack_folder_name: str, first_level_dependencies: dict) -> None:
     """
     Updates pack metadata with found parsed dependencies results.
 
@@ -137,7 +139,7 @@ class PackDependencies:
     """
 
     @staticmethod
-    def _search_for_pack_items(pack_id, items_list):
+    def _search_for_pack_items(pack_id: str, items_list: list) -> list:
         """
         Filtering of content items that belong to specific pack.
 
@@ -151,7 +153,9 @@ class PackDependencies:
         return list(filter(lambda s: next(iter(s.values())).get('pack') == pack_id, items_list))
 
     @staticmethod
-    def _search_packs_by_items_names(items_names, items_list, exclude_ignored_dependencies=True):
+    def _search_packs_by_items_names(items_names: Union[str, list],
+                                     items_list: list,
+                                     exclude_ignored_dependencies: bool = True) -> set:
         """
         Searches for implemented script/integration/playbook.
 
@@ -179,7 +183,9 @@ class PackDependencies:
         return {p for p in pack_names if p not in constants.IGNORED_DEPENDENCY_CALCULATION}
 
     @staticmethod
-    def _search_packs_by_items_names_or_ids(items_names, items_list, exclude_ignored_dependencies=True):
+    def _search_packs_by_items_names_or_ids(items_names: Union[str, list],
+                                            items_list: list,
+                                            exclude_ignored_dependencies: bool = True) -> set:
         """
         Searches for implemented packs of the given items.
 
@@ -210,7 +216,9 @@ class PackDependencies:
         return packs
 
     @staticmethod
-    def _search_packs_by_integration_command(command, id_set, exclude_ignored_dependencies=True):
+    def _search_packs_by_integration_command(command: str,
+                                             id_set: dict,
+                                             exclude_ignored_dependencies: bool = True) -> set:
         """
         Filters packs by implementing integration commands.
 
@@ -234,7 +242,7 @@ class PackDependencies:
         return {p for p in pack_names if p not in constants.IGNORED_DEPENDENCY_CALCULATION}
 
     @staticmethod
-    def _detect_generic_commands_dependencies(pack_ids):
+    def _detect_generic_commands_dependencies(pack_ids: set) -> list:
         """
         Detects whether dependency is mandatory or not. In case two packs implements the same command,
         mandatory is set to False.
@@ -249,7 +257,7 @@ class PackDependencies:
         return [(p, False) if len(pack_ids) > 1 else (p, True) for p in pack_ids]
 
     @staticmethod
-    def _label_as_mandatory(pack_ids):
+    def _label_as_mandatory(pack_ids: set) -> list:
         """
         Sets pack as mandatory.
 
@@ -263,7 +271,7 @@ class PackDependencies:
         return [(p, True) for p in pack_ids]
 
     @staticmethod
-    def _label_as_optional(pack_ids):
+    def _label_as_optional(pack_ids: set) -> list:
         """
         Sets pack as optional.
 
@@ -277,7 +285,10 @@ class PackDependencies:
         return [(p, False) for p in pack_ids]
 
     @staticmethod
-    def _collect_scripts_dependencies(pack_scripts, id_set, verbose_file, exclude_ignored_dependencies=True):
+    def _collect_scripts_dependencies(pack_scripts: list,
+                                      id_set: dict,
+                                      verbose_file: VerboseFile,
+                                      exclude_ignored_dependencies: bool = True) -> set:
         """
         Collects script pack dependencies.
 
@@ -332,8 +343,10 @@ class PackDependencies:
         return dependencies_packs
 
     @staticmethod
-    def _differentiate_playbook_implementing_objects(implementing_objects, skippable_tasks, id_set_section,
-                                                     exclude_ignored_dependencies=True):
+    def _differentiate_playbook_implementing_objects(implementing_objects: list,
+                                                     skippable_tasks: set,
+                                                     id_set_section: list,
+                                                     exclude_ignored_dependencies: bool = True) -> set:
         """
         Differentiate implementing objects by skippable.
 
@@ -366,7 +379,8 @@ class PackDependencies:
         return dependencies
 
     @staticmethod
-    def _collect_playbooks_dependencies(pack_playbooks, id_set, verbose_file, exclude_ignored_dependencies=True):
+    def _collect_playbooks_dependencies(pack_playbooks: list, id_set: dict, verbose_file: VerboseFile,
+                                        exclude_ignored_dependencies: bool = True) -> set:
         """
         Collects playbook pack dependencies.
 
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -874,6 +875,8 @@ def _collect_pack_items(pack_id, id_set):
         pack_items['classifiers'] = PackDependencies._search_for_pack_items(pack_id, id_set['Classifiers'])
         pack_items['mappers'] = PackDependencies._search_for_pack_items(pack_id, id_set['Mappers'])
         pack_items['widgets'] = PackDependencies._search_for_pack_items(pack_id, id_set['Widgets'])
+        pack_items['dashboards'] = PackDependencies._search_for_pack_items(pack_id, id_set['Dashboards'])
```
Itay4, line 878: are all content entities now covered?

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -966,6 +969,52 @@ def _find_pack_dependencies(pack_id, id_set, verbose_file, exclude_ignored_depen
 
         return pack_dependencies
 
+    @staticmethod
+    def build_all_dependencies_graph(pack_ids, id_set, verbose_file, exclude_ignored_dependencies=True):
```
Itay4, line 973: add type hints

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -966,6 +969,52 @@ def _find_pack_dependencies(pack_id, id_set, verbose_file, exclude_ignored_depen
...
+            id_set (dict): id set json.
+            verbose_file (VerboseFile): path to dependency explanations file.
+            exclude_ignored_dependencies (bool): Determines whether to include unsupported dependencies or not.
+
+        Returns:
+            DiGraph: all dependencies of given packs.
+        """
+        dependency_graph = nx.DiGraph()
+        for pack in pack_ids:
+            dependency_graph.add_node(pack, mandatory_for=[])
```
Itay4, line 988: why do we need to initalize `mandatory_for` with empty list?

demisto_sdk/commands/find_dependencies/find_dependencies.py
Itay4, line 973: why do we need `exclude_ignored_dependencies` arg? it is `True` by default in `_find_pack_dependencies` anyway

demisto_sdk/commands/find_dependencies/find_dependencies.py
hod-alpert: I'm not sure what other content entities there are, so I don't know if i'd say that, but there are no failures currently on this step.

And the dashboards and reports are not covered, they are only collected as items in the pack now, but the dependencies are not calculated from them. I opened another issue for that here https://github.com/demisto/etc/issues/30118

demisto_sdk/commands/find_dependencies/find_dependencies.py
hod-alpert: I see that the standard in the file is to specify the types in the docstring so that's what I did, do you want me to add type hints to all methods in this file instead?

hod-alpert: What if someone wants to use the whole `build_all_dependencies_graph` method with `exclude_ignored_dependencies` as `False`?

Itay4, line 973: please add test for `build_all_dependencies_graph` as well

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -966,6 +969,52 @@ def _find_pack_dependencies(pack_id, id_set, verbose_file, exclude_ignored_depen
...
+        for pack in dependency_graph.nodes():
+            dependencies = PackDependencies._find_pack_dependencies(
+                pack, id_set, verbose_file=verbose_file, exclude_ignored_dependencies=exclude_ignored_dependencies)
```
Itay4, line 991:
```suggestion
                pack, id_set, verbose_file = verbose_file, exclude_ignored_dependencies=exclude_ignored_dependencies)
```

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -966,6 +969,52 @@ def _find_pack_dependencies(pack_id, id_set, verbose_file, exclude_ignored_depen
...
+                    continue
+                if dependency_name not in dependency_graph:
+                    dependency_graph.add_node(dependency_name, mandatory_for=[])
+                dependency_graph.add_edge(pack, dependency_name)
+                if is_mandatory:
+                    dependency_graph.nodes()[dependency_name]['mandatory_for'].append(pack)
+        return dependency_graph
+
+    @staticmethod
+    def get_dependencies_subgraph_by_dfs(dependencies_graph, source_pack):
```
Itay4, line 1003: add type hints

demisto_sdk/commands/find_dependencies/tests/find_dependencies_test.py
```diff
@@ -1119,6 +1120,33 @@ def test_search_packs_by_items_names_or_ids(item_names, section_name, expected_r
 
 
 class TestDependencyGraph:
+    @pytest.mark.parametrize('source_node, expected_nodes_in, expected_nodes_out',
+                             [('1', ['1', '2', '3'], ['4']),
```
Itay4, line 1124: i think it would be better if we use pack names here instead of numbers, just so it would be more intuitive

demisto_sdk/commands/find_dependencies/find_dependencies.py
Itay4: so should we add a TODO here for that ?

demisto_sdk/commands/find_dependencies/find_dependencies.py
Itay4: yes please, so mypy can run on it

Itay4: if you think it makes sense you can keep it, just dont see why would someone want to do so

demisto_sdk/commands/find_dependencies/find_dependencies.py
hod-alpert: But these are method invocation params
```
dependencies = PackDependencies._find_pack_dependencies(
                pack, id_set, verbose_file=verbose_file, exclude_ignored_dependencies=exclude_ignored_dependencies)
```

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -967,7 +992,59 @@ def _find_pack_dependencies(pack_id, id_set, verbose_file, exclude_ignored_depen
...
+        for pack in pack_ids:
+            dependency_graph.add_node(pack, mandatory_for_packs=[])
+        for pack in dependency_graph.nodes():
+            dependencies = PackDependencies._find_pack_dependencies(
+                pack, id_set, verbose_file=verbose_file, exclude_ignored_dependencies=exclude_ignored_dependencies)
+            for dependency_name, is_mandatory in dependencies:
+                if dependency_name == pack:
+                    continue
+                if dependency_name not in dependency_graph:
+                    dependency_graph.add_node(dependency_name, mandatory_for_packs=[])
```
yaakovi, line 1021: when will this happen? is that purely for having a complete solution?

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -967,7 +992,59 @@ def _find_pack_dependencies(pack_id, id_set, verbose_file, exclude_ignored_depen
...
+                dependency_graph.add_edge(pack, dependency_name)
+                if is_mandatory:
+                    dependency_graph.nodes()[dependency_name]['mandatory_for_packs'].append(pack)
```
yaakovi, line 1024: why do we need this direction?

demisto_sdk/commands/find_dependencies/find_dependencies.py
hod-alpert: In the pack_ids list there is a selection of lists with ignore list.
As far as I have seen `Base` pack is in it.

In such case - we don't want to calculate the dependencies for the things that are in the ignore list, but we do want to calculate them as dependencies.

In that case we need to add them as node to the graph.

demisto_sdk/commands/find_dependencies/find_dependencies.py
hod-alpert: In the previous way - when you calculate a graph for each pack - you can just write in it's mandatory dependencies a boolean indicator and it's obvious that. it's mandatory for the source node.

Now that we calculate the whole graph in one time - a boolean is not enough because there is no source node in this case, so the solution I had was to implement it with a list of packs that the current node is mandatory for

demisto_sdk/commands/find_dependencies/find_dependencies.py
hod-alpert: @yaakovi Any thoughts?

demisto_sdk/commands/find_dependencies/tests/find_dependencies_test.py
```diff
@@ -1119,6 +1120,60 @@ def test_search_packs_by_items_names_or_ids(item_names, section_name, expected_r
...
+        graph.add_node('pack4')
+        graph.add_edge('pack1', 'pack2')
+        graph.add_edge('pack2', 'pack3')
+        dfs_graph = PackDependencies.get_dependencies_subgraph_by_dfs(graph, source_node)
+        for i in expected_nodes_in:
+            assert i in dfs_graph.nodes()
+        for i in expected_nodes_out:
+            assert i not in dfs_graph.nodes()
+
+    def test_build_all_dependencies_graph(self, id_set, mocker):
```
Itay4, line 1150: add docstring please

demisto_sdk/commands/find_dependencies/tests/find_dependencies_test.py
```diff
@@ -1119,6 +1120,60 @@ def test_search_packs_by_items_names_or_ids(item_names, section_name, expected_r
...
+        assert [n for n in dependency_graph.neighbors('pack1')] == ['pack2', 'pack3']
+        assert [n for n in dependency_graph.neighbors('pack2')] == ['pack3']
+        assert [n for n in dependency_graph.neighbors('pack3')] == []
+        assert [n for n in dependency_graph.neighbors('pack4')] == []
+
+        # Asserting mandatory dependencies
+        nodes = dependency_graph.nodes(data=True)
+        assert nodes['pack1']['mandatory_for_packs'] == []
+        assert nodes['pack2']['mandatory_for_packs'] == ['pack1']
+        assert nodes['pack1']['mandatory_for_packs'] == []
```
Itay4, line 1174:
```suggestion
        assert nodes['pack3']['mandatory_for_packs'] == []
```

demisto_sdk/commands/find_dependencies/tests/find_dependencies_test.py
```diff
@@ -1119,6 +1120,60 @@ def test_search_packs_by_items_names_or_ids(item_names, section_name, expected_r
...
+        assert nodes['pack1']['mandatory_for_packs'] == []
```
Itay4, line 1175:
```suggestion
        assert nodes['pack4']['mandatory_for_packs'] == []
```

demisto_sdk/commands/find_dependencies/tests/find_dependencies_test.py
```diff
@@ -1119,6 +1120,60 @@ def test_search_packs_by_items_names_or_ids(item_names, section_name, expected_r
...
+                            'pack4': []}
+            return dependencies[pack_id]
+        mocker.patch(
+            'demisto_sdk.commands.find_dependencies.find_dependencies.PackDependencies._find_pack_dependencies',
+            side_effect=mock_find_pack_dependencies
+        )
+        pack_ids = ['pack1', 'pack2', 'pack3', 'pack4']
+        dependency_graph = PackDependencies.build_all_dependencies_graph(pack_ids, {}, VerboseFile(''))
+
+        # Asserting Dependencies
```
Itay4, line 1164:
```suggestion
        # Asserting Dependencies (mandatory and non-mandatory)
```

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -967,7 +992,59 @@ def _find_pack_dependencies(pack_id, id_set, verbose_file, exclude_ignored_depen
...
+        """
+        dependency_graph = nx.DiGraph()
+        for pack in pack_ids:
+            dependency_graph.add_node(pack, mandatory_for_packs=[])
+        for pack in dependency_graph.nodes():
+            dependencies = PackDependencies._find_pack_dependencies(
+                pack, id_set, verbose_file=verbose_file, exclude_ignored_dependencies=exclude_ignored_dependencies)
+            for dependency_name, is_mandatory in dependencies:
+                if dependency_name == pack:
+                    continue
```
Itay4, line 1019: see if we can cover the missed lines here in test

yaakovi (commented):
looks good to me

Itay4 (approved):

PR #857, opened 2020-11-03
fixed to Validate ID was not modified

moishce:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29413

## Description
fixed to Validate ID was not modified

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/integration.py
+++ b/demisto_sdk/commands/common/hook_validations/integration.py
@@ -351,15 +351,15 @@ class IntegrationValidator(ContentEntityValidator):
     def is_valid_beta(self):
         # type: () -> bool
         """Validate that that beta integration has correct beta attributes"""
-
+        valid_status = True
         if not all([self._is_display_contains_beta(), self._has_beta_param()]):
             self.is_valid = False
-            return False
-        if self.old_file:
+            valid_status = False
+        if not self.old_file:
             if not all([self._id_has_no_beta_substring(), self._name_has_no_beta_substring()]):
                 self.is_valid = False
-                return False
-        return True
+                valid_status = False
+        return valid_status
 
     def _id_has_no_beta_substring(self):
         # type: () -> bool
--- a/demisto_sdk/commands/common/hook_validations/integration.py
+++ b/demisto_sdk/commands/common/hook_validations/integration.py
@@ -350,7 +350,7 @@ class IntegrationValidator(ContentEntityValidator):
 
     def is_valid_beta(self):
         # type: () -> bool
-        """Validate that that beta integration has correct beta attributes"""
+        """Validate that beta integration has correct beta attributes"""
         valid_status = True
         if not all([self._is_display_contains_beta(), self._has_beta_param()]):
             self.is_valid = False
--- a/demisto_sdk/commands/common/hook_validations/structure.py
+++ b/demisto_sdk/commands/common/hook_validations/structure.py
@@ -17,7 +17,8 @@ from demisto_sdk.commands.common.errors import Errors
 from demisto_sdk.commands.common.hook_validations.base_validator import \
     BaseValidator
 from demisto_sdk.commands.common.tools import (get_content_file_type_dump,
-                                               get_remote_file)
+                                               get_remote_file,
+                                               is_file_path_in_pack)
 from demisto_sdk.commands.format.format_constants import \
     OLD_FILE_DEFAULT_1_FROMVERSION
 from pykwalify.core import Core
@@ -61,7 +62,7 @@ class StructureValidator(BaseValidator):
         self.file_type = self.get_file_type()
         self.current_file = self.load_data_from_file()
         self.fromversion = fromversion
-        if is_new_file:
+        if is_new_file or not is_file_path_in_pack(self.file_path):
             self.old_file = {}
         else:
             self.old_file = get_remote_file(old_file_path if old_file_path else file_path, tag=tag)
--- a/demisto_sdk/commands/common/hook_validations/structure.py
+++ b/demisto_sdk/commands/common/hook_validations/structure.py
@@ -62,6 +62,7 @@ class StructureValidator(BaseValidator):
         self.file_type = self.get_file_type()
         self.current_file = self.load_data_from_file()
         self.fromversion = fromversion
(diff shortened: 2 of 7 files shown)


ShahafBenYakir:
Please add explanation on how this fixes the issue.
Also, add a test for this scenario.

moishce:
After I fixed it to look for the old file then it failed on the build on validating:
_id_has_no_beta_substring, _name_has_no_beta_substring
and he checked it only for modified files
So I fixed it that would do it only for new files

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -351,15 +351,15 @@ def is_changed_subtype(self):
     def is_valid_beta(self):
         # type: () -> bool
         """Validate that that beta integration has correct beta attributes"""
```
ShahafBenYakir, line 353:
```suggestion
        """Validate that beta integration has correct beta attributes"""
```

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -351,15 +351,15 @@ def is_changed_subtype(self):
...
-            return False
-        if self.old_file:
+            valid_status = False
+        if not self.old_file:
             if not all([self._id_has_no_beta_substring(), self._name_has_no_beta_substring()]):
                 self.is_valid = False
-                return False
-        return True
+                valid_status = False
+        return valid_status
```
ShahafBenYakir, line 362: I think that after adding the `not self.old_file` this should work, so why do we need the `valid_status` argument?

demisto_sdk/tests/test_files/valid-beta-integration.yml
```diff
@@ -1,6 +1,6 @@
 category: Messaging
 commonfields:
-  id: Telegram_Beta
+  id: Telegram
```
ShahafBenYakir, line 3: This is for testing right?

moishce: right

demisto_sdk/commands/common/hook_validations/integration.py
moishce: I want to return true or false only at the end of the function so that it passes all the tests and does not return in the first failure

ShahafBenYakir: Mmmmm... why? if the first check fails the status will be `False` no matter what will happen in the second check. Is there something else happening in the second check that I am missing?

moishce: Yes, inside the functions it puts the error into the handle_error and it prints the error, and if it finishes in the first function then only the error of the first will return

ShahafBenYakir: Cool I see

roysagi (commented):
Please add a changelog entry for this.

demisto_sdk/tests/test_files/valid-beta-integration.yml
ShahafBenYakir: If we are done testing, let's revert this.

demisto_sdk/commands/common/hook_validations/structure.py
```diff
@@ -61,7 +62,7 @@ def __init__(self, file_path, is_new_file=False, old_file_path=None, predefined_
         self.file_type = self.get_file_type()
         self.current_file = self.load_data_from_file()
         self.fromversion = fromversion
-        if is_new_file or predefined_scheme:
+        if is_new_file or not is_file_path_in_pack(self.file_path):
```
ShahafBenYakir, line 65: Add a comment on what this condiition checks

demisto_sdk/commands/validate/tests/validators_test.py
```diff
@@ -381,6 +381,40 @@ def test_run_all_validations_on_file(self, _, file_path):
...
+        """
+        Given
+        - A invalid file in packs or beta integration
+
+        When
+        - running run_all_validations_on_file on that file
+
+        Then
+        -  The file will be validated and failed
+        """
```
ShahafBenYakir, line 414:
```suggestion
        """
        Given
        - An invalid file inside packs or a beta integration

        When
        - running run_all_validations_on_file on that file

        Then
        -  The file will be validated and failed
        """
```
I don't see we tested a beta integration file, we should... no?

ShahafBenYakir (commented):
Let's add a test for chagning an item ID

demisto_sdk/commands/validate/tests/validators_test.py
```diff
@@ -381,6 +381,64 @@ def test_run_all_validations_on_file(self, _, file_path):
...
+        INVALID_WIDGET_PATH,
+        INVALID_BETA_INTEGRATION
+    ]
+
+    @pytest.mark.parametrize('file_path', INVALID_FILES_PATHS_FOR_ALL_VALIDATIONS)
+    @patch.object(ImageValidator, 'is_valid', return_value=True)
+    def test_run_all_validations_on_file_failed(self, _, file_path):
+        """
+        Given
+        - An invalid file inside packs or an invalid beta integration
```
ShahafBenYakir, line 408:
```suggestion
        - An invalid file inside a pack
```

demisto_sdk/commands/validate/tests/validators_test.py
```diff
@@ -381,6 +381,64 @@ def test_run_all_validations_on_file(self, _, file_path):
...
+        """
+        Given
+        - file in packs thet id was modified
+
+        When
+        - running run_all_validations_on_file on that file
+
+        Then
+        -  The file will be validated and failed Because the id is not equal to the old file
+        """
```
ShahafBenYakir, line 429:
```suggestion
        - Integration with a modified ID.

        When
        - running 'run_all_validations_on_file' on that file.

        Then
        -  The file will fail validation and because its id changed.
        """
```

ShahafBenYakir (approved):

ShahafBenYakir (approved):

PR #860, opened 2020-11-08
fixed isArray in format

jochman:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/30208

## Description
New behavior:
If a command has already contained isArray and it's True: will preserve
if it's false: will preserve and print a warning
if there are no arguments: Will auto-add with a print.

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Fixed an issue in the **format** command where format failed when trying to update invalid Docker image.
 * The **format** command will now preserve the **isArray** argument in integration's reputation commands and will show a warning if it set to **false**.
 
+
 # v1.2.8
 * Added a new validation that checks in playbooks for the usage of `DeleteContext` in **validate** command.
 * Added a new validation that checks in playbooks for the usage of specific instance in **validate** command.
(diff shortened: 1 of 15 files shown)


Itay4:
please add test

jochman:
> please add test and update changelog

Already covered 
image

Itay4:
do we have test that it is false before format and stays that way after ?

jochman:
> do we have test that it is false before format and stays that way after ?

Yes.

jochman:
@Itay4 see test_isarray_false.
I've contained all the unittests in a class for better performance (pytest execute classes parallel)

Itay4 (approved):

Itay4 (changes requested):
update changelog

Itay4:
seems like your changelog is synced with master
also change it to something like:
`* The **format** command will now preserve the *isArray* argument in integratio reputation command and display a warning in case it is set to ***false***.`

Itay4 (approved):

PR #862, opened 2020-11-09
removed exception on pack dependencies calculation to allow for empty packs

guykeller:
## Status 
- [ ] In Progress
- [x] Ready

## Description
Currently, when a pack contains only `pack_metadata` (for example) an exception will be raised, though it is a legal pack.
This is now fixed.

### DIFF
--- a/demisto_sdk/commands/find_dependencies/find_dependencies.py
+++ b/demisto_sdk/commands/find_dependencies/find_dependencies.py
@@ -9,7 +9,7 @@ from typing import Union
 import click
 import networkx as nx
 from demisto_sdk.commands.common import constants
-from demisto_sdk.commands.common.tools import print_error
+from demisto_sdk.commands.common.tools import print_error, print_warning
 from demisto_sdk.commands.create_id_set.create_id_set import IDSetCreator
 
 MINIMUM_DEPENDENCY_VERSION = LooseVersion('6.0.0')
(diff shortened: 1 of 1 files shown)


hod-alpert (approved):

guykeller:
@hod-alpert 
From what I've seen, the code currently handles the case of an empty pack quite well.
LMK if it hurts any component you've added in your logic

hod-alpert:
> @hod-alpert
> From what I've seen, the code currently handles the case of an empty pack quite well.
> LMK if it hurts any component you've added in your logic

The only disadvantage of that is that warning print will not be in the log file for now since logging is not implemented in the sdk.
If you have any idea how to handle that - that will be great, if not - will deal with that later

yaakovi:
@guykeller do you have a content build that ran with that version of the sdk? (would like to see how the CommonReports pack looks)

guykeller:
@yaakovi No, what's the ambiguity? Seems to me like it would show as a warning

anara123:
@guykeller We change a behaviour here. Instead of failing build, we just warn and don't fail the build. Thus I agree, spin a build in content using this sdk

hod-alpert:
@yaakovi CommonReports and CommonDashboard will not fail and will not send warnings too, 
I modified the sdk to collect reports and dashboards.

The dependencies are not calculated on it (there is an open issue) but the items are collected.

guykeller:
https://github.com/demisto/content-private/runs/1391162072?check_suite_focus=true
@anara123

anara123 (approved):

guykeller:
https://app.circleci.com/pipelines/github/demisto/content/43367/workflows/13a14a2d-e53a-4606-bb19-c237b87b1e7f/jobs/187131
Nightly passed successfully, merging the PR

PR #868, opened 2020-11-11
Contribution Convert Refactor

avidan-H:
## Status
- [x] Ready

## Description
Contribution conversion has been split out into its own file/class under the `init` command directory. Furthermore the `init` command arguments related to contribution zip file conversion have been removed as this functionality is not intended to be used from the command line but rather only as an import to other code. While the convenience function of `convert_contribution_to_pack` has been left (in the new class `ContributionConverter`) for the simple case of creating a new pack from a contribution, the functionality has been broken down into multiple functions which allow for code that imports this class to handle different cases (such as when updating an existing pack) by invoking those functions as per their own requirements.

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -649,7 +649,7 @@ def generate_test_playbook(**kwargs):
     '-h', '--help'
 )
 @click.option(
-    "-n", "--name", help="The name of the directory and file you want to create")
+    "-n", "--name", help="The name of the directory and file you want to create", required=True)
 @click.option(
     "--id", help="The id used in the yml file of the integration or script"
 )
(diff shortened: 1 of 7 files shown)


Itay4 (approved):

avidan-H:
@Itay4 I think this is in an ok state now. The last commit should resolve the issue where integration/script/playbook `README.md` package files were being overwritten with empty readmes in the update flow.

1. An unresolved issue is that the non-interactive format overrides a yml's `tests` section so that the value is `no tests (autoformatted)` - the formatting should leave the value that was in the yml file.
2. A nice enhancement would be to preserve the yml file's layout (assuming that the format of the existing file is technically valid) since formatting the file can result in a misleadingly large diff that is mostly due to slight textual formatting differences and field rearrangements when the contents are valid and more or less the same.

PR #882, opened 2020-11-18
support mappers in download command

guyfreund:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/30698

## Description
Support mappers in download command


CHANGELOG.md
```diff
@@ -171,6 +171,7 @@
 * Fixed an issue where **generate-docs** was inserting unnecessary escape characters.
 * Fixed an issue in the **update-release-notes** command where changes to the pack_metadata were not detected.
 * Fixed an issue where **validate** did not check for missing release notes in old format files.
+* Updated **download** command to support mappers.
```
ShahafBenYakir, line 175:
```suggestion
* Updated the **download** command to support mappers.
```

demisto_sdk/commands/download/downloader.py
```diff
@@ -189,7 +189,7 @@ def handle_list_files_flag(self) -> bool:
         """
         if self.list_files:
             self.all_custom_content_objects = self.get_custom_content_objects()
-            list_files = [[cco['name'], cco['entity'][:-1]] for cco in self.all_custom_content_objects
+            list_files = [[cco['name'], cco['type']] for cco in self.all_custom_content_objects
```
ShahafBenYakir, line 192: why did we change this? what is in `'type'`?

demisto_sdk/commands/download/downloader.py
```diff
@@ -691,9 +691,8 @@ def update_data(file_path_to_write: str, file_path_to_read: str, file_ending: st
             file_data: dict = get_json(file_path_to_write)
             if pack_obj_data:
                 merge(file_data, preserved_data)
-            json_depth: int = get_depth(file_data)
             with open(file_path_to_write, 'w') as jf:
-                json.dump(obj=file_data, fp=jf, indent=json_depth)
+                json.dump(obj=file_data, fp=jf, indent=4)
```
ShahafBenYakir, line 695: Why have we abandoned the getJson function? I remember we needed it in the past

demisto_sdk/commands/download/downloader.py
guyfreund: what we get from `find_type` function. we changed it because its much more accurate to use the file type than the entity dir. for example: mapper & classifier are in the same entity dir (Classifiers) but have diff type

demisto_sdk/commands/download/downloader.py
guyfreund: are you talking of `get_depth` function? if so then yes, it was a mistake to use it.

ShahafBenYakir (approved):

roysagi (approved):

PR #901, opened 2020-11-25
fixing wrong v2 integration validation

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Description
Integrations should be recognized as v2 via the `display` field and not `name`.


demisto_sdk/commands/common/tools.py
```diff
@@ -1440,3 +1440,19 @@ def arg_to_list(arg: Union[str, List[str]], separator: str = ",") -> List[str]:
             return json.loads(arg)
         return [s.strip() for s in arg.split(separator)]
     return [arg]
+
+
+def is_v2_file(current_file, check_in_display=False):
```
Itay4, line 1445: please add type hints

Itay4 (approved):

PR #905, opened 2020-11-26
Validate content package phase1

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/28473
fixes: https://github.com/demisto/etc/issues/32733
fixes: https://github.com/demisto/etc/issues/33132

## Description
adding git to content objects and moving validate git handling to content objects


demisto_sdk/commands/common/content/content.py
```diff
@@ -127,3 +128,143 @@ def content_descriptor(self) -> Optional[ContentDescriptor]:
...
+            content_repo.active_branch).iter_change_type('M')}
+
+        committed: Dict = {path.name: Pack(path) for path in committed_set}
+
+        if committed_only:
+            return committed
+
+        staged_set = {Path(os.path.join(*Path(item.a_path).parts[:Path(item.a_path).parts.index('Packs') + 2])) for item
+                      in content_repo.head.commit.diff(
+            paths=list((Path().cwd() / 'Packs').glob('*/'))).iter_change_type('M')}
```
yaakovi, line 149: we should also mark renamed files as modified when the rename is not 100%

not sure if this implementation collects them.

demisto_sdk/commands/common/content/content.py
```diff
@@ -127,3 +128,143 @@ def content_descriptor(self) -> Optional[ContentDescriptor]:
...
+        prev_ver = prev_ver.replace('origin/', '')
+        content_repo: Repo = self.git()
+
+        # staging all local changes
+        if not no_auto_stage:
+            content_repo.git.add('.')
+
+        committed = {(Path(item.a_path), Path(item.b_path)) for item
+                     in content_repo.remote().refs[prev_ver].commit.diff(
+            content_repo.active_branch).iter_change_type('R')}
```
yaakovi, line 258: rename should be considered only when it is 100%

roysagi: what do you mean by 100%? cause the distinction of R vs M is a git standard - I separated the two to be inline with the git terms as well as handle the renamed format (tuple instead of the standard string) as a separate function.

demisto_sdk/__main__.py
```diff
@@ -875,8 +880,9 @@ def update_pack_releasenotes(**kwargs):
     try:
         validate_manager = ValidateManager(skip_pack_rn_validation=True, prev_ver=prev_ver)
         validate_manager.setup_git_params()
-        modified, added, old, changed_meta_files, _packs = validate_manager.get_modified_and_added_files(
-            '...', prev_ver)
+        modified, added, old, changed_meta_files = validate_manager.get_changed_files_from_git()
+        _packs = get_packs(modified).union(get_packs(old)).union(
```
Itay4, line 884: why is the `_` needed here?

demisto_sdk/commands/common/content/content.py
```diff
@@ -15,9 +16,12 @@
     ContentDescriptor, Documentation)
 from demisto_sdk.commands.common.content.objects_factory import \
     path_to_pack_object
-from git import InvalidGitRepositoryError, Repo
+from demisto_sdk.commands.common.tools import find_type
+from git import GitCommandError, InvalidGitRepositoryError, Repo
 from wcmatch.pathlib import Path
 
+CIRCLE_CONTENT_ROOT_PATH = os.path.abspath(os.path.join(__file__, '../../..'))
```
Itay4, line 23: is this const used?

demisto_sdk/commands/common/content/content.py
```diff
@@ -127,3 +132,185 @@ def content_descriptor(self) -> Optional[ContentDescriptor]:
             descriptor_object = ContentDescriptor(path)
 
         return descriptor_object
+
+    def modified_packs(self, prev_ver='master', committed_only=False, staged_only=False) -> Dict[str, Pack]:
```
Itay4, line 136: please add type hints and docstring
also, i think we can improve the func name a bit, `get_modified_packs` ?

demisto_sdk/commands/common/content/content.py
```diff
@@ -127,3 +132,185 @@ def content_descriptor(self) -> Optional[ContentDescriptor]:
...
+        content_repo: Repo = self.git()
```
Itay4, line 137: perhaps this should be a class attribute or it makes sense to initialize it for every func ?

demisto_sdk/commands/common/content/content.py
```diff
@@ -127,3 +132,185 @@ def content_descriptor(self) -> Optional[ContentDescriptor]:
...
+
+        # staging all local changes
+        self._stage_files(content_repo)
+
+        committed_set = {Path(os.path.join(*Path(item.a_path).parts[:Path(item.a_path).parts.index('Packs') + 2]))
```
Itay4, line 142: this code line is too loaded imo, can we break it?

demisto_sdk/commands/common/content/content.py
```diff
@@ -127,3 +132,185 @@ def content_descriptor(self) -> Optional[ContentDescriptor]:
...
+        if committed_only:
+            return committed
+
+        staged_set = {Path(os.path.join(*Path(item.a_path).parts[:Path(item.a_path).parts.index('Packs') + 2])) for item
+                      in content_repo.head.commit.diff(
+            paths=list((Path().cwd() / 'Packs').glob('*/'))).iter_change_type('M')}
+
+        staged: Dict = {path.name: Pack(path) for path in staged_set}
+
+        if staged_only:
```
Itay4, line 157: no biggy and im not sure if it optimiizes but i think we can avoid calculating both commited and staged if only stages is required

demisto_sdk/commands/common/content/content.py
```diff
@@ -127,3 +132,185 @@ def content_descriptor(self) -> Optional[ContentDescriptor]:
...
+            return staged
+
+        staged.update(committed)
+
+        return staged
+
+    def added_packs(self, prev_ver='master', committed_only=False, staged_only=False) -> Dict[str, Pack]:
```
Itay4, line 164: did you consider merging the code duplication of this func and the previous ?

demisto_sdk/commands/common/content/content.py
```diff
@@ -127,3 +132,185 @@ def content_descriptor(self) -> Optional[ContentDescriptor]:
...
+
+        if staged_only:
+            return staged
+
+        staged.update(committed)
+
+        return staged
+
+    def modified_files(self, prev_ver='master', committed_only=False, staged_only=False,
+                       no_auto_stage=False) -> Set[Path]:
```
Itay4, line 193: should we have `no_auto_stage` in packs funcs as well?

demisto_sdk/commands/common/git_tools.py
```diff
@@ -37,3 +50,273 @@ def get_changed_files(from_branch: str = 'master', filter_results: Callable = No
...
+    # If git base not provided - check against origin/prev_ver unless using release branch
+    if '/' not in prev_ver and not (branch_name.startswith('20.') or branch_name.startswith('21.')):
+        prev_ver = 'origin/' + prev_ver
+    return prev_ver
+
+
+def filter_staged_only(modified_files, added_files, old_format_files, changed_meta_files):
+    """The function gets sets of files which were changed in the current branch and filters
+    out only the files that were changed in the current commit"""
+    all_changed_files = run_command('git diff --name-only --staged').split()
```
Itay4, line 67: does it make sense to use the git package here as well ?

demisto_sdk/commands/common/git_tools.py
```diff
@@ -37,3 +50,273 @@ def get_changed_files(from_branch: str = 'master', filter_results: Callable = No
...
+            if (not is_origin_demisto and not remote_configured) and not no_configuration_prints:
+                error_message, error_code = Errors.changes_may_fail_validation()
+                base_validator.handle_error(error_message, error_code, file_path="General-Error", warning=True,
+                                            drop_line=True)
+
+            if not no_configuration_prints and not staged:
+                click.echo("Collecting all local changed files against the content master")
+
+            # only changes against prev_ver (without local changes)
+            all_changed_files_string = run_command('git diff --name-status {}'.format(prev_ver))
```
Itay4, line 150: lets use f-strings instead of format

demisto_sdk/commands/common/git_tools.py
```diff
@@ -37,3 +50,273 @@ def get_changed_files(from_branch: str = 'master', filter_results: Callable = No
...
+    modified_files = modified_files.intersection(formatted_changed_files)
+    added_files = added_files.intersection(formatted_changed_files)
+    old_format_files = old_format_files.intersection(formatted_changed_files)
+    changed_meta_files = changed_meta_files.intersection(formatted_changed_files)
+    return modified_files, added_files, old_format_files, changed_meta_files
+
+
+def get_modified_and_added_files(compare_type,
+                                 prev_ver,
+                                 ignored_errors=dict(),
```
Itay4, line 84: it is better practice to set `None` as default value instead of empty object

demisto_sdk/commands/common/hook_validations/pack_unique_files.py
```diff
@@ -326,7 +328,10 @@ def _is_approved_tags(self) -> bool:
 
     def get_master_private_repo_meta_file(self, metadata_file_path: str):
         current_repo = Repo(Path.cwd(), search_parent_directories=True)
-        old_meta_file_content = current_repo.git.show(f'master:{metadata_file_path}')
+        try:
+            old_meta_file_content = current_repo.git.show(f'master:{metadata_file_path}')
+        except GitCommandError:
```
Itay4, line 333: lets print in verbose here

demisto_sdk/__main__.py
```diff
@@ -888,8 +889,9 @@ def update_pack_releasenotes(**kwargs):
     try:
         validate_manager = ValidateManager(skip_pack_rn_validation=True, prev_ver=prev_ver)
         validate_manager.setup_git_params()
-        modified, added, old, changed_meta_files, _packs = validate_manager.get_modified_and_added_files(
-            '...', prev_ver)
+        modified, added, changed_meta_files, old = validate_manager.get_changed_files_from_git()
+        _packs = get_packs(modified).union(get_packs(old)).union(
+            get_packs(added))
```
bakatzir, line 894: WDYT?
```suggestion
        _packs = get_packs(modified + old+ added)
```

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -104,6 +109,18 @@ def __init__(
...
+        try:
+            self.git_util = GitUtil(repo=Content.git())
+            self.branch_name = self.git_util.get_current_working_branch()
+        except InvalidGitRepositoryError:
+            if self.use_git:
+                raise
+            else:
+                self.git_util = None  # type: ignore
+                self.branch_name = ''
+                pass
```
bakatzir, line 122: how is this ok? can you explain how you will be able to execute `self.get_changed_files_from_git()` in this case?
an inline comment in the code will help.

demisto_sdk/tests/integration_tests/validate_integration_test.py
```diff
@@ -4,7 +4,7 @@
 from demisto_sdk.__main__ import main
 from demisto_sdk.commands.common import tools
 from demisto_sdk.commands.common.constants import DEFAULT_IMAGE_BASE64
-from demisto_sdk.commands.common.git_tools import git_path
+from demisto_sdk.commands.common.git_util import GitUtil
```
bakatzir, line 7: I might have missed it, but should we add 1/2 cases for unified integrations and scripts validate flow?
Just as you've changes in these flows as well in `validate_manager` and I do not think they have integration tests.

demisto_sdk/commands/common/tests/git_tools_test.py
```diff
@@ -0,0 +1,97 @@
+import os
```
bakatzir, line 1: testing only the `legacy` tools, no?
consider renaming the file.

demisto_sdk/commands/common/git_util.py
```diff
@@ -0,0 +1,319 @@
...
+                            self.repo.commit('HEAD~1').diff().iter_change_type(requested_status)}
+                else:
+                    return {(Path(item.a_path), Path(item.b_path)) for item in
+                            self.repo.commit('HEAD~1').diff().iter_change_type(requested_status)}
+            except gitdb.exc.BadName:
+                # in case no last commit exists - just pass
+                pass
+        return set()
+
+    def check_if_remote_exists(self, remote):
```
bakatzir, line 296: should not it be a closure func inside `_handle_prev_ver`?

demisto_sdk/commands/common/legacy_git_tools.py
```diff
@@ -0,0 +1,325 @@
...
+    return git_path.replace('\n', '')
+
+
+def get_current_working_branch() -> str:
+    branches = run_command('git branch')
+    branch_name_reg = re.search(r'\* (.*)', branches)
+    if branch_name_reg:
+        return branch_name_reg.group(1)
+
+    return ''
```
bakatzir, line 31: these funcs are defined twice. :bydesign:?

demisto_sdk/__main__.py
roysagi: It was originaly like that, not sure why it is needed but I don't think I should change that in this PR which is large in of itself.

demisto_sdk/commands/common/git_util.py
roysagi: no, this check is used outside of this internal use as well in order to check if a given remote exists if not I print an error to the user and revert to the default remote.

demisto_sdk/commands/common/legacy_git_tools.py
roysagi: you mean withing the git_util and the legacy? yes, we are yet to completely rid of the legacy git tools so we should have the option to get the branch name outside of the git util.

demisto_sdk/commands/common/tests/git_tools_test.py
roysagi: done

demisto_sdk/commands/validate/validate_manager.py
roysagi: Note - if the exception occurs and we use git - we raise it so it will never get to `self.get_changed_files_from_git()`

Anyway I'll add a comment.

demisto_sdk/tests/integration_tests/validate_integration_test.py
roysagi: you might be right but I think this is unrelated to this change we should do this as a separate task in my opinion.

bakatzir: As we had an issue with that last week, I think you should add it, but your call.

bakatzir (approved):

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -838,265 +856,162 @@ def validate_no_missing_release_notes(self, modified_files, old_format_files, ad
...
-                if (not is_origin_demisto and not remote_configured) and not self.no_configuration_prints:
-                    error_message, error_code = Errors.changes_may_fail_validation()
-                    self.handle_error(error_message, error_code, file_path="General-Error", warning=True,
-                                      drop_line=True)
+            if self.skip_dependencies:
+                click.echo("Skipping pack dependencies check")
 
-                if not self.no_configuration_prints and not self.staged:
-                    click.echo("Collecting all local changed files against the content master")
+    def get_changed_files_from_git(self):
```
Itay4, line 919: please add return type hint and docstring

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -838,265 +857,170 @@ def validate_no_missing_release_notes(self, modified_files, old_format_files, ad
...
+                    filtered_set.add(formatted_path)
 
-                # ignore changes in TESTS_DIRECTORIES files.
-                elif any(test_dir in file_path for test_dir in TESTS_DIRECTORIES):
-                    if file_path not in self.ignored_files:
-                        self.ignored_files.add(file_path)
-                        if print_ignored_files:
-                            click.secho('Ignoring file path: {} - test file'.format(file_path), fg="yellow")
-                    continue
+            # handle a case where a file was deleted locally though recognised as added against master.
```
yaakovi, line 985: nice

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -838,265 +857,170 @@ def validate_no_missing_release_notes(self, modified_files, old_format_files, ad
...
-                # identify added files
-                elif file_status.lower() == 'a' and find_type(file_path) and not file_path.startswith('.'):
-                    added_files_list.add(file_path)
-                # identify renamed files
-                elif file_status.lower().startswith('r') and find_type(file_path):
-                    # if a code file changed, take the associated yml file.
-                    if find_type(file_data[2]) in [FileType.POWERSHELL_FILE, FileType.PYTHON_FILE]:
-                        modified_files_list.add(file_path)
+        # redirect non-test code files to the associated yml file
+        if file_type in [FileType.PYTHON_FILE, FileType.POWERSHELL_FILE]:
```
yaakovi, line 1003: add JS as well (for rest of the code):
image

yaakovi (commented):
nice work!

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -105,6 +108,20 @@ def __init__(
...
+            self.branch_name = self.git_util.get_current_working_branch()
+        except InvalidGitRepositoryError:
+            # if we are using git - fail the validation by raising the exception.
+            if self.use_git:
+                raise
+            # if we are not using git - simply move on.
+            else:
+                self.git_util = None  # type: ignore
+                self.branch_name = ''
+                pass
```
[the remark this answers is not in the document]
roysagi: done

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -838,265 +857,170 @@ def validate_no_missing_release_notes(self, modified_files, old_format_files, ad
...
+            error_message, error_code = Errors.running_on_master_with_git()
+            if self.handle_error(error_message, error_code, file_path='General',
+                                 warning=(not self.is_external_repo or self.is_circle), drop_line=True):
+                return False
+        return True
 
-        Returns:
-            tuple. 3 sets representing modified files, added files and files of old format who have changed.
-        """
+    def git_config_prints(self):
```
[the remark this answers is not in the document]
roysagi: done

demisto_sdk/commands/validate/validate_manager.py
roysagi: thanks

demisto_sdk/commands/validate/validate_manager.py
roysagi: adding

demisto_sdk/commands/common/git_util.py
```diff
@@ -0,0 +1,319 @@
...
+        """Get all the files that were changed in the last commit of a given type when checking a branch against itself.
+        Args:
+            prev_ver (str): The base branch against which the comparison is made.
+            requested_status (str): M, A, R, D - the git status to return
+        Returns:
+            Set: of Paths to files changed in the the last commit or an empty set if not
+            running on master against master.
+        """
+        # when checking branch against itself only return the last commit.
+        if self.get_current_working_branch() == prev_ver:
```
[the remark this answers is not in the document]
roysagi: done

demisto_sdk/commands/common/git_util.py
```diff
@@ -0,0 +1,319 @@
...
+                else:
+                    return {(Path(item.a_path), Path(item.b_path)) for item in
+                            self.repo.commit('HEAD~1').diff().iter_change_type(requested_status)}
+            except gitdb.exc.BadName:
+                # in case no last commit exists - just pass
+                pass
+        return set()
+
+    def check_if_remote_exists(self, remote):
+        if '/' in remote:
```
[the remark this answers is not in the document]
roysagi: it is just a check for the user that if he wants to check against a branch which is not under `origin` I check that the remote is defined in his env, if not it reverts to the default remote in the env (and prints that error to the user).

It effectively acts the same as the current git diff but it is another layer of validation to make sure we are running in a good manner.

demisto_sdk/commands/common/git_util.py
```diff
@@ -0,0 +1,319 @@
...
+        Args:
+            prev_ver (str): The base branch against which the comparison is made.
+            committed_only (bool): Whether to return only committed files.
+            staged_only (bool): Whether to return only staged files.
+        Returns:
+            Set: A set of Paths to the modified files.
+        """
+        remote, branch = self._handle_prev_ver(prev_ver)
+
+        # when checking branch against itself only return the last commit.
```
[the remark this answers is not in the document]
roysagi: when the prev-ver is the same as your current branch - for example master against master - then we check only the last commit (HEAD~1) - if this is not the case `_only_last_commit` will return an empty set and the function will continue.

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -838,265 +856,162 @@ def validate_no_missing_release_notes(self, modified_files, old_format_files, ad
...
-
-        Args:
-            compare_type (str): whether to run diff with two dots (..) or three (...)
-            prev_ver (str): Against which branch to run the comparision - master/last release
+        # on master don't check RN
+        elif self.branch_name == 'master':
+            self.skip_pack_rn_validation = True
+            error_message, error_code = Errors.running_on_master_with_git()
+            if self.handle_error(error_message, error_code, file_path='General',
+                                 warning=(not self.is_external_repo or self.is_circle), drop_line=True):
```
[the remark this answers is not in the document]
yaakovi: you set it as a warning if used in CircleCI so builds should be ok

yaakovi (approved):

PR #916, opened 2020-12-02
added UT -  checks if validate fails when non-required param changed to…

abaumgarten:
## Status
- [] In Progress
- [x] Ready

## Description
added UT that checks if validate command fails when an non-required integration param was changed to required.


abaumgarten:
Build is failing on the error from here: demisto/etc#31320

roysagi (approved):

PR #920, opened 2020-12-06
added a flag --skip-id-set-creation for validate command

abaumgarten:
## Status
- [] In Progress
- [x] Ready

## Description
Added a flag --skip-id-set-creation for validate
Added the capability to run validate command without creating id_set validation.

### DIFF
--- a/demisto_sdk/commands/validate/validate_manager.py
+++ b/demisto_sdk/commands/validate/validate_manager.py
@@ -87,7 +87,7 @@ class ValidateManager:
         self.print_ignored_files = print_ignored_files
         self.print_ignored_errors = print_ignored_errors
         self.skip_dependencies = skip_dependencies or not use_git
-        self.skip_id_set_creation = skip_id_set_creation
+        self.skip_id_set_creation = skip_id_set_creation or self.skip_dependencies
         self.compare_type = '...'
         self.staged = staged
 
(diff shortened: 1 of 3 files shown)


abaumgarten:
Build is failing on the error from here: demisto/etc#31320

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1170,7 +1172,7 @@ def get_id_set_file(skip_dependencies, id_set_path):
         """
         id_set = {}
         if not os.path.isfile(id_set_path):
-            if not skip_dependencies:
+            if not skip_id_set_creation:
```
guykeller, line 1175: 1. Let's make this one condition
2. Still not seeing why this is needed - when do we not want to create an ID set, given no path to an existing ID set has been specified?

roysagi: @guykeller - some users do not need the id_set validation also this is in use in some test cases. This is an extra feature for this command so the capability to turn it off is a good option.

roysagi (approved):
@abaumgarten - please check that this change makes the script id validation for playbooks run even when running validate with `-a`

demisto_sdk/commands/validate/validate_manager.py
guykeller: So let's add that to the PR, so people will be able to know why it was created

PR #928, opened 2020-12-07
added an exception if Dockerfile exists

orhovy:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes:  

## Description
The problem was that if a Dockerfile exists in content repo we exit with an exception which is written under flake8 while flake8 had nothing to do with the exception.
After this fix flake8 wont be changed, and also lint wont fail on existing docker file.
It will write an error in the traceback regarding an existing file and will run on that file instead of a new file

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Added support for *prev-ver* flag in **update-release-notes** command.
 * Improved retry support when building docker images for linting.
 * Added the *--skip-id-set-creation* flag to **validate** command in order to add the capability to run validate command without creating id_set validation.
+* Fixed an issue with existing Dockerfile in the **lint** command.
 
 # 1.2.11
 * Fixed an issue where the ***generate-docs*** command reset the enumeration of line numbering after an MD table.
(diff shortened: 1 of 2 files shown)


glicht:
@orhovy I think this is the wrong approach. We should have a generic exception catching. Probably in the outside function. That handles the exception properly. I don't want to add code as you added to every line that might throw an exception.

orhovy:
> @orhovy I think this is the wrong approach. We should have a generic exception catching. Probably in the outside function. that handles the exception properly. I don't want to add code as you added to every line that might throw an exception.

We are planning on doing so, I am opening an dev task regarding adding an generic exception to the lint and to its report.
However currently if such thing will happend flake8 will be reported as the linter that failed with None as a text.
I would like to fix that as the issue says.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Added support for *prev-ver* flag in **update-release-notes** command.
 * Improved retry support when building docker images for linting.
 * Added the *--skip-id-set-creation* flag to **validate** command in order to add the capability to run validate command without creating id_set validation.
+* Fixed an issue with existing Dockerfile in the **lint** command.
```
roysagi, line 8:
```suggestion
* Fixed an issue where an existing Dockerfile would raise an error in the **lint** command.
```

roysagi (approved):

glicht:
So let's do the generic exception handling instead of this code. As what we are doing here just pollutes the code and makes it more complicated.

orhovy:
closing , as will be handled differently as suggested in a different pr with some additional changes

PR #931, opened 2020-12-08
Do not enforce package-dir on deprecated content

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31099

## Must have
- [x] Tests


demisto_sdk/commands/validate/tests/validators_test.py
```diff
@@ -897,6 +897,20 @@ def test_validate_no_old_format__without_toversion(self, mocker):
...
+        """
+            Given:
+                - an old format_file with deprecated: true
+            When:
+                - running validate_no_old_format on the file
+            Then:
+                - return a True as the file is valid
+        """
+        validate_manager = ValidateManager()
+        old_format_files = {"demisto_sdk/tests/test_files/Unifier/SampleScriptPackage/"
```
Itay4, line 910: please avoid using files as test data, use the `repo` fixture instead

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -685,7 +685,9 @@ def validate_no_old_format(self, old_format_files):
         for file_path in old_format_files:
             click.echo(f"Validating old-format file {file_path}")
             yaml_data = get_yaml(file_path)
-            if 'toversion' not in yaml_data:  # we only fail on old format if no toversion (meaning it is latest)
+            # we only fail on old format if no toversion (meaning it is latest)
+            if 'toversion' not in yaml_data and 'deprecated' not in yaml_data or \
```
Itay4, line 689: would be a bit more elegant to something like:
```python
not yaml_data.get('deprecated')
```

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -685,7 +685,9 @@ def validate_no_old_format(self, old_format_files):
         for file_path in old_format_files:
             click.echo(f"Validating old-format file {file_path}")
             yaml_data = get_yaml(file_path)
-            if 'toversion' not in yaml_data:  # we only fail on old format if no toversion (meaning it is latest)
+            # we only fail on old format if no toversion (meaning it is latest)
+            if 'toversion' not in yaml_data and not yaml_data.get('deprecated') or \
+                    ('deprecated' in yaml_data and yaml_data['deprecated'] is False):
```
Itay4, line 690: i think you dont need the `or` clause
also need test for the second case

Shellyber: I think there are cases when deprecated is in the yml but assigned to False.
Better safe then sorry

Itay4: yea i understand, then the value of `yaml_data.get('deprecated')` will be `False` and it will get in the condition

demisto_sdk/commands/validate/tests/validators_test.py
```diff
@@ -897,6 +897,32 @@ def test_validate_no_old_format__without_toversion(self, mocker):
         assert not validate_manager.validate_no_old_format(old_format_files)
         assert handle_error_mock.call_count == 2
 
+    def test_validate_no_old_format_deprecated_content(self, repo):
```
Itay4, line 900: please also add test for script for script/integration without the `deprecated` field in the yml

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -685,7 +685,8 @@ def validate_no_old_format(self, old_format_files):
         for file_path in old_format_files:
             click.echo(f"Validating old-format file {file_path}")
             yaml_data = get_yaml(file_path)
-            if 'toversion' not in yaml_data:  # we only fail on old format if no toversion (meaning it is latest)
+            # we only fail on old format if no toversion (meaning it is latest)
```
Itay4, line 688: need to update the comment here

Itay4 (approved):

PR #941, opened 2020-12-14
Check all Params and Args are implemented in the Code

orhovy:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/19667

## Description
Added to lint gathering facts the args and params value which is gathered from the yml file.
Added 2 base validations to xsoar linter that validate that all occurrences of an param and of an argument are implemented in the code. (if there are some args of the name limit in different commands we would make sure that all occurrences are implemented)
Circle run on all content: 
https://app.circleci.com/pipelines/github/demisto/content/56249/workflows/dcd6e3d7-f2e1-4035-b79d-61ac746ec0af

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


bakatzir (commented):
Very good work.

I'm contemplating if there is a case where we would call an arg/param more than once[so the sol with the remove is not rock solid], and the need to add it as an allowed error code. 
Before we think of that, please make sure the run on the content repo is successful, and let's see how many failing integrations/scripts we have.

Regardless, let's add an integration/workflow test.

orhovy:
> Very good work.
> 
> I'm contemplating if there is a case where we would call an arg/param more than once[so the sol with the remove is not rock solid], and the need to add it as an allowed error code.
> Before we think of that, please make sure the run on the content repo is successful, and let's see how many failing integrations/scripts we have.
> 
> Regardless, let's add an integration/workflow test.

I have added an integration test to xsoar_linter_integration_tests.

demisto_sdk/commands/common/constants.py
```diff
@@ -939,6 +934,22 @@ class PB_Status:
...
 
+FEED_OPTIONAL_PARAMS_SERVER_IMP = [
+    {
+        'additionalinfo': "Incremental feeds pull only new or modified indicators that have been sent from the"
+                          "integration. As the determination if the indicator is new or modified happens on the"
+                          " 3rd-party vendor's side, and only indicators that are new or modified are sent to"
+                          " Cortex XSOAR, all indicators coming from these feeds are labeled new or modified.",
+        'display': 'Incremental Feed',
+        'name': 'feedIncremental',
+        'options': ['RED', 'AMBER', 'GREEN', 'WHITE'],
```
bakatzir, line 945:
```suggestion
```

demisto_sdk/tests/workflow_test.py
```diff
@@ -106,7 +106,7 @@ def run_validations(self):
         """
         Run all of the following validations:
         * secrets
-        * lint -g --no-test
```
roysagi, line 109: Why take out the xsoar linter here

demisto_sdk/commands/common/constants.py
orhovy: This causes validate failure, the options are checked and should be kept

PR #950, opened 2020-12-20
Fix windows eol issue

guykeller:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/31249

## Description
Currently, running `demisto-sdk format` on some cases (Windows machines in particular) identifies eol characters that are automatically added on Windows as changes.
That causes format to detect all files as changed.
This behavior is now fixed.

@idovandijk Please try this branch on your environment and use the new argument. Let me know if this solves your issue.


roysagi (commented):
@guykeller don't we need to make this change in `validate` as well?

guykeller:
@roysagi We do, but let's verify first that @idovandijk doesn't experience the issue with this branch

idovandijk:
@guykeller it looks like the detection is correct but `format` is not really changing the file 
image

demisto_sdk/commands/format/format_module.py
```diff
@@ -89,7 +90,11 @@ def format_manager(input: str = None,
         files = get_files_in_dir(input, ['json', 'yml', 'py'])
     else:
         files = [file['name'] for file in
-                 get_changed_files(filter_results=lambda _file: not _file.pop('status') == 'D')]
+                 get_changed_files(from_branch='origin/master',
```
yaakovi, line 93: @roysagi this change solved an issue but may cause problems for contributors

guykeller: Perhaps we can take the branch as an optional parameter.
Not the prettiest solution but doable @yaakovi @roysagi

PR #966, opened 2020-12-23
Check all commands implemented  are implemented in the code-  XSOAR Linter validation

orhovy:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes:  https://github.com/demisto/etc/issues/27018

## Description
Added to lint gathering facts the commands which is gathered from the yml file.
Added a base validation to xsoar linter that validate that all commands from yml are implemented in the code using two most common ways:
1 - a dict of commands as keys and function as values
2 - a long if elif else section when each time check if command equals to the command from yml
Circle run on all content: 
https://app.circleci.com/pipelines/github/demisto/content/57987/workflows/a2b01779-b394-47d7-8fed-2ff128349a14

## Screenshots

## Must have
- [x] Tests
- [x] Documentation


demisto_sdk/commands/lint/linter.py
```diff
@@ -403,6 +405,8 @@ def _run_xsoar_linter(self, py_num: float, lint_files: List[Path]) -> Tuple[int,
                 myenv['LONGRUNNING'] = 'True'
             if py_num < 3:
                 myenv['PY2'] = 'True'
+            myenv['commands'] = ','.join([str(elem) for elem in self._facts['commands']]) \
+                if self._facts['commands'] else ''
             stdout, stderr, exit_code = run_command_os(
                 command=build_xsoar_linter_command(lint_files, py_num, self._facts.get('support_level', 'base')),
```
bakatzir, line 411: can't the `commands` be passed here?

demisto_sdk/commands/lint/resources/pylint_plugins/base_checker.py
```diff
@@ -35,6 +41,7 @@ class CustomBaseChecker(BaseChecker):
 
     def __init__(self, linter=None):
         super(CustomBaseChecker, self).__init__(linter)
+        self.commands = os.getenv('commands').split(',') if os.getenv('commands') else []
```
bakatzir, line 44: please add an explanation why it is from the os.getenv and you can not pass it from the `_facts`

demisto_sdk/commands/lint/linter.py
orhovy: No , xsoar linter runs as a pylint plugin , each support level is a standalone plugin.
Pylints plugins are running inside the pylint code, this is the reason why we cant use our constants and fact inside it.

demisto_sdk/commands/lint/resources/pylint_plugins/base_checker.py
orhovy: added in the linter file, when setting the env vars.

bakatzir (approved):
Since we already have it in your other approved PR, please make sure in the code we do not run this validation on scripts.

orhovy:
@bakatzir this validation wont run on scripts as scripts do not have `commands:` key under `script:` in the yml file.

PR #967, opened 2020-12-24
remove created field from init command

barchen1:
## Status
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/31105

## Description
Remove `created` field from sdk-init command

## Must have
- [x] Tests
- [x] Documentation


roysagi (approved):

demisto_sdk/commands/common/hook_validations/pack_unique_files.py
```diff
@@ -240,13 +240,14 @@ def _is_pack_meta_file_structure_valid(self):
...
             # check created field in iso format
             created_field = metadata.get(PACK_METADATA_CREATED, '')
-            if not self.check_timestamp_format(created_field):
-                suggested_value = parser.parse(created_field).isoformat() + "Z"
-                if self._add_error(
-                        Errors.pack_timestamp_field_not_in_iso_format(PACK_METADATA_CREATED,
-                                                                      created_field, suggested_value),
-                        self.pack_meta_file):
-                    return False
+            if created_field:
```
guykeller, line 243: When will this condition be true?

guykeller (approved):
essentially approved, but would love an answer on that one comment I added

demisto_sdk/commands/common/hook_validations/pack_unique_files.py
barchen1: It will be true if the pack_metadata file includes `created` field.

We dont remove `created` field from old content, so we still need to have this check, but we first need to check that there is a `created` field in the pack_metadata, if so we want to check that it is in iso format.

PR #969, opened 2020-12-27
init templates

dantavori:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/28448
relates: https://github.com/demisto/etc/issues/30900

## Description

## Screenshots

## Must have
- [x] Tests
- [x] Documentation


roysagi (approved):
Looks good see my comments.

demisto_sdk/commands/init/templates/BaseIntegration/BaseIntegration.yml
```diff
@@ -0,0 +1,52 @@
...
+    - default: false
+      description: '[Enter a description of the argument, including any important information users need to know, for example, default values.]'
+      isArray: false
+      name: dummy
+      required: true
+      secret: false
+    deprecated: false
+    description: '[Enter a description of the command, including any important information users need to know, for example required permissions.]'
+    execution: false
+    name: baseintegration-dummy
```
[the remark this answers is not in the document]
dantavori: I agree, we should change it in content as well (should I commit your suggestion until then?)

demisto_sdk/commands/init/templates/BaseIntegration/README.md
```diff
@@ -0,0 +1,5 @@
+This README contains the full documentation for your integration.
```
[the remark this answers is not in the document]
dantavori: No need - that's the local file which is taken as is. If we take the **remote** file, as part of the process its name is changed from README_example.md to README.md

demisto_sdk/commands/init/initiator.py
```diff
@@ -31,12 +31,67 @@ class Initiator:
...
            script (bool): Indicates whether to create a script.
            full_output_path (str): The full path to the newly created pack/integration/script
     """
 
-    def __init__(self, output: str, name: str = '', id: str = '', integration: bool = False, script: bool = False,
-                 pack: bool = False, demisto_mock: bool = False, common_server: bool = False):
+    TEST_DATA_DIR = 'test_data'
+
+    ''' INTEGRATION TEMPLATES CONSTANTS '''
+    DEFAULT_INTEGRATION_TEMPLATE = 'BaseIntegration'
```
anara123, line 42: what is the difference between BaseIntegration and HelloWorld ? We better think more user friendly options

demisto_sdk/commands/init/templates/BaseIntegration/BaseIntegration.yml
```diff
@@ -0,0 +1,52 @@
...
+- display: API Key
+  additionalinfo: The API Key to use for connection
+  name: apikey
+  required: true
+  type: 4
+- display: Trust any certificate (not secure)
+  name: insecure
+  required: false
+  type: 8
+- display: Use system proxy settings
```
anara123, line 20: I don't like the fact that we copy the files from content repo, because then we will not get updates on these templates.
Maybe as part of the demisto-sdk version release, these templates will be downloaded and added to the project. This way each release demitso-sdk will contain the updated version of the templates.

dantavori: These files are only backups in case the files extraction from content fails

demisto_sdk/commands/init/initiator.py
dantavori: HeloWorld is the template that was so far, BaseIntegration is the new one made by @fvigo. I guess this was discussed before, but anyway if we decide to change their names this should happen in content repo first and we can open a separate PR for it here later

roysagi: @anara123 - This is a more slimmed-down initial integration since HelloWorld is robust and easy to lose yourself in.
This one creates an easier and cleaner starting point and consulting with @fvigo it seemed like something that would benefit contributors more.

demisto_sdk/commands/init/templates/BaseIntegration/BaseIntegration.yml
roysagi: @anara123 - that is a good idea though to update the templates we should open an issue about it especially since the SDK release automation is in progress - https://github.com/demisto/etc/issues/22907
For now keeping the integration on hand as a fallback does seem like a good idea to me.

PR #983, opened 2020-12-30
SDK repo mypy handle check_untyped_defs errors

bakatzir:
## Status
- [x] Ready

## Desc
I've tended to most of the check_untyped_defs errors and commented out it in the mypy.ini
i've ignored a few errors, mainly regarding to files that are working, but that I felt it it would be harder effort to change.
## Related Issues
fixes: https://github.com/demisto/etc/issues/26633


demisto_sdk/commands/generate_test_playbook/test_playbook_generator.py
```diff
@@ -159,13 +159,13 @@ def create_automation_task(_id, automation_name, item_type: ContentItemType, arg
     }
 
 
-def create_verify_outputs_task(id, conditions=[]):
+def create_verify_outputs_task(id_, conditions=[]):
```
amshamah419, line 162: Why the arg name change?

amshamah419 (approved):
Don't forget to add the change to the changelog.

demisto_sdk/commands/generate_test_playbook/test_playbook_generator.py
bakatzir: id-is-a-bad-variable-name-in-python

https://stackoverflow.com/questions/77552/id-is-a-bad-variable-name-in-python#:~:text=It's%20bad%20to%20name%20any,know%20the%20name%20is%20overridden.&text=id%20is%20a%20built%2Din%20function%20in%20Python.,id%20will%20override%20the%20function.

PR #986, opened 2020-12-30
fixing issue where id_set creation was created even when skipping flag was set

roysagi:
## Description
fixing issue where id_set creation was created even when skipping flag was set


Itay4:
can we add test?

roysagi:
> can we add test?

yup just did

Itay4 (approved):

yaakovi (approved):

PR #988, opened 2020-12-30
Add legal duplicates in merge-id-sets command

ChanochShayner:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
Add legal duplicates in the merge-id-sets command.

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


roysagi:
@ChanochShayner - please add a changelog entry

demisto_sdk/commands/create_id_set/tests/update_id_set_test.py
```diff
@@ -248,15 +248,17 @@ def test_has_duplicate():
...
         })
 
         id_set['Layouts'].append({
             'urlRep': {
                 'typeID': 'urlRep',
                 'kind': 'Details',
-                'path': 'Layouts/layout-details-urlrep2.json'
+                'path': 'Layouts/layout-details-urlrep2.json',
+                'pack': 'urlRep2'
             }
```
ChanochShayner, line 262: @anara123 without the difference in the pack names, 
Should this be considered a duplicate?
Until now it was considered that way, and now it depends on the name of the pack.

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -1451,6 +1451,9 @@ def has_duplicate(id_set_subset_list, id_to_check, object_type=None, print_logs=
             if dict1.get('kind', '') != dict2.get('kind', ''):
                 return False
 
+        if dict1.get('pack') == dict2.get('pack'):
```
roysagi, line 1454: didn't we say in this case it shouldn't get to the `united_id_set`? but in this case it does.
Also please add a dev comment here explaining this.

ChanochShayner: It is not considered as duplicates, and it is not added to united_id_set - 
Because I added a test that makes sure there is no such entry already on the list.
The reason it should not be added to the united_id_set is that such an entity already exists there.

anara123: I think you better change the condition to be 
```
if dict1.get('pack') == dict2.get('pack'):
    return True
```
This way the for will run over all the duplication combinations and if there are two non identical packs, it will return True

anara123 (approved):
see my comment

PR #996, opened 2021-01-03
disable pylint unsubscriptable-object for py3.9

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: No issue 

## Description
disable pylint unsubscriptable-object for py3.9

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/lint/commands_builder.py
+++ b/demisto_sdk/commands/lint/commands_builder.py
@@ -1,7 +1,7 @@
 # STD python packages
 import os
 from pathlib import Path
-from typing import List
+from typing import List, Optional
 
 from demisto_sdk.commands.lint.resources.pylint_plugins.base_checker import \
     base_msg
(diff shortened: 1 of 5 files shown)


demisto_sdk/commands/lint/commands_builder.py
```diff
@@ -222,7 +222,10 @@ def build_pylint_command(files: List[Path]) -> str:
     # Prints only errors
     command += " -E"
     # disable xsoar linter messages
-    command += " --disable=bad-option-value"
+    disable = ['bad-option-value']
+    if image_name is not None and 'demisto/python3:3.9' in image_name:
```
roysagi, line 226: Please add a TODO message here to indicate this should be removed once - https://github.com/PyCQA/pylint/issues/3882#issuecomment-745148724 - an new pylint version is released - also please open an issue about this.

glicht: This is not a very good way to detect the python version. We may have an image that uses `demisto/python3` as a base and thus running python 3.9.x. Better to use: `get_python_version_from_image`. Note that calling this function multiple time is ok as we use an lru_cache.

https://github.com/demisto/demisto-sdk/blob/master/demisto_sdk/commands/lint/helpers.py#L274

glicht (changes requested):

demisto_sdk/commands/lint/commands_builder.py
GalRabin: We can get the python version from `self._facts["python_version"]`, Its collected from the docker interperter. (https://github.com/demisto/demisto-sdk/blob/1340804dc19ccb909313b15cf0c039dff86bb0c5/demisto_sdk/commands/lint/linter.py#L220)

The docker name or tag doesn't necessary include the python version.

roysagi (approved):

jochman:
@glicht waiting for your approval

glicht (approved):

PR #1000, opened 2021-01-04
Raise nice error if can't connect to server

orlichter1:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32301

## Description
Upload now checks gracefully if it can connect to demisto server

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/upload/uploader.py
+++ b/demisto_sdk/commands/upload/uploader.py
@@ -77,6 +77,11 @@ class Uploader:
     def upload(self):
         """Upload the pack / directory / file to the remote Cortex XSOAR instance.
         """
+        if self.demisto_version == "0":
+            click.secho("Could not connect to XSOAR server. Try checking your connection configurations.",
+                        fg="bright_red")
+            return 1
+
         status_code = 0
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/upload/uploader.py
```diff
@@ -77,6 +77,11 @@ def __init__(self, input: str, insecure: bool = False, verbose: bool = False):
     def upload(self):
         """Upload the pack / directory / file to the remote Cortex XSOAR instance.
         """
+        if self.demisto_version == "0":
+            click.secho("Could not connect to XSOAR server. Try checking your connection configurations.",
```
yaakovi, line 81: if the problem is caused by the connection to the server, wouldn't it raise an exception in an earlier stage (when calling the `configure` function of demisto-py)?

orlichter1: That's what I thought too, but if you have a look at the function, it doesn't actually do any connecting to the server, only configures a class if you use an API

yaakovi (approved):

PR #1002, opened 2021-01-04
Support running with REHL using podman

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/29360

## Description
- Using the internal IP of the instance instead of the default docker bridge's IP.
- Added a method that will query the instance and check if this is a RHEL instance or not and return a boolean accordingly.
- Not running docker threshold tests on RHEL instance


hod-alpert:
Running another nightly after the latest changes here https://app.circleci.com/pipelines/github/demisto/content/58688/workflows/685e2d7c-e18f-4615-871c-e15e30c7caf5

glicht (approved):

PR #1007, opened 2021-01-06
extracting the name property from layoutscontainer data (#1006)

### LINKED ISSUE #1006: Cannot download indicator layoutscontainer objects
## What happened?

1. I added an indicator layout to one of our instances manually via the layout editor (6.0.1, both SDK versions 1.2.11 and current master)
2. I wanted to download the custom content via the SDK in order to add it to my content pack but it wouldn't even list the content
  * `$ demisto-sdk download -lf` would not show the content
3. So obviously I could not download the layout JSON via the SDK

## Why?

After adding some debug statements to the SDK's source code I was able to see that it indeed got the custom content from the instance *but was not listing it because it was not able to derive the content's name from the JSON file*. It tries to look for a property called `typeId` in layout files which does not exist for this particular content I created.

meilke:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
Fixes #1006

### What happened?

1. I added an indicator layout to one of our instances manually via the layout editor (6.0.1, both SDK versions 1.2.11 and current master)
2. I wanted to download the custom content via the SDK in order to add it to my content pack but it wouldn't even list the content
  * `$ demisto-sdk download -lf` would not show the content
3. So obviously I could not download the layout JSON via the SDK

### Why?

After adding some debug statements to the SDK's source code I was able to see that it indeed got the custom content from the instance *but was not listing it because it was not able to derive the content's name from the JSON file*. It tries to look for a property called `typeId` in layout files which does not exist for this particular content I created.

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -451,7 +451,7 @@ def get_entity_name_by_entity_type(data: dict, content_entity: str):
     :return: The file name
     """
     if content_entity == LAYOUTS_DIR:
-        return data.get('typeId', '')
+        return data.get('typeId', data.get('name', ''))
     else:
         return data.get('name', '')
 


meilke:
Is there anything else I can provide?

bakatzir:
@meilke Hey, thanks for the suggestion and sorry for the late response.
Can you add here or DM me in our slack DFIR the custom layout you've created?
This way I will be able to create a test for your addition.

meilke:
@bakatzir I was mainly working with this layout: https://github.com/demisto/content/blob/master/Packs/FeedCyrenThreatInDepth/Layouts/layoutscontainer-Cyren_Threat_InDepth_Indicator_Template.json

bakatzir:
tended in: https://github.com/demisto/demisto-sdk/pull/1106

PR #1012, opened 2021-01-07
Always clean mock file after creation

hod-alpert:
## Description
Since the query data is sorted - we need to always clean the newly created mock files and not only when problematic keys are set.

### DIFF
--- a/demisto_sdk/commands/test_content/mock_server.py
+++ b/demisto_sdk/commands/test_content/mock_server.py
@@ -1,7 +1,6 @@
 from __future__ import print_function
 
 import ast
-import json
 import os
 import string
 import time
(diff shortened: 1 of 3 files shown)


avidan-H (changes requested):
Should also probably adjust the method name to be more accurate for what it does now - `normalize_mock_file` - will need to adjust wherever it is invoked as well. For the docstring, you can add something like the following:
```
        """Normalize the mock file of a test playbook

		Normalizes a mock file by doing the following:
		1. Replaces the timestamp/variable data in a mockfile as identified by
        the keys provided in the mock file's associated 'problematic_keys.json'
        file, with constant values.
		2. Standardizes the query parameter data order for all requests.

        Args:
            playbook_or_integration_id (string): ID of the test playbook or
                integration for which the associated mock file will be
                normalized.
			path (string [Optional]): Path to the directory in which to search
                for the required data (temp directory or repo). If not
                provided, method will use the current directory.
        """
```

Also, need to replace the log message on line 315 with,
```
        self.logging_module.debug(f'normalize_mock_file was called for test "{playbook_or_integration_id}"')
```
if you update the method name.

demisto_sdk/commands/test_content/mock_server.py
```diff
@@ -323,65 +338,44 @@ def clean_mock_file(self, playbook_or_integration_id, path=None):
...
+            self.logging_module.debug(f'diff_cmd_output={diff_cmd_output}')
+            if diff_cmd_output.endswith('are identical'):
+                self.logging_module.debug('normalized mock file and original mock file are identical')
+            else:
+                self.logging_module.debug('the normalized mock file differs from the original')
+
+        except CalledProcessError:
+            self.logging_module.debug('the normalized mock file differs from the original')
+
+        self.logging_module.debug('Replace old mock with cleaned one.')
```
avidan-H, line 376:
```suggestion
        self.logging_module.debug('Replacing original mock file with the normalized one.')
```

avidan-H (approved):

PR #1019, opened 2021-01-10
Add schema change notifier

DinaMeylakh:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


demisto_sdk/utils/server_schema_change_slack_notifier.py
```diff
@@ -0,0 +1,137 @@
...
+        commit_hash: current commit hash
+
+    Return: Tuple of
+        commit_github_url: Url of commit on github.
+        relevant_changed_files: List of relevant files that were changed.
+        change_diffs: Dictionary with changed_files keys and changesets as values.
+        commit_author: Author of commit.
+    """
+    commit_github_url = os.path.join(GITHUB_BASE_URL, commit_hash)
+    output_stream = os.popen("git log -1 --pretty=format:'%an'")
```
roysagi, line 39: might be easier to use gitPython here instead of all the streams - https://gitpython.readthedocs.io/en/stable/tutorial.html

roysagi (approved):
Looks good to me - though you could use gitpython and it might make the git uses more readable and maybe easier.

Also @Itay4 - might have some input on the slack section.

demisto_sdk/utils/server_schema_change_slack_notifier.py
```diff
@@ -0,0 +1,137 @@
...
+        message: slack message in a format like slack_sdk attachment argument expects.
+    """
+
+    title = f"Commit {commit[0:7]} Modified Schema Struct Files"
+    changes = [{"value": f"Changed by: {commit_author}"}]
+    for changed_file in changed_files:
+        changes.append({
+            "title": f"{changed_file}",
+            "value": f"```{diffs[changed_file]}```",
+            "short": False  # type: ignore
```
Itay4, line 93: why we need the ignore here?

demisto_sdk/utils/server_schema_change_slack_notifier.py
DinaMeylakh: I tried but inside the server build I got an error 'git.exc.NoSuchPathError' but with a real path '/builds/xsoar/server'
This is probably somehow connected to the difference docker runs python code and os.popen.
I don't really want to get to the bottom of this since:
1. os.popen functions do work
2. The readability gitpython adds is minimal for my taste
3. Using gitpython adds another library installation in the build/ docker image which I think is unnecessary 
4. It would take up some time

You can checkout the code in the last commit to see if you think its worth the trouble: https://github.com/demisto/demisto-sdk/pull/1019/commits/86992acd911a3c98ecb2e870e3ecdbab3e3a5090

roysagi: I see - then please add a dev note as to what is being done here for readability and maintainability. - same for the other popen git call.

roysagi (approved):

PR #1025, opened 2021-01-12
Supporting unmockable playbooks

hod-alpert:
## Related Issues
fixes: https://github.com/demisto/etc/issues/31386

## Description
Supporting unmockable playbooks

### DIFF
--- a/demisto_sdk/commands/test_content/TestContentClasses.py
+++ b/demisto_sdk/commands/test_content/TestContentClasses.py
@@ -96,6 +96,7 @@ class TestConfiguration:
         self.timeout = test_configuration.get('timeout', default_test_timeout)
         self.memory_threshold = test_configuration.get('memory_threshold', Docker.DEFAULT_CONTAINER_MEMORY_USAGE)
         self.pid_threshold = test_configuration.get('pid_threshold', Docker.DEFAULT_CONTAINER_PIDS_USAGE)
+        self.is_mockable = test_configuration.get('is_mockable')
         self.test_integrations: List[str] = self._parse_integrations_conf(test_configuration)
         self.test_instance_names: List[str] = self._parse_instance_names_conf(test_configuration)
 
@@ -513,7 +514,11 @@ class BuildContext:
                                             integration_name in unmockable_integrations]
             if test_name and (not test_record.test_integrations or unmockable_integrations_used) or not self.isAMI:
                 unmockable_tests.append(test_record)
-
+                # In case a test has both - an unmockable integration and is configured with is_mockable=False -
+                # it will be added twice if we don't continue.
+                continue
+            if test_record.is_mockable is False:
+                unmockable_tests.append(test_record)
         return unmockable_tests
 
     @staticmethod
--- a/demisto_sdk/commands/test_content/tests/BuildContext_test.py
+++ b/demisto_sdk/commands/test_content/tests/BuildContext_test.py
@@ -13,7 +13,8 @@ def generate_test_configuration(playbook_id: str,
                                 toversion: str = '',
                                 timeout: int = None,
                                 memory_threshold: int = None,
-                                pid_threshold: int = None
+                                pid_threshold: int = None,
+                                is_mockable: bool = None
                                 ) -> dict:
     playbook_config = {
         'playbookID': playbook_id,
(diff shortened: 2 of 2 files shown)


demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -513,7 +514,9 @@ def _get_unmockable_tests_from_conf(self) -> list:
                                             integration_name in unmockable_integrations]
             if test_name and (not test_record.test_integrations or unmockable_integrations_used) or not self.isAMI:
                 unmockable_tests.append(test_record)
-
+                continue
+            if test_record.is_mockable is False:
```
Itay4, line 518:
```suggestion
            if not test_record.is_mockable:
```

demisto_sdk/commands/test_content/tests/BuildContext_test.py
```diff
@@ -13,7 +13,8 @@ def generate_test_configuration(playbook_id: str,
                                 toversion: str = '',
                                 timeout: int = None,
                                 memory_threshold: int = None,
-                                pid_threshold: int = None
+                                pid_threshold: int = None,
+                                is_mockable: bool = None
```
Itay4, line 17:
```suggestion
                                is_mockable: bool = None,
```

demisto_sdk/commands/test_content/tests/BuildContext_test.py
```diff
@@ -13,7 +13,8 @@ def generate_test_configuration(playbook_id: str,
                                 toversion: str = '',
                                 timeout: int = None,
                                 memory_threshold: int = None,
-                                pid_threshold: int = None
```
Itay4, line 16: please add docstring to this func and for all args with default value of `None` set the type int to be `Optional[...]`

demisto_sdk/commands/test_content/tests/BuildContext_test.py
```diff
@@ -34,6 +35,8 @@ def generate_test_configuration(playbook_id: str,
         playbook_config['memory_threshold'] = memory_threshold
     if pid_threshold:
         playbook_config['pid_threshold'] = pid_threshold
+    if is_mockable is not None:
```
Itay4, line 38:
```suggestion
    if is_mockable:
```

demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -513,7 +514,9 @@ def _get_unmockable_tests_from_conf(self) -> list:
                                             integration_name in unmockable_integrations]
             if test_name and (not test_record.test_integrations or unmockable_integrations_used) or not self.isAMI:
                 unmockable_tests.append(test_record)
-
+                continue
```
Itay4, line 517: why do we continue here?

demisto_sdk/commands/test_content/tests/BuildContext_test.py
```diff
@@ -295,3 +298,23 @@ def test_playbook_with_version_mismatch_is_skipped(mocker, tmp_path):
                                              content_conf_json=content_conf_json,
                                              filtered_tests_content=filtered_tests)
     assert 'playbook_with_version_mismatch' in build_context.tests_data_keeper.skipped_tests
+
+
+def test_unmockable_playbook_configuration(mocker, tmp_path):
```
Itay4, line 303: do we have test for playbook that is not set as `is_mockable` (the default scenario) and see it does not come up in the `unmockable_test_ids` ?

demisto_sdk/commands/test_content/TestContentClasses.py
hod-alpert: @Itay4 I don't want this to affect the tests unless specifically given.
When not specifically given the value will be `None` and with your suggestion it will make all tests unmockable.

demisto_sdk/commands/test_content/TestContentClasses.py
hod-alpert: In case a test has both - an unmockable integration and is configured with is_mockable=False - it will be added twice if we don't continue.

demisto_sdk/commands/test_content/tests/BuildContext_test.py
hod-alpert: I'll add one.

demisto_sdk/commands/test_content/TestContentClasses.py
Itay4: please add a comment for that here to explain that

Itay4 (approved):
just note my comment

PR #1034, opened 2021-01-14
Find dependencies optional cases

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32590

## Description
* Playbook dependencies from incident and indicator fields are now marked as optional.
* Mappers dependencies from incident types and incident fields are now marked as optional.
* Classifier dependencies from incident types are now marked as optional.

added more test cases to reflect that

## Must have
- [x] Tests
- [x] Documentation


yuvalbenshalom (changes requested):
looks good, but need to handle the edge case that is common one:
if dependency is for common types, then it should be still regarded as mandatory because it is most probably not the case of a type that you may use.

demisto_sdk/commands/find_dependencies/tests/find_dependencies_test.py
```diff
@@ -678,6 +689,89 @@ def test_collect_playbooks_dependencies_on_incident_fields(self, id_set):
 
         assert IsEqualFunctions.is_sets_equal(found_result, expected_result)
 
+    def test_collect_playbooks_dependencies_on_incident_fields__phishing_pack(self, id_set):
+        """
+        Given
+            - A playbook entry in the id_set.
```
anara123, line 695: add more description here. because the reason for phishing to be optional is because `Search And Delete Emails - EWS` has a playbook or classifier that use a field from Phishing

anara123 (approved):

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -439,19 +440,35 @@ def _collect_playbooks_dependencies(pack_playbooks: list, id_set: dict, verbose_
...
+            # playbook dependencies from incident fields should be marked as optional unless CommonTypes pack,
+            # as customers do not have to use the OOTB inputs.
             incident_fields = playbook_data.get('incident_fields', [])
             packs_found_from_incident_fields = PackDependencies._search_packs_by_items_names_or_ids(
                 incident_fields, id_set['IncidentFields'], exclude_ignored_dependencies)
             if packs_found_from_incident_fields:
-                pack_dependencies_data = PackDependencies._label_as_mandatory(packs_found_from_incident_fields)
+                common_types_pack_dependency = False
+                if COMMON_TYPES_PACK in packs_found_from_incident_fields:
+                    common_types_pack_dependency = True
```
yaakovi, line 451: why do you need the flag? just update the mandatory dependency here.

also, this block causes lots of code duplication, please move to a function

yaakovi (commented):
LGTM

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -284,6 +285,27 @@ def _label_as_optional(pack_ids: set) -> list:
...
+
+        """
+        common_types_pack_dependency = False
+        if COMMON_TYPES_PACK in packs_found_from:
+            packs_found_from.remove(COMMON_TYPES_PACK)
+            common_types_pack_dependency = True
+        pack_dependencies_data = PackDependencies._label_as_optional(packs_found_from)
+        if common_types_pack_dependency:
+            pack_dependencies_data.extend(PackDependencies._label_as_mandatory({COMMON_TYPES_PACK}))
+        return pack_dependencies_data
```
anara123, line 307: It is not clear what this function does. update the description and the param name/description `packs_found_from`
And why do we exclude CommonTypes?

bakatzir: Say that I release a pack, Zimperium, which fetches incidents and has a classifier and a mapper.
the mapper maps to an incident field called `deviceid` which i've added to the `CommonTypes` pack.
only a mandatory dependency will coerce the customers to update their `CommonTypes` pack and will allow a smooth fetch process.

I'll update the desc and var names.

anara123: OK, so you want the customer constantly to update their CommonTypes pack. Got it

yuvalbenshalom (approved):

PR #1035, opened 2021-01-14
use lru in get_remote_file

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32717

## Description
as the approved files are rather small:
```
-rw-r--r--  1 bkatzir  PALOALTONETWORK\Domain Users   658B Jan  6 15:52 Tests/Marketplace/approved_tags.json
-rw-r--r--  1 bkatzir  PALOALTONETWORK\Domain Users   776B Jan 11 17:29 Tests/Marketplace/approved_usecases.json
```
I've put the maxsize on 8.

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -1,5 +1,4 @@
 import argparse
-import functools
 import glob
 import io
 import json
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -161,6 +160,7 @@ def run_command(command, is_silenced=True, exit_on_error=True, cwd=None):
     return output
 
 
+@lru_cache(maxsize=8)
```
glicht, line 163: I think you can increase this. 8 is quite small. You can put in 50.

glicht (approved):

demisto_sdk/commands/common/tools.py
bakatzir: they recommended powers of 2. i'll do 64.

bakatzir: they recommended powers of 2. i'll do 64.

PR #1044, opened 2021-01-18
Improved the deprecated message validation required from integrations

bakatzir:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32712

## Description
i've added tests to another validation which was not tested

## Must have
- [x] Tests


demisto_sdk/commands/common/errors.py
```diff
@@ -464,7 +464,9 @@ def invalid_deprecated_integration_display_name():
     @staticmethod
     @error_code_decorator
     def invalid_deprecated_integration_description():
-        return 'The description of all deprecated integrations should start with "Deprecated.".'
+        return 'The description of all deprecated integrations should start with "Deprecated."' \
+               ' and contain both the phrases: "Use the", "integration instead"' \
+               ' or the phrase: "No available replacement".'
```
yuvalbenshalom, line 469: I am not sure if we want to permit this level of freedom. maybe we should print the 3-4 strict examples from the issue in the message and enforce all of dose possibilities with a strict regex?

yuvalbenshalom (commented):
lets decide if we want a strict form of message or a loos one.
@glicht WDYT?
see my comment.

demisto_sdk/commands/common/errors.py
bakatzir: I implemented it more loosely on purpose, I see no need for the stricter version, but your call.

yaakovi: I'm with Yuval on this one. we should use a strict format

yuvalbenshalom (approved):

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -103,13 +104,40 @@ def is_valid_file(self, validate_rn: bool = True, skip_test_conf: bool = False)
...
+                error_message, error_code = Errors.invalid_deprecated_integration_display_name()
+                if self.handle_error(error_message, error_code, file_path=self.file_path):
+                    is_valid = False
+        return is_valid
+
+    def _is_valid_deprecated_integration_description(self) -> bool:
+        is_valid = True
+        is_deprecated = self.current_file.get('deprecated', False)
+        description = self.current_file.get('description', '')
+        deprecated_v2_regex = r'Deprecated\. Use the .+ integration instead\.'
```
glicht, line 128: Small remark. I would build the code so it iterates via a list of regexes instead of if/elif/else. That way it is easier in the future to add additional regex patterns as we see fit.

glicht (approved):

PR #1046, opened 2021-01-18
Reset containers before mockable tests as well

hod-alpert:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
When the demisto server initializes it starts the default python2 container without any proxy configurations (because they are not configured on the server yet)

Therefore, when running unmockable tests playbooks with integrations that uses that container (such as Mimecast v2, BigFix etc.) some of the commands can go to the old non-configured container and the requests will be routed to the integration instance itself instead of using the proxy playback.

The solution is to reset all containers before starting to run mockable tests and reset it again when starting to run unmockable tests.

### DIFF
--- a/demisto_sdk/commands/test_content/TestContentClasses.py
+++ b/demisto_sdk/commands/test_content/TestContentClasses.py
@@ -1579,7 +1579,6 @@ class ServerContext:
         Args:
             queue: The queue to fetch tests to execute from
         """
-
         self._reset_containers()
         while queue.unfinished_tasks:
             try:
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -1570,10 +1570,17 @@ def _execute_unmockable_tests(self):
...
+        Iterates the tests queue and executes them as long as there are tests to execute.
+        Before the tests execution starts we will reset the containers to make sure the proxy configuration is correct
+        - We need it before the mockable tests because the server starts the python2 default container when it starts
+            and it has no proxy configurations.
+        - We need it before the unmockable tests because at that point all containers will have the proxy configured
+            and we want to clean those configurations when testing unmockable playbooks
         Args:
             queue: The queue to fetch tests to execute from
         """
+
```
avidan-H, line 1582:
```suggestion
```

avidan-H (approved):

PR #1047, opened 2021-01-18
Using the *verbose* argument in the **find-dependencies** command wil…

bakatzir:
…l now log to the console

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32844
relates to: https://github.com/demisto/etc/issues/23929

## Description
usage of verbose is now to log to the console like in the other commands in the repo.


yaakovi (commented):
as a file, the user could focus on the requested pack (top) and ignore the rest of it easily.
when printed to the console, it will create massive spam of data related to the dependencies of other packs.

I think it is better to print only the first level dependencies here and not the whole tree.
don't think anyone will use it but we can enable verbose levels as we have in the lint command:
* 0 - no additional logs
* 1 - only first level dependency prints
* 2 - all level dependency prints

bakatzir:
@yaakovi I agree that is a bit redundant, I've made the changes as it was a loweffort, LMK WDYT.

yaakovi (approved):

bakatzir:
Will force merge over the coverage when all other tests will pass.

PR #1051, opened 2021-01-19
adding detached to layoutcontainer schema

roysagi:
Adding detached to layoutcontainer schema - https://github.com/demisto/server/commit/2675e451ede1c620dd673dc8c9c1991c4ea30175

### DIFF
--- a/demisto_sdk/commands/common/schemas/layoutscontainer.yml
+++ b/demisto_sdk/commands/common/schemas/layoutscontainer.yml
@@ -22,6 +22,8 @@ mapping:
     type: str
   system:
     type: bool
+  detached:
+    type: bool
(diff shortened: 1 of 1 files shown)


Itay4:
@roysagi not sure i understand why we want this in our schema, would we release with this set to true ?

roysagi:
@Itay4 - no but I did see `system` in there which is effectively the same:
image

The questions here are:
1) Do we want our devs and contrib's validation to fail if this field is present? - I suspect not
2) Do want to enforce this to be False? - I can add a quick validation for this.

what do you think?

Itay4:
in my opinion we should just let format remove it like we have for many other fields

PR #1060, opened 2021-01-21
When Context Example is empty don't generate the section in generate-docs

ohaim1008:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/32937

## Description
We didn't consider an string that contain an empty dict = '{}' as empty output, therefor a section was created.

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
+++ b/demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
@@ -392,7 +392,7 @@ def test_generate_commands_section():
 
 def test_generate_command_section_with_empty_cotext_example():
     """
-    When an string represent an emmpty dict '{}' is the context output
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/generate_docs/tests/generate_docs_test.py
```diff
@@ -390,6 +390,27 @@ def test_generate_commands_section():
     assert '\n'.join(section) == '\n'.join(expected_section)
 
 
+def test_generate_command_section_with_empty_cotext_example():
+    """
+    When an string represent an emmpty dict '{}' is the context output
```
bakatzir, line 395:
```suggestion
    When an string representד an empty dict '{}' is the context output
```

bakatzir (approved):

PR #1066, opened 2021-01-25
Performance improvement for all files validations

hod-alpert:
Recently the file's validation duration has increased significanly in the build.

Working with @bakatzir we profiled it and saw that at least 20% of the time it spends on opening the id_set file.

So an easy optimization is to open the id_set.json once and cache it for reuse.
This alone have decreased the duration in ~8 minutes.

In addition, instead of creating the content id_set, downloading it from the GCP bucket can also save additional ~2 minutes. 

All together - time reduction of ~9 minutes.

### DIFF
--- a/demisto_sdk/commands/validate/validate_manager.py
+++ b/demisto_sdk/commands/validate/validate_manager.py
@@ -79,7 +79,6 @@ class ValidateManager:
         self.no_configuration_prints = silence_init_prints
         self.skip_conf_json = skip_conf_json
         self.is_backward_check = is_backward_check
-        self.id_set_validations = validate_id_set
         self.is_circle = only_committed_files
         self.validate_all = validate_all
         self.use_git = use_git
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -98,6 +97,10 @@ def __init__(
         if not id_set_path:
             id_set_path = 'Tests/id_set.json'
         self.id_set_path = id_set_path
+        # create the id_set only once per run.
+        self.id_set_validator = IDSetValidator(is_circle=self.is_circle, configuration=Configuration(),
```
bakatzir, line 101: we have the `self.id_set_path` property in the validate_manager, i suggest using it when instantiating `IDSetValidator`, seems like the "./Tests/id_set.json" is hardcoded too many times.

hod-alpert: Not sure if I totally understand what needs to be done there, is the new commit what you meant?

bakatzir (approved):

demisto_sdk/commands/validate/validate_manager.py
bakatzir: indeed. thx.

PR #1072, opened 2021-01-26
Prevent Content-Bot from repeating same comment in PR

darkushin:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33090

## Description
Before posting a comment into a PR, check if this comment was already posted. If it was, don't post it again to avoid spamming PRs.

### DIFF
--- a/demisto_sdk/commands/test_content/execute_test_content.py
+++ b/demisto_sdk/commands/test_content/execute_test_content.py
@@ -67,7 +67,7 @@ def execute_test_content(**kwargs):
     if not build_context.unmockable_tests_to_run.empty() or not build_context.mockable_tests_to_run.empty():
         raise Exception('Not all tests have been executed')
     if build_context.tests_data_keeper.playbook_skipped_integration and build_context.build_name != 'master':
-        comment = SKIPPED_INTEGRATION_COMMENT + ':\n {}'. \
+        comment = SKIPPED_INTEGRATION_COMMENT + ':\n- {}'. \
             format('\n- '.join(build_context.tests_data_keeper.playbook_skipped_integration))
         _add_pr_comment(comment, logging_manager)
     build_context.tests_data_keeper.print_test_summary(build_context.isAMI, logging_manager)
--- a/demisto_sdk/commands/test_content/execute_test_content.py
+++ b/demisto_sdk/commands/test_content/execute_test_content.py
@@ -31,6 +31,12 @@ def _add_pr_comment(comment, logging_module):
         if res and res.get('total_count', 0) == 1:
             issue_url = res['items'][0].get('comments_url') if res.get('items', []) else None
             if issue_url:
+                # Check if the comment already exists. If it does, don't add another comment to avoid spamming PRs:
+                response = requests.get(issue_url, headers=headers, verify=False)
+                issue_comments = _handle_github_response(response, logging_module)
+                for existing_comment in issue_comments:
+                    if comment == existing_comment.get('body'):
+                        return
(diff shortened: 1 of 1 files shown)


darkushin:
@ShahafBenYakir @Itay4 I changed the flow so that Content-Bot will post a comment only if it doesn't already exist in the issue.
[This PR](https://github.com/demisto/content/pull/11097) shows that the comment is repeated multiple times before the change, and [this PR](https://github.com/demisto/content/pull/11098) shows that it appears only once after adding the changes.

Do you see a reason to update the comment with the latest one and not simply ignoring new ones as they are identical to the existing comment? (is the time of the comment important? in which case we should update the comment and not ignore it)

ShahafBenYakir:
The time is less important but it could be the case where I edit an integration and then I edit another one. If both had skipped tests then only the first one would be mentioned. I think we would want both.

demisto_sdk/commands/test_content/execute_test_content.py
```diff
@@ -8,6 +8,8 @@
 from demisto_sdk.commands.test_content.TestContentClasses import (
     BuildContext, ServerContext)
 
+SKIPPED_INTEGRATION_COMMENT = 'The following integrations are skipped and critical for the test'
```
ShahafBenYakir, line 11: Is this message only for skipped integrtions?

darkushin: Yes, in line 72 we check for all skipped integrations, and only in this case we use this comment.

ShahafBenYakir: I mean, we don't check for tests?

ShahafBenYakir: What would you say the effort is to put it here as well? And how ugly do you think it will make our code?

darkushin: shouldn't be a big effort and shouldn't make the code too ugly. I believe that a simple check as done for the skipped integrations can be done for the skipped tests. I will add it and then we can see if we want to keep this change.

ShahafBenYakir: Feels to me that it might take some more effort, If you think it is better to open a dev task for this then let's.

ShahafBenYakir: Feels to me that it might take some more effort, If you think it is better to open a dev task for this then let's.

demisto_sdk/commands/test_content/execute_test_content.py
```diff
@@ -57,7 +67,7 @@ def execute_test_content(**kwargs):
     if not build_context.unmockable_tests_to_run.empty() or not build_context.mockable_tests_to_run.empty():
         raise Exception('Not all tests have been executed')
     if build_context.tests_data_keeper.playbook_skipped_integration and build_context.build_name != 'master':
-        comment = 'The following integrations are skipped and critical for the test:\n {}'. \
+        comment = SKIPPED_INTEGRATION_COMMENT + ':\n- {}'. \
             format('\n- '.join(build_context.tests_data_keeper.playbook_skipped_integration))
```
ShahafBenYakir, line 71:
```suggestion
        comment = f'{SKIPPED_INTEGRATION_COMMENT}:\n- {'\n-'.join(build_context.tests_data_keeper.playbook_skipped_integration)}'. 
```

ShahafBenYakir (approved):

PR #1073, opened 2021-01-26
Added support for the 'cve' reputation command in default arg validat…

bakatzir:
…ion, filter it in deps

## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/33117

## Must have
- [x] Tests

### DIFF
--- a/demisto_sdk/commands/common/tests/integration_test.py
+++ b/demisto_sdk/commands/common/tests/integration_test.py
@@ -239,7 +239,7 @@ class TestIntegrationValidator:
         assert validator.is_valid_subtype() is answer
 
     DEFUALT_ARGS_1 = [
-        {"name": "cve", "arguments": [{"name": "cve", "required": False, "default": True}]}]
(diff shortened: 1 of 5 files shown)


demisto_sdk/commands/common/tests/integration_test.py
```diff
@@ -238,13 +238,16 @@ def test_id_valid_subtype(self, current, answer):
         validator = IntegrationValidator(structure)
         assert validator.is_valid_subtype() is answer
 
+    DEFUALT_ARGS_1 = [
```
yaakovi, line 241: we had `DEFUALT_ARGS_2` but not `DEFUALT_ARGS_1`? 🤦

yaakovi (approved):

demisto_sdk/commands/common/tests/integration_test.py
bakatzir: 🤯

PR #1075, opened 2021-01-27
id_set dup logic rmed

bakatzir:
## Status
- [x] Ready

relates to: https://github.com/demisto/etc/issues/32746

rmed redundant id_set creation and better the code flow.
validate_files is down from 18 to 15 min now.

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/id.py
+++ b/demisto_sdk/commands/common/hook_validations/id.py
@@ -1,4 +1,3 @@
-import json
 import os
 import re
 from collections import OrderedDict
@@ -17,8 +16,8 @@ from demisto_sdk.commands.common.update_id_set import (get_classifier_data,
 from demisto_sdk.commands.unify.unifier import Unifier
 
 
-class IDSetValidator(BaseValidator):
-    """IDSetValidator was designed to make sure we create the id_set.json in the correct way so we can use it later on.
+class IDSetValidations(BaseValidator):
+    """IDSetValidations was designed to make sure all the inter connected content entities are valid.
 
     The id_set.json file is created using the update_id_set.py script. It contains all the data from the various
     executables we have in Content repository - Playbooks/Scripts/Integration. The script extracts the command and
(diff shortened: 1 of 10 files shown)


demisto_sdk/commands/common/hook_validations/id.py
```diff
@@ -41,41 +40,22 @@ class IDSetValidator(BaseValidator):
...
-                    error_message, error_code = Errors.id_set_conflicts()
-                    if self.handle_error(error_message, error_code, file_path="id_set.json"):
-                        raise
-                    else:
-                        pass
-
-                raise
-
-            return id_set
+            self.id_set_file = id_set_file
```
roysagi, line 50: you changed the name here from `id_set` to `id_set_file` - yet you did not change this anywhere else so is this redundant?

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -72,7 +72,7 @@ def __init__(
             print_ignored_files=False, skip_conf_json=True, validate_id_set=False, file_path=None,
             validate_all=False, is_external_repo=False, skip_pack_rn_validation=False, print_ignored_errors=False,
             silence_init_prints=False, no_docker_checks=False, skip_dependencies=False, id_set_path=None, staged=False,
-            skip_id_set_creation=False
+            skip_id_set_creation=True
```
roysagi, line 75: This is a logic change - why do we do this?
if id_set is not created pack dependencies is not validated which is something you do want locally if possible as far as I know.

Also if we do keep it we need to update `__main__` as the flag description should change to `--create-id-set` as the default now is not to do this.

roysagi (changes requested):

demisto_sdk/commands/validate/validate_manager.py
bakatzir: By default we should not create it, but use the id_set we get from the buvket. we can always override and force creation with the given arg.

demisto_sdk/commands/common/hook_validations/id.py
bakatzir: I changed it from `id_set_path` to `id_set_file`. which I have changed all its occurrences.
Or maybe I'm misunderstanding you?

roysagi (approved):
Looks good to me - Please add a changelog about the flag and behavior change.

PR #1076, opened 2021-01-27
Adding option to ignore id set errors

roysagi:
Adding option to ignore id set errors


bakatzir (approved):

bakatzir:
relates to: https://github.com/demisto/etc/issues/32746

PR #1090, opened 2021-02-01
Fix cpu_count in unittests

hod-alpert:
Fixing the cpu_count method to return 2.

Since Circle build has an issue in it's virtualization where it has only 2 VCPU's but the 'cpu_count' method returns all physical CPU's (36) which causes it to use too many processes in the process pools.

fixes: https://github.com/demisto/etc/issues/33244

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,11 +1,11 @@
 """Configuring tests for the content suite
 """
-import pytest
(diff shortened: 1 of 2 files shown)


yaakovi (approved):
add test 😜

PR #1101, opened 2021-02-07
Fixed process integration function

ChanochShayner:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
create ID set command did not add this integration to ID set -
image (43)

Because the function did not correctly create its Path - 
image (44)

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/common/update_id_set.py
+++ b/demisto_sdk/commands/common/update_id_set.py
@@ -869,8 +869,8 @@ def process_integration(file_path: str, print_logs: bool) -> list:
                 res.append(get_integration_data(file_path))
         else:
             # package integration
-            package_name = os.path.basename(file_path)
-            file_path = os.path.join(file_path, '{}.yml'.format(package_name))
+            file_path = glob.glob(os.path.join(file_path, '*yml'))
+            file_path = str(file_path[0]) if file_path else ''
             if os.path.isfile(file_path):
                 # locally, might have leftover dirs without committed files
                 if print_logs:


demisto_sdk/commands/common/update_id_set.py
```diff
@@ -869,8 +869,8 @@ def process_integration(file_path: str, print_logs: bool) -> list:
                 res.append(get_integration_data(file_path))
         else:
             # package integration
-            package_name = os.path.basename(file_path)
-            file_path = os.path.join(file_path, '{}.yml'.format(package_name))
+            file_path = glob.glob(os.path.join(file_path, '*yml'))
```
yaakovi, line 872: better to fix the problematic content and not change this code.
we already have this assumption in many places.

using glob can cause problems when used locally:
* what if you have a unified file?
* what if you have leftovers?

roysagi: but the original code was incorrect as well as we do support `integration-name.yml` as a valid format that was not caught.
If we are to fix content in such a way it might be better to solve this issue: https://github.com/demisto/etc/issues/29240
in the meantime, I think we should open an issue to solve this and currently fix this in such a way that let's us pass the build correctly (as you said the main issue with this is more than likely to be the local runs rather than the build runs).

One minor improvement we can implement right now is to go through the list and take out anything that ends with `_unified.yml` solving at least part of the problem.

@yaakovi @ChanochShayner - What do you think?

ChanochShayner:
@yaakovi @roysagi You can see that after correcting the path of this integration it exists in the ID set.
Closes this PR.
https://github.com/demisto/content/pull/11321
image

PR #1105, opened 2021-02-08
Replacing the server placeholder with the server's internal IP

hod-alpert:
Use the internal server_ip for setting integration params.

Specifically - for integrations that needs to be configured with the server's IP in run-time.

@DeanArbel @teizenman Can you add the test failure issue here?

Here's a build that should run the TAXII playbook: https://app.circleci.com/pipelines/github/demisto/content/66657/workflows/a02be122-10a6-480a-a007-418bd8adf466/jobs/275217

fixes: https://github.com/demisto/etc/issues/32567

But it will be fixed only on the next sdk release.


teizenman (approved):

DeanArbel (approved):
Impressive.

PR #1110, opened 2021-02-09
Added condition for creating id_set

Shellyber:
## Status
- [ ] In Progress
- [x] Ready


demisto_sdk/commands/generate_docs/generate_script_doc.py
```diff
@@ -43,8 +45,12 @@ def generate_script_doc(input_path, examples, output: str = None, permissions: s
...
 
         # get the script usages by the id set
-        id_set_creator = IDSetCreator(output='', print_logs=False)
-        id_set = id_set_creator.create_id_set()
+        if not os.path.isfile(DEFAULT_ID_SET_PATH):
+            id_set_creator = IDSetCreator(output='', print_logs=False)
+            id_set = id_set_creator.create_id_set()
+        else:
+            id_set = open_id_set_file(DEFAULT_ID_SET_PATH)
+
```
bakatzir, line 53: It takes some time to create the id set, so maybe it is better to skip the `used_in` section in the script readme if non existent.
@roysagi WDYT?

roysagi: I think the best course of action here is to make it interactable - ask the user if he would like us to create the id_set - if no just skip it and print a warning about what would be missing.

Shellyber: @roysagi I can't, I'm doing this cause it takes time to create the id_set and I want to reduce it for the simple contribution flow. This runs automatically while generating the script doc

Shellyber:
At the moment we do not support empty id_set.. 
Need to think if we want to support this.
@roysagi

roysagi:
We should support generating docs without an available id set - we should open an issue about that.
That means the current PR's feature is good as is without any need for additional changes - @bakatzir FYI

roysagi (approved):
Please add a changelog

PR #1111, opened 2021-02-09
allowing indicator field upload

meilke:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
N/A

## Description
Up until now the SDK does not allow to upload indicator fields from the indicator fields folder of a pack. One has to wrap the individual JSON files into a temporary JSON array file and then upload it via the UI (Settings -> Advanced -> Fields -> Upload Icon). The API endpoint used in this operation is `/incidentfields/import` (which is wrapped in the Python client as `client.import_incident_fields()`).

The fix does something similar here: it re-uses the upload logic from the incident field pack object class. I decided against subclassing the incident field pack object class because I feel indicator fields deserve either a dedicated endpoint for uploads or the incident upload endpoint is generalized. So subclassing would create an unnecessary dependency between the two.

## Screenshots
N/A

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/common/content/objects/pack_objects/indicator_field/indicator_field.py
+++ b/demisto_sdk/commands/common/content/objects/pack_objects/indicator_field/indicator_field.py
@@ -1,5 +1,8 @@
 from typing import Union
+from tempfile import NamedTemporaryFile
+import json
 
+import demisto_client
 from demisto_sdk.commands.common.constants import (INCIDENT_FIELD,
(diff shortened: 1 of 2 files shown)


JakubOrzol:
Nice one it would be very useful!

ShahafBenYakir:
Very nice thanks @meilke

yaakovi (approved):

PR #1116, opened 2021-02-10
adding prev-ver support on pack_metadata file check

roysagi:
adding prev-ver support on pack_metadata file check


bakatzir (approved):
Just add the prev_ver to the docstring of the init to explain its purpose.

PR #1125, opened 2021-02-16
introducing new error for mdx server installation failure

roysagi:
If MDX server fails we should fail the validation but not crash the command.
Also improved the error message with a suggestion about what should be done to correct this.

### DIFF
--- a/demisto_sdk/commands/common/errors.py
+++ b/demisto_sdk/commands/common/errors.py
@@ -183,6 +183,7 @@ ERROR_CODE = {
     "readme_error": {'code': "RM100", 'ui_applicable': False, 'related_field': ''},
     "image_path_error": {'code': "RM101", 'ui_applicable': False, 'related_field': ''},
     "readme_missing_output_context": {'code': "RM102", 'ui_applicable': False, 'related_field': ''},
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/errors.py
```diff
@@ -482,6 +483,12 @@ def readme_missing_output_context(command, context_paths):
         return f'The Following context paths for command {command} are found in YML file ' \
                f'but are missing from the README file: {context_paths}'
 
+    @staticmethod
+    @error_code_decorator
+    def error_starting_mdx_server(line, packs):
+        return f'Failed starting mdx server. stdout: {line}.\n' \
+               f'Try running the following command: `npm install {" ".join(packs)}`'
```
glicht, line 490: The command to run is simply: `npm install`. There is no need to specify the packages as they are specified in `package.json`.
```suggestion
               f'Try running the following command: npm install'
```

glicht (changes requested):

demisto_sdk/commands/common/errors.py
roysagi: done

glicht (approved):

PR #1145, opened 2021-03-02
Added support in docker threshold test for powershell integrations

hod-alpert:
Added support in powershell docker threshold test.

Up until now - the docker images that were collected weren't the right ones.

Here's an example of how it looks (and fails) with that new fix.
image


glicht (approved):
Looks good. 

We need to raise the default thresholds for powershell integrations as they seem different than python. Let's set pid on: 24 and memory on 140MB.

PR #1150, opened 2021-03-03
Increased minimum from version

ShahafBenYakir:
## Status
- [ ] In Progress
- [x] Ready

fixes: https://github.com/demisto/etc/issues/34389

## Description
Increased minimum supported version to 5.5.0


ShahafBenYakir:
CLosing as @reutshal pointed out that it is not in one place (?). I think we should fix it

PR #1157, opened 2021-03-07
fixed the issue where adding an image that's not integration image as…

merit-maita:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/34284

## Description
Fixed the issue where adding an image that's not integration image asks for a RN and a new version.

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/common/tests/tools_test.py
+++ b/demisto_sdk/commands/common/tests/tools_test.py
@@ -105,7 +105,8 @@ class TestGenericFunctions:
         (VALID_SCRIPT_PATH, FileType.SCRIPT),
         (VALID_WIDGET_PATH, FileType.WIDGET),
         (IGNORED_PNG, None),
-        ('', None)
+        ('', None),
+        ('Author_image.png', None)
     ]
 
     @pytest.mark.parametrize('path, _type', data_test_find_type)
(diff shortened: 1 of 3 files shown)


ShahafBenYakir:
What will happen for documentation images?

yaakovi (changes requested):
add unit-tests:
* for find_type - an image named `Author_image.png` returned the expected type.
* for validate manager, added/modified author image doesn't fail the release notes validation.

yaakovi:
> @merit-maita @yaakovi What will happen for documentation (README) images?

doc images have their own file type DOC_IMAGE. their behavior should not be affected

merit-maita:
@ShahafBenYakir 
as @yaakovi mentioned, one more thing, every image that does not end with _image.png, is considered as DOC_IMG, and skipped automatically, as we saw in Guy's screenshot, and it exists in the DOC_FILES folder thats also ignored. the reason why the build you mentioned failed was because it ended with _image.png, so it was considered as an image that has an effect, but it's Author_image, so it had to be excluded

demisto_sdk/commands/common/tests/tools_test.py
```diff
@@ -105,7 +105,8 @@ def test_get_dict_from_file(self, path, _type):
         (VALID_SCRIPT_PATH, FileType.SCRIPT),
         (VALID_WIDGET_PATH, FileType.WIDGET),
         (IGNORED_PNG, None),
-        ('', None)
+        ('', None),
+        ('Author_image.png', None)
```
yaakovi, line 109:
```suggestion
        ('Author_image.png', None),
```

yaakovi (approved):

PR #1162, opened 2021-03-09
adding support for local sdk conf

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/34168

## Description
adding support for `.demisto-sdk-conf` file for local customization of command runs.


demisto_sdk/__main__.py
```diff
@@ -69,6 +70,22 @@ def __init__(self):
...
+    config_file_path = '.demisto-sdk-conf'
+    if os.path.isfile(config_file_path):
+        try:
+            config = ConfigParser(allow_no_value=True)
+            config.read(config_file_path)
+
+            if command in config.sections():
+                for key in config[command]:
+                    if key in args:
+                        args[key] = config[command][key]
```
yaakovi, line 83: I think it is better to have it the other way around:
the command arguments should override the default configuration

demisto_sdk/__main__.py
```diff
@@ -69,6 +70,22 @@ def __init__(self):
 pass_config = click.make_pass_decorator(DemistoSDK, ensure=True)
 
 
+def check_configuration_file(command, args):
```
yaakovi, line 73: need to think about the behavior of conflict arguments, for example:
configuration states `-g` and command has the `-i` input.

yaakovi (approved):

PR #1177, opened 2021-03-18
Adding ignore untracked files feature to validate command

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
Adding ignore untracked files feature to validate command.

## Screenshots
A small test - 
added an RN file and made sure it was untracked:
image

Ran `demisto-sdk validate -g --staged` saw the file is being recognized and validated:
image

Ran `demisto-sdk validate -g --staged -iu` with the new flag - see that file is not validated:
image


demisto_sdk/__main__.py
```diff
@@ -250,6 +250,11 @@ def unify(**kwargs):
     help='Whether the validation should ignore unstaged files.'
          'This applies only when the -g flag is supplied.'
 )
+@click.option(
+    '-iu', '--ignore-untracked',
```
glicht, line 254: The default option should be to ignore untracked. The user shouldn't need to specify this flag. So maybe change to `include-untracked` only for cases that you want to include untracked (although I am not clear when this is needed).

glicht (changes requested):

demisto_sdk/__main__.py
roysagi: This will change the existing behavior we have now. 

It is possible though, it will create some confusion as some developers run validate locally with untracked files, like when they check RN or make changes which their env does not auto-stage the change.

Also if I recall correctly there was an issue where running the hooks with only `--staged` without checking untracked files, missed some files in the hook which were only caught in the build. 
The cause was some limitation between `pycharm` and `git` on how staged files are identified. This caused us to add the untracked files to the "staged" flag. Reversing this might cause the issue to return - not sure if we want to do that.

glicht (approved):

PR #1186, opened 2021-03-22
Fix overwrite existing files when updating existing pack

avidan-H:
## Status
- [x] Ready

## Related Issues
can see the failure in `process_pack` step of the github actions workflow [here](https://github.com/demisto/contribution-management/runs/2164144039?check_suite_focus=true)

## Description
In the flow of updating an existing content pack. After the contribution zip file would be unarchived into the destination pack directory, the contents of the unarchived directories would be moved to the corresponding directories in which they belong when in the content repo's Pack structure. e.g. `content/Packs/ExamplePack/integration/integration-exampleintegration.yml` would be moved to `content/Packs/ExamplePack/Integrations/integration-exampleintegration.yml`. This would work fine for content items like integrations and scripts because the package structure of the unified yml that was being copied to the pack was already present, no the unified. And then the `split-yml` part of contribution conversion would properly overwrite any existing files already in the package structure. But for files that are always in a unified yaml format both in the content repo as well as in a contribution zip file (as downloaded from the server), then the invocation of `shutil.move(src_file_path, dst_path)` in the `ContributionConverter`'s method `convert_contribution_dir_to_pack_contents` fails because the file already exists in the destination directory.

## Screenshots
image

## Must have
- [x] Tests
- [x] ~~Documentation~~ N/A

### DIFF
--- a/demisto_sdk/commands/init/contribution_converter.py
+++ b/demisto_sdk/commands/init/contribution_converter.py
@@ -231,7 +231,8 @@ class ContributionConverter:
                 for _, _, files in os.walk(src_path, topdown=False):
                     for name in files:
                         src_file_path = os.path.join(src_path, name)
-                        shutil.move(src_file_path, dst_path)
+                        dst_file_path = os.path.join(dst_path, name)
+                        shutil.move(src_file_path, dst_file_path)
             else:
                 # replace dst folder with src folder
                 shutil.move(src_path, dst_path)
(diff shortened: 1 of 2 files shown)


avidan-H:
@Itay4 you can see the added unit test failing in [this](https://app.circleci.com/pipelines/github/demisto/demisto-sdk/8636/workflows/4736dbe0-7394-45f7-a03d-1421f576c041/jobs/35660) circle job (from before the commits that introduce the fix).

Itay4 (approved):

PR #1189, opened 2021-03-25
Unimplemented-Warning-Error

tomneeman151293:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/35080

## Description
Fixed a problem where demisto-sdk lint did search for raise  NotImplementedError in else clauses only

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * Fixed issue where **update-release-notes** deleted release note file if command was called more than once.
 * Fixed issue where **update-release-notes** added docker image release notes every time the command was called.
 * Fixed an issue where running **update-release-notes** on a pack with newly created integration, had also added a docker image entry in the release notes.
-* Fixed an issue where `XSOAR-linter` did not find *NotImplementedError* outside else clauses in main.
+* Fixed an issue where `XSOAR-linter` did not find *NotImplementedError* in main.
 
 # 1.3.2
 * Updated the format of the outputs when using the *--json-file* flag to create a JSON file output for the **validate** and **lint** commands.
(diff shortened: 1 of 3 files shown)


tomneeman151293:
Bug was that the raise notimplemented was only checked in the else clause. fixed 
will soon post failing tests.

tomneeman151293:
@ShahafBenYakir 
Build failure
https://app.circleci.com/pipelines/github/demisto/demisto-sdk/8670/workflows/945da070-47af-4d8d-9e01-20f6914cf79a/jobs/35911
(2 tests added, both where the not implemented is inside if and in just a single line in main func)
both are failing before change

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Fixed issue where **update-release-notes** deleted release note file if command was called more than once.
 * Fixed issue where **update-release-notes** added docker image release notes every time the command was called.
 * Fixed an issue where running **update-release-notes** on a pack with newly created integration, had also added a docker image entry in the release notes.
+* Fixed an issue where `XSOAR-linter` did not find *NotImplementedError* outside else clauses in main.
```
roysagi, line 7: I think "clauses" are ambiguous maybe a different explanation about what is the change here.

demisto_sdk/commands/lint/tests/test_pylint_plugin/xsoar_level_checker_test.py
```diff
@@ -164,6 +164,51 @@ def main() -> bool:
...
+            def main() -> bool:
+                try:
+                    if True:
+                        return True
+                    if command not in commands:
+                        raise NotImplementedError("this command was not implemented")
+                    else:
+                        return True
+                except Exception:
+                    pass
```
ShahafBenYakir, line 186: Let's check elif too

```suggestion
                try:
                    if True:
                        return True
                    elif False:
                        return False
                    if command not in commands:
                        raise NotImplementedError("this command was not implemented")
                    else:
                        return True
                except Exception:
                    pass
```

ShahafBenYakir (approved):
Looks good

tomneeman151293:
Added elif test, will merge after build

PR #1201, opened 2021-04-02
splitting dependencies for layouts

roysagi:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/35366

## Description
Splitting the dependencies detection for incident and indicator layouts.

### DIFF
--- a/demisto_sdk/commands/find_dependencies/find_dependencies.py
+++ b/demisto_sdk/commands/find_dependencies/find_dependencies.py
@@ -4,7 +4,7 @@ import os
 import sys
 from copy import deepcopy
 from distutils.version import LooseVersion
-from typing import Union
+from typing import Optional, Union
 
 import click
 import networkx as nx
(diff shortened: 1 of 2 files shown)


Itay4:
in theory we shouldnt we have the same for classifiers and mappers?

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -529,22 +539,40 @@ def _collect_layouts_dependencies(pack_layouts: list,
 
         for layout in pack_layouts:
             layout_data = next(iter(layout.values()))
+            layout_path = layout_data.get("file_path")
+            layout_info = get_json(layout_path)
```
teizenman, line 543: This might be bad when we run find dependencies on all content, it's another IO operation for each layout.
Don't we have any better way to identify the type?

PR #1203, opened 2021-04-04
Lint - increase attempts to 3 and log docker error in find py ver

Itay4:
## Status
- [x] Ready

## Description
increase number of attempts to find py ver to 3 and log the exception

### DIFF
--- a/demisto_sdk/commands/common/logger.py
+++ b/demisto_sdk/commands/common/logger.py
@@ -32,7 +32,8 @@ def logging_setup(verbose: int, quiet: Optional[bool] = False,
     # Setting debug log file if in circleci
     if log_path:
         file_handler = logging.FileHandler(filename=os.path.join(log_path, 'lint_debug_log.log'))
-        file_handler.setFormatter(fmt)
+        file_handler.setFormatter(
+            logging.Formatter('[%(asctime)s] - [%(threadName)s] - [%(levelname)s] - %(message)s'))
         file_handler.setLevel(level=logging.DEBUG)
         logger.addHandler(file_handler)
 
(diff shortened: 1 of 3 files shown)


bakatzir (approved):

demisto_sdk/commands/lint/helpers.py
```diff
@@ -282,8 +282,8 @@ def get_python_version_from_image(image: str) -> float:
     """
     docker_client = docker.from_env()
     py_num = 2.7
```
yaakovi, line 284: since more and more integration and scripts are using python3, should we switch the fallback to python3?

yaakovi (approved):

demisto_sdk/commands/lint/helpers.py
Itay4: since most are on 3.8, ill set it to that

guyfreund:
@Itay4 As agreed with @hod-alpert we can merge this PR in the meantime, we are not sure it is solving the issue but it provides more logs and better error handling so in the next case of failure will know more.

demisto_sdk/commands/common/logger.py
```diff
@@ -32,7 +32,8 @@ def logging_setup(verbose: int, quiet: Optional[bool] = False,
     # Setting debug log file if in circleci
     if log_path:
         file_handler = logging.FileHandler(filename=os.path.join(log_path, 'lint_debug_log.log'))
-        file_handler.setFormatter(fmt)
+        file_handler.setFormatter(
+            logging.Formatter('[%(asctime)s] - [%(threadName)s] - [%(levelname)s] - %(message)s'))
```
Itay4, line 36: @hod-alpert why not modify `fmt` ?

PR #1228, opened 2021-04-21
added docker_util

esharf:
## Status
- [ ] In Progress
- [x] Ready

## Description
a new docker util

## Must have
- [x] Tests

### DIFF
--- a/demisto_sdk/commands/common/docker_util.py
+++ b/demisto_sdk/commands/common/docker_util.py
@@ -51,11 +51,9 @@ class ContainerRunner:
                 2. (StatusCode) the command exit code
                 3. (Error) the container stderr
         """
-        """
-            one important thing that you need to know about docker is that a dead container is basically an image
-            and that is the reason why do we need to commit to create the image.
-            so we can build a new container on the basis of this dead container.
-        """
+        # one important thing that you need to know about docker is that a dead container is basically an image.
+        # and that is the reason why do we need to commit to create the image.
+        # so we can build a new container on the basis of this dead container.
         if self._container_obj is not None:
             self._image_name = self.container.commit().id
             self.remove_container()
--- a/demisto_sdk/commands/common/docker_util.py
+++ b/demisto_sdk/commands/common/docker_util.py
@@ -43,7 +43,7 @@ class ContainerRunner:
         """
         Args:
             command: the command to run inside the docker container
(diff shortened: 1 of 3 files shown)


yaakovi (commented):
add docstrings to all public functions and to test functions (in the "When-Given-Then" format)

demisto_sdk/commands/common/docker_util.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        Examples
+            docker rmi MyTestImage
+            can be replaced with
+            DockerTools.remove_image("MyTestImage")
+            in your code
+        Args:
+            name_or_id: the name or id of the image
+            ignore_image_not_found: dont raise an exception in that the image is already not existing
+            force: like the -f in docker rmi command
```
yaakovi, line 170:
```suggestion
            ignore_image_not_found: don't raise an exception in case the image already removed
            force: same as the -f option in the "docker rmi" command
```

demisto_sdk/commands/common/docker_util.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        Examples
+            docker rm MyTestContainer
+            can be replaced with
+            DockerTools.remove_container("MyTestContainer")
+            in your code
+        Args:
+            name_or_id: the name or id of the container
+            ignore_container_not_found: dont raise an exception in that the container is already not existing
+            force: like the -f in docker rm command
```
yaakovi, line 148:
```suggestion
            ignore_container_not_found: don't raise an exception in case the container already removed
            force: same as the -f option in the "docker rm" command
```

yaakovi (approved):

demisto_sdk/commands/common/docker_util.py
```diff
@@ -0,0 +1,180 @@
...
+        self._base_image_name = image
+        self._image_name = image
+        self.container_name = container_name
+        self._container_obj: Optional[docker.models.containers.Container] = None
+
+    @property
+    def container(self) -> docker.models.containers.Container:
+        """
+        Notes:
+            try to avoid using the container in this way.
```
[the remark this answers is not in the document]
esharf: the reason for that is because it’s possible to break the logic (e.g. it’s possible to run something without commit after and so).
but I didn’t want to hide the options that the docker SDK give us.

yaakovi (approved):

PR #1229, opened 2021-04-22
handling image identification for unified ymls.

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/35960#issuecomment-823105907

## Description
handling image entity type identification for unified yml validation json output.


avidan-H (approved):
Just need a unit test. Can basically just duplicate the `test_json_outputs_where_no_image_in_integration` and instead of deleting the image, just replace it with an image that is larger than 10kb.

PR #1240, opened 2021-04-27
Supporting gitlab-ci build

hod-alpert:
## Description:
Supporting gitlab and circle build infra:
- Replacing circle env variables with gitlab env variable.
- Moving files that should be used by multiple jobs in the same pipeline to artifacts.
- Adding mechanism for querying gitlab pipelines as well as circle workflows in the locking mechanism. 
- Using gitlab env variable for slack messages in case the build is in gitlab ci.

## Related issue:
-  https://github.com/demisto/etc/issues/35417
-  https://github.com/demisto/etc/issues/35418

## Related PRs:
- https://github.com/demisto/content/pull/12203
- https://github.com/demisto/content-private/pull/344

### DIFF
--- a/demisto_sdk/commands/test_content/IntegrationsLock.py
+++ b/demisto_sdk/commands/test_content/IntegrationsLock.py
@@ -10,9 +10,10 @@ from google.cloud import storage
 
 LOCKS_PATH = 'content-locks'
 BUCKET_NAME = os.environ.get('GCS_ARTIFACTS_BUCKET')
-CIRCLE_BUILD_NUM = os.environ.get('CIRCLE_BUILD_NUM')
-WORKFLOW_ID = os.environ.get('CIRCLE_WORKFLOW_ID')
+BUILD_NUM = os.environ.get('CI_BUILD_ID')
+WORKFLOW_ID = os.environ.get('CI_PIPELINE_ID')
 CIRCLE_STATUS_TOKEN = os.environ.get('CIRCLECI_STATUS_TOKEN')
+GITLAB_STATUS_TOKEN = os.environ.get('GITLAB_STATUS_TOKEN')
 
 
 @contextmanager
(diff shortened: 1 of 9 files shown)


DinaMeylakh (changes requested):
Let's talk about this, we might be able to do better than to change all the constants.  Maybe add one class to rule them all, dynamically initializing depending on if the run is from circle or gitlab.

DinaMeylakh (approved):

demisto_sdk/commands/test_content/IntegrationsLock.py
```diff
@@ -98,10 +99,20 @@ def workflow_still_running(workflow_id: str, test_playbook) -> bool:
         return True
     else:
         try:
-            workflow_details_response = requests.get(f'https://circleci.com/api/v2/workflow/{workflow_id}',
-                                                     headers={'Accept': 'application/json'},
-                                                     auth=(CIRCLE_STATUS_TOKEN, ''))
-            workflow_details_response.raise_for_status()
+            if os.environ.get('CIRCLECI'):
```
yaakovi, line 102: shouldn't the CI be checked according to the workflow?
scenario:
* Circle workflow locks an integration
* Gitlab workflow want to lock the same integration

shouldn't Gitlab query Circle?

also, need to update the function doc-string.

avidan-H (commented):
If the `ARTIFACTS_FOLDER` environment variable is supposed to be defined in our instances, (as it seems to me from the change you made in `ParallelLoggingManager.py`), how come we aren't using it (with a reasonable default path) when setting environment variables for paths of various files that should be inside the artifacts directory instead of hardcoding the string? For example, `ENV_RESULTS_PATH = './artifacts/env_results.json'` in `TestContentClasses.py` and `FILTER_CONF` in `constants.py`. That way, if we need to change the artifacts directory path, we don't need to update all these hardcoded paths again.

demisto_sdk/commands/test_content/IntegrationsLock.py
```diff
@@ -10,9 +10,10 @@
 
 LOCKS_PATH = 'content-locks'
 BUCKET_NAME = os.environ.get('GCS_ARTIFACTS_BUCKET')
-CIRCLE_BUILD_NUM = os.environ.get('CIRCLE_BUILD_NUM')
-WORKFLOW_ID = os.environ.get('CIRCLE_WORKFLOW_ID')
+BUILD_NUM = os.environ.get('CI_BUILD_ID')
+WORKFLOW_ID = os.environ.get('CI_PIPELINE_ID')
 CIRCLE_STATUS_TOKEN = os.environ.get('CIRCLECI_STATUS_TOKEN')
+GITLAB_STATUS_TOKEN = os.environ.get('GITLAB_STATUS_TOKEN')
```
avidan-H, line 16: I just want to confirm I understand correctly. What's the reason we aren't unifying this to a single `STATUS_TOKEN` environment variable (or something along those lines)? Because for the span of time in which we're going to have the build running in both CircleCI and GitLab, we'll be considering both builds with regards to locking/unlocking integrations for a test playbook?

yaakovi (approved):
you haven't updated the doc-string for `IntegrationsLock.py:workflow_still_running`

PR #1244, opened 2021-04-29
Test Content - clean server config and reset containers if needed

Itay4:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29349

## Description
- added reset containers before applying the server conf
- cleared server conf after test
- added reset containers if server conf were applied

## Must have
- [x] Tests

### DIFF
--- a/demisto_sdk/commands/test_content/TestContentClasses.py
+++ b/demisto_sdk/commands/test_content/TestContentClasses.py
@@ -285,14 +285,17 @@ class TestPlaybook:
         """
         for integration in self.integrations:
             integration.disable_integration_instance(client)
-        updated_keys = self._clear_server_keys(client)
+        updated_keys = self._set_prev_server_keys(client, server_context)
         if updated_keys:
             server_context._reset_containers()
 
-    def _clear_server_keys(self, client: DefaultApi) -> bool:
-        """Deletes server configurations, by setting the value to empty string, that were applied for a test, if applied
+    def _set_prev_server_keys(self, client: DefaultApi, server_context: 'ServerContext') -> bool:
+        """Sets previous stored system (server) config, if existed
+
         Args:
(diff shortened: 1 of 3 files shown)


Itay4:
nightly build trigger - https://app.circleci.com/pipelines/github/demisto/content/82790/workflows/fe81fcd1-4cab-40b4-bbf1-e023f5940b98/jobs/342280

demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -270,14 +273,44 @@ def configure_integrations(self, client: DefaultApi) -> bool:
...
             client: The demisto_client to use
+            server_context (ServerContext): The ServerContext instance in which the TestContext instance is created in
         """
         for integration in self.integrations:
             integration.disable_integration_instance(client)
+        updated_keys = self._clear_server_keys(client)
+        if updated_keys:
+            server_context._reset_containers()
+
+    def _clear_server_keys(self, client: DefaultApi) -> bool:
```
hod-alpert, line 292: Is there a chance that the server keys configured for a specific integration will be using key that should have non empty value for the rest of the build?
such as `python.pass.extra.keys` or any of the sort?

if there is no such case - then it's LGTM.

But if there is - maybe it's better to modify the `update_server_configuration` to return the previous configuration and cache them in the server_context somewhere and the use them when cleaning the configuration, otherwise you might clean necessary conf.

Itay4: as discussed, i think we're good here as empty string is the default taken on the backend

hod-alpert (approved):

Itay4:
@glicht appreciate if you could have a look as well

hod-alpert (approved):

PR #1251, opened 2021-05-03
find-dependencies will run on private repos

jochman:
find-dependencies will merge the dependencies from local and content repo.

fixes: https://github.com/demisto/etc/issues/33848


jochman:
image
image

jochman:
@idovandijk fyi

demisto_sdk/__main__.py
```diff
@@ -912,11 +912,9 @@ def id_set_command(**kwargs):
 
 
 # ====================== merge-id-sets ====================== #
-@main.command(name='merge-id-sets',
-              hidden=True,
-              short_help='Merge two id_sets')
-@click.help_option(
-    '-h', '--help'
```
roysagi, line 919: why remove the help options?

roysagi (commented):
This does seem good but since this is the find-dependencies command here we should run a content build (regular, nightly, and upload flow if possible) to check this change does not affect the current flows.

demisto_sdk/__main__.py
jochman: We don't need to specify it. `--help` will always be there, even without it.
The help option will only enable `-h`, an option that is not needed.

jochman:
this pr has been scrambled.
https://github.com/demisto/demisto-sdk/pull/1264

PR #1257, opened 2021-05-04
Update baseintegration template

jochman:
Update template to a better convention.

Minimum calls to `demiso` class
updated docker image 
readabillity improvements

### DIFF
--- a/demisto_sdk/commands/init/templates/BaseIntegration/BaseIntegration.py
+++ b/demisto_sdk/commands/init/templates/BaseIntegration/BaseIntegration.py
@@ -15,23 +15,19 @@ https://github.com/demisto/content/blob/master/Packs/HelloWorld/Integrations/Hel
 """
 
 import traceback
-from typing import Any, Dict
-
-import requests
+from typing import Any, Optional
 
 import demistomock as demisto
+import urllib3
 from CommonServerPython import *  # noqa # pylint: disable=unused-wildcard-import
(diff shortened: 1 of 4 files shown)


demisto_sdk/commands/init/templates/BaseIntegration/BaseIntegration.py
```diff
@@ -83,93 +79,80 @@ def test_module(client: Client) -> str:
...
 
 
 # TODO: REMOVE the following dummy command function
 def baseintegration_dummy_command(client: Client, args: Dict[str, Any]) -> CommandResults:
-
-    dummy = args.get('dummy', None)
-    if not dummy:
-        raise ValueError('dummy not specified')
+    # dummy is a required argument
+    dummy = args['dummy']
```
roysagi, line 93: Don't we want to refrain from using direct access as it is less safe? @Itay4 what do you think?

jochman: I think that if an argument is required, we should is direct access as if it fails, it is a problem. Also, a much clearer code.
If it is not required, we should use the `.get` method.

Itay4: i think it would just be another thing for us to explain to contributions so i'd just use `.get()` for all to avoid confusion

roysagi (approved):

PR #1260, opened 2021-05-05
Fixed a bug in postman-codegen

jochman:
## Status
- [ ] In Progress
- [z] Ready

## Description
postman-codegen failed when the collection contained folders and subfolders.


demisto_sdk/commands/common/tools.py
```diff
@@ -1756,3 +1757,25 @@ def get_approved_tags() -> list:
         'Tests/Marketplace/approved_tags.json',
         github_repo=GithubContentConfig.OFFICIAL_CONTENT_REPO_NAME
     ).get('approved_list', [])
+
+
+def get_pack_metadata(file_path: str) -> dict:
```
anara123, line 1762: Why do you need this function? I don't see it used anywhere.

anara123 (approved):
see my comment. And open an issue so we could track the enhancement better

PR #1264, opened 2021-05-05
private repo to content dependencies check

jochman:
find-dependencies will merge the dependencies from local and content repo.

fixes: https://github.com/demisto/etc/issues/33848
image
image

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -911,11 +911,9 @@ def id_set_command(**kwargs):
 
 
 # ====================== merge-id-sets ====================== #
-@main.command(name='merge-id-sets',
-              hidden=True,
-              short_help='Merge two id_sets')
-@click.help_option(
-    '-h', '--help'
+@main.command(
+    hidden=True,
+    short_help='Merge two id_sets'
 )
 @click.option(
     '-i1', '--id-set1', help='First id_set.json file path', required=True
@@ -926,7 +924,7 @@ def id_set_command(**kwargs):
 @click.option(
     '-o', '--output', help='File path of the united id_set', required=True
 )
-def merge_id_sets_command(**kwargs):
+def merge_id_sets(**kwargs):
     check_configuration_file('merge-id-sets', kwargs)
     first = kwargs['id_set1']
     second = kwargs['id_set2']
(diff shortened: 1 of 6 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -21,22 +21,23 @@
 import requests
 import urllib3
 import yaml
+# disable insecure warnings
```
roysagi, line 24: why did you move this comment?

demisto_sdk/__main__.py
```diff
@@ -911,11 +911,9 @@ def id_set_command(**kwargs):
 
 
 # ====================== merge-id-sets ====================== #
-@main.command(name='merge-id-sets',
-              hidden=True,
-              short_help='Merge two id_sets')
-@click.help_option(
-    '-h', '--help'
```
roysagi, line 918: why remove the help?

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -1151,21 +1178,28 @@ def build_dependency_graph(pack_id: str, id_set: dict, verbose: bool,
                 leaf_dependencies = PackDependencies._find_pack_dependencies(
                     leaf, id_set, verbose=verbose, exclude_ignored_dependencies=exclude_ignored_dependencies)
 
-                if leaf_dependencies:
```
roysagi, line 1154: I think this handles `leaf_dependencies=None` which was an issue in the past.

roysagi (changes requested):

demisto_sdk/__main__.py
jochman: Because it is uneeded. It will use global help.
What going on here is adding -h as help. --help will always work.

demisto_sdk/commands/common/tools.py
jochman: by mistake. I've removed it

jochman:
sdk nightly and upload trigger builds: https://app.circleci.com/pipelines/github/demisto/content?branch=test-new-deps

roysagi (approved):

PR #1266, opened 2021-05-06
Fix another branch finding issue

jochman:
## Status
- [ ] In Progress
- [x] Ready

@teizenman

### DIFF
--- a/demisto_sdk/commands/common/constants.py
+++ b/demisto_sdk/commands/common/constants.py
@@ -817,7 +817,7 @@ class GithubContentConfig:
     def __init__(self, repo_name: Optional[str] = None):
         if not repo_name:
             try:
-                urls = GitUtil().repo.remote().urls
+                urls = list(GitUtil().repo.remote().urls)
                 self.CURRENT_REPOSITORY = self._get_repository_name(urls)
             except (InvalidGitRepositoryError, AttributeError):  # No repository
                 self.CURRENT_REPOSITORY = self.OFFICIAL_CONTENT_REPO_NAME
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -208,24 +209,29 @@ def get_remote_file(
...
                     'Authorization': f"Bearer {githhub_config.Credentials.TOKEN}",
                     'Accept': f'application/vnd.github.VERSION.raw',
-                }
+                })  # Sometime we need headers
+                if not res.ok:  # sometime we need param token
+                    res = github_request(params={'token': githhub_config.Credentials.TOKEN})
+                res.raise_for_status()
             else:
-                print_color(
+                raise DemistoException(
```
roysagi, line 226: this will stop the command run - do we want that?

demisto_sdk/commands/common/constants.py
```diff
@@ -817,7 +817,7 @@ class GithubContentConfig:
     def __init__(self, repo_name: Optional[str] = None):
         if not repo_name:
             try:
-                urls = GitUtil().repo.remote().urls
+                urls = list(GitUtil().repo.remote().urls)
```
roysagi, line 820: doesn't `.urls` return a list already?

demisto_sdk/commands/common/tools.py
```diff
@@ -208,24 +209,29 @@ def get_remote_file(
     tag = tag.replace('origin/', '').replace('demisto/', '')
 
     github_path = urljoin(github_config.CONTENT_GITHUB_LINK, tag, full_file_path)
-
+    github_request = partial(requests.get, github_path, verify=False, timeout=10)
```
roysagi, line 212: don't like partial here. the way it was before was way clearer please revert this change.

demisto_sdk/commands/common/constants.py
jochman: a generator. 
So when you're trying to iterate over it again it fails.

demisto_sdk/commands/common/tools.py
jochman: I think it's much more clearer 😉

demisto_sdk/commands/common/tools.py
jochman: It'll not fail as it's in the outer exception. But I've dealt with it.

roysagi (approved):

PR #1277, opened 2021-05-10
Added a configuration decorator

jochman:
## Status
- [x] In Progress
- [ ] Ready

## Description
A decorator to use the configuration with demisto-sdk

new:
You can use now type hints with the decorator itself as demonstrated in the `split-yml` command.
Use the `demisto-sdk configuration_file_path` to check where you should put your configuration in any given situation.
If running in a git repo - it'll use its root.
else it'll use the default path of `click.get_app_dir()`

Use just like the current usage:
put the `.demisto-sdk-conf` in the root of your git repository.
whenever you'll use any command with a configuration file - it'll automatically take it.

Example:
Use the configuration file:
```ini
[ split-yml ]
no_demisto_mock=True
no_common_server=True
no_auto_create_dir=True
no_pipenv=True
```
And run with `demisto-sdk split-yml -i Packs/ThreatExchange/Integrations/integration-ThreatExchange.yml -o .` 
and you'll see it works and non of Pipfile, CommonServerPython, demistomock were created.

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -4,6 +4,7 @@ import logging
 import os
 import sys
 from configparser import ConfigParser, MissingSectionHeaderError
+from functools import partial
 from pathlib import Path
 from typing import IO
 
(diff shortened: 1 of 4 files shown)


tomneeman151293:
@jochman Whats the status here?

tomneeman151293:
Closing until status is supplied

PR #1298, opened 2021-05-25
Checking  playbooks

merit-maita:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/37285

## Description
added more explanation to PB110 error message.

## Must have
- [ ] Tests
- [ ] Documentation


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Changelog
+* Added some explanation to PB110 error.
```
yaakovi, line 2: no need to update the changelog

demisto_sdk/commands/common/errors.py
```diff
@@ -959,7 +959,9 @@ def invalid_script_id(script_entry_to_check, pb_task):
     def content_entity_version_not_match_playbook_version(main_playbook, entities_names, main_playbook_version):
         return f"Playbook {main_playbook} with version {main_playbook_version} uses {entities_names} " \
                f"with a version that does not match the main playbook version. The from version of" \
-               f" {entities_names} should be at least {main_playbook_version}."
+               f" {entities_names} should be at least {main_playbook_version}.\n" \
+               f"Possible reason for such an error, would be that the name of the sub-Pb in the parent Pb's yml file" \
```
yaakovi, line 963: better to separate this option into another error.
when validating a playbook we search each sub-playbook in the ID set.
if we don't find it there - raise an error (no such playbook).
if we do find it, check the version and raise this error if needed (the logic we have now)

yaakovi (changes requested):
don't forget to add a unit test

demisto_sdk/commands/common/hook_validations/id.py
```diff
@@ -289,31 +289,37 @@ def is_entity_version_match_playbook_version(self, implemented_entity_list_from_
         Returns:
             bool. Whether the playbook's version match playbook's entities.
         """
+        invalid_version_entities = []
+        is_valid = True
         for entity_data_dict in entity_set_from_id_set:
-            if not implemented_entity_list_from_playbook:
-                return True
```
yaakovi, line 294: restore those lines before the loop (performance improvement)

demisto_sdk/commands/common/hook_validations/id.py
```diff
@@ -289,31 +289,37 @@ def is_entity_version_match_playbook_version(self, implemented_entity_list_from_
...
 
             if is_entity_used_in_playbook:
                 entity_version = all_entity_fields.get("fromversion", "")
                 is_version_valid = not entity_version or LooseVersion(entity_version) <= LooseVersion(
                     main_playbook_version)
-                if is_version_valid:
-                    implemented_entity_list_from_playbook.remove(entity_name)
+                if not is_version_valid:
+                    invalid_version_entities.append(entity_name)
+                implemented_entity_list_from_playbook.remove(entity_name)
```
yaakovi, line 308: better to copy the list before as this is causing a side effect of clearing the list (it is past by reference to the function) and might affect later logic in the outer scopes.

demisto_sdk/commands/common/tests/id_test.py
```diff
@@ -921,6 +930,10 @@ def test_are_playbook_entities_versions_valid(self, repo, mocker):
             self.playbook_with_invalid_sub_playbook_version, pack.path)
         assert not is_sub_playbook_version_invalid
 
+        is_sub_playbook_invalid = self.validator._are_playbook_entities_versions_valid(
+            self.playbook_with_sub_playbook_not_in_id_set, pack.path)
+        assert not is_sub_playbook_invalid
```
yaakovi, line 935: check that the right error was raised

demisto_sdk/commands/common/hook_validations/id.py
```diff
@@ -397,7 +409,7 @@ def is_file_valid_in_set(self, file_path, file_type, ignored_errors=None):
                 is_valid = self._is_mapper_incident_types_found(mapper_data)
             elif file_type == constants.FileType.PLAYBOOK:
                 playbook_data = get_playbook_data(file_path)
-                is_valid = self._are_playbook_entities_versions_valid(playbook_data, file_path)
+                is_valid, error = self._are_playbook_entities_versions_valid(playbook_data, file_path)
```
yaakovi, line 412: why do you need to return the error?
it is not used
```suggestion
                is_valid, _ = self._are_playbook_entities_versions_valid(playbook_data, file_path)
```

demisto_sdk/commands/common/hook_validations/id.py
```diff
@@ -290,8 +293,11 @@ def is_entity_version_match_playbook_version(self, implemented_entity_list_from_
             bool. Whether the playbook's version match playbook's entities.
         """
         invalid_version_entities = []
-        is_valid = True
+        implemented_entities = implemented_entity_list_from_playbook.copy()
```
yaakovi, line 296: use `implemented_entities` across the loop (lines 304 and 406) - as we remove items, the search in the list will be faster each time.

yaakovi (approved):

PR #1299, opened 2021-05-25
Format using git

merit-maita:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/37350

## Description
Fixed **format** to use git as a default value.

## Must have
- [ ] Tests
- [ ] Documentation


demisto_sdk/commands/format/format_module.py
```diff
@@ -86,8 +86,9 @@ def format_manager(input: str = None,
     Returns:
         int 0 in case of success 1 otherwise
     """
-    if input:
-        files = get_files_in_dir(input, ['json', 'yml', 'py', 'md'])
+    if input != 'None':
+        if isinstance(input, str):
```
yaakovi, line 90: i think it is better to keep this code as it was and fix the function call in `__main__.py`:
```python
format_manager(
    input=str(input) if input else None,
    ...
)
```

@roysagi WDYT?

CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 # Changelog
-
+* Fixed **format** to use git as a default value.
```
yaakovi, line 2:
```suggestion
* Fixed **format** to use git as a default value.

```

demisto_sdk/commands/format/format_module.py
```diff
@@ -87,7 +87,8 @@ def format_manager(input: str = None,
         int 0 in case of success 1 otherwise
     """
     if input:
-        files = get_files_in_dir(input, ['json', 'yml', 'py', 'md'])
+        if isinstance(input, str):
+            files = get_files_in_dir(input, ['json', 'yml', 'py', 'md'])
```
yaakovi, line 91: condition is redundant as we cast the input to string or passing None (which will not get here)
```suggestion
        files = get_files_in_dir(input, ['json', 'yml', 'py', 'md'])
```

yaakovi (approved):

PR #1308, opened 2021-05-30
remove DISABLE_SDK_VERSION_CHECK env var

jochman:
## Status
- [ ] In Progress
- [x] Ready

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -390,7 +390,7 @@ def get_last_remote_release_version():
 
     :return: tag
     """
-    if not os.environ.get('CI'):  # Check only when no on CI. If you want to disable it - use `DEMISTO_SDK_SKIP_VERSION_CHECK` environment varablie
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -390,7 +390,7 @@ def get_last_remote_release_version():
 
     :return: tag
     """
-    if not os.environ.get('DEMISTO_SDK_SKIP_VERSION_CHECK') and not os.environ.get('CI'):
+    if not os.environ.get('CI'):  # Check only when no on CI. If you want to disable it - use `DEMISTO_SDK_SKIP_VERSION_CHECK` environment varablie
```
yaakovi, line 393:
```suggestion
    if not os.environ.get('CI'):  # Check only when not on CI. If you want to disable it - use `DEMISTO_SDK_SKIP_VERSION_CHECK` environment variable
```

roysagi (approved):

PR #1314, opened 2021-05-31
updating ruamel

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Description
Rummel has broken BC with their safe loader at version 0.17.5 (also checked the new 0.17.6) and the issue persists.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -34,7 +34,7 @@ PyPDF2>=1.26.0
 pyspellchecker>=0.6.0
 PyYAML>=5.3.1
 requests>=2.22.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -34,7 +34,7 @@ PyPDF2>=1.26.0
 pyspellchecker>=0.6.0
 PyYAML>=5.3.1
 requests>=2.22.0
-ruamel.yaml>=0.16.5
+ruamel.yaml>=0.16.5, <0.17.5
```
Itay4, line 37: lets just lock a specific version

PR #1324, opened 2021-06-01
Fix format command issue when not having an input

PA-Rweins:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/23560, https://github.com/demisto/etc/issues/37649

## Description
Solves issue when format command doesn't have an input which caused format command to not properly.
Solved by checking if it's None before casting to string.

## Must have
- [x] Tests
- [ ] Documentation


teizenman (approved):
Nice

PR #1362, opened 2021-06-13
set IF100 to core packs only

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Description
set IF100 (bad words in incident fields) to core packs only.


roysagi (approved):
Add changelog entry please

PR #1376, opened 2021-06-16
Find dependencies for send notification

MosheEichler:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/29940

## Description
Remove the integrations that implement send-notification command from the _collect_playbooks_dependencies function

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


demisto_sdk/commands/create_id_set/tests/create_id_set_test.py
```diff
@@ -198,42 +203,114 @@ def test_create_command_to_implemented_integration_map(repo):
...
+
+        id_set_creator.add_command_to_implementing_integrations_mapping()
+
+        playbook1 = id_set_creator.id_set['playbooks'][0]['Playbook1']
+        playbook2 = id_set_creator.id_set['playbooks'][1]['Playbook2']
+        assert playbook1['command_to_integration']['specific-command'] == ['SecondaryInteg']
+        assert playbook1['command_to_integration']['generic-command'] == ['MainInteg', 'SecondaryInteg']
+        assert playbook1['command_to_integration']['send-notification'] == ''
+        assert playbook2['command_to_integration']['no-integration'] == ''
+        assert playbook2['command_to_integration']['send-notification'] == 'Slack'
```
reutshal, line 316: Not sure I understand this test, it contains more than one use case.
What is playbook 1? what is 2? what are we expecting? 
I suggest splitting it into two tests.

reutshal (changes requested):

demisto_sdk/commands/create_id_set/tests/create_id_set_test.py
```diff
@@ -299,3 +299,97 @@ def test_do_not_modify_specific_brand(repo):
...
+    def test_generic_command_that_uses_a_specific_brand(repo):
+        """
+        Given
+        - A playbook with a generic command (send-notification, etc.) that uses a specific brand
+
+        When
+        - Updating the commands_to_integrations fields in playbooks
+
+        Then
+        - Do not update 'command_to_integration' fields with any other brands
```
reutshal, line 360: We do need to update in this case aren't we?

reutshal (changes requested):
See my final comments

demisto_sdk/commands/create_id_set/tests/create_id_set_test.py
MosheEichler: The reason we do not need to update, is because `command_to_integration` is already updated by 
Screen Shot 2021-07-25 at 12 40 12
from this file
demisto_sdk/commands/common/update_id_set.py.

the purpose of `id_set_creator.add_command_to_implementing_integrations_mapping()` function is to update `command_to_integration` in case of a **_non-generic-command_** that is **_not_** depend on a specific brand

reutshal (approved):

PR #1392, opened 2021-06-23
added log info for empty cov report

esharf:
## Status
- [ ] In Progress
- [x] Ready

### DIFF
--- a/demisto_sdk/commands/lint/helpers.py
+++ b/demisto_sdk/commands/lint/helpers.py
@@ -514,8 +514,10 @@ def generate_coverage_report(html=False, xml=False, report=True, cov_dir='covera
         try:
             cov.report(file=report_data)
         except coverage.misc.CoverageException as warning:
-            logger.warning(str(warning))
-            return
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/lint/helpers.py
```diff
@@ -514,8 +514,10 @@ def generate_coverage_report(html=False, xml=False, report=True, cov_dir='covera
         try:
             cov.report(file=report_data)
         except coverage.misc.CoverageException as warning:
-            logger.warning(str(warning))
-            return
+            if isinstance(warning.args, tuple) and warning.args and warning.args[0] == 'No data to report.':
+                logger.info(f'No coverage data in file {cov_file}')
+                return
+            raise warning
```
roysagi, line 520: why raise? why not just leave it as a warning?

esharf: because the reason for this try-except block was just to catch this kind of error

roysagi (approved):

PR #1399, opened 2021-06-23
Apply is_valid_file validation from v2 validation to every version number

tomneeman151293:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/38487

## Description
Validation of correct display name for every versioned script or integration to be supported by any v<number> and not only v2

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/integration.py
+++ b/demisto_sdk/commands/common/hook_validations/integration.py
@@ -1,7 +1,7 @@
 import os
 import re
-from typing import Dict
-from typing import Optional
+from typing import Dict, Optional
+
 import yaml
 from demisto_sdk.commands.common.constants import (BANG_COMMAND_NAMES,
                                                    DBOT_SCORES_DICT,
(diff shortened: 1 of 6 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -1554,20 +1554,24 @@ def arg_to_list(arg: Union[str, List[str]], separator: str = ",") -> List[str]:
     return [arg]
 
 
-def is_v2_file(current_file, check_in_display=False):
```
roysagi, line 1557: please make sure this method is not used anywhere else.

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -954,13 +954,15 @@ def all_feed_params_exist(self) -> bool:
 
     def is_valid_display_name(self):
         # type: () -> bool
-        if not is_v2_file(self.current_file, check_in_display=True):
+        maybe_version_number: Optional[str] = get_file_version_suffix_if_exists(self.current_file,
```
roysagi, line 957:
```suggestion
        version_number: Optional[str] = get_file_version_suffix_if_exists(self.current_file,
```

demisto_sdk/commands/common/tools.py
tomneeman151293: Made sure

dantavori (approved):
awesome

PR #1415, opened 2021-06-28
Fail merge id set dups

PA-Rweins:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/38532

## Description
Created a flag for the merge-id-sets command to fail the command if duplicates are found.

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -1032,19 +1032,26 @@ def create_id_set(**kwargs):
     help='File path of the united id_set',
     required=True
 )
+@click.option(
+    '-fd',
+    '--fail-duplicates',
+    help="Fails the process if any duplicates are found.",
+    is_flag=True
+)
 def merge_id_sets(**kwargs):
     """Merge two id_sets"""
     check_configuration_file('merge-id-sets', kwargs)
     first = kwargs['id_set1']
     second = kwargs['id_set2']
     output = kwargs['output']
+    fail_duplicates = kwargs['fail_duplicates']
 
     _, duplicates = merge_id_sets_from_files(
         first_id_set_path=first,
         second_id_set_path=second,
         output_id_set_path=output
     )
-    if duplicates:
+    if duplicates and fail_duplicates:
         print_error(f'Failed to merge ID sets: {first} with {second}, '
                     f'there are entities with ID: {duplicates} that exist in both ID sets')
         sys.exit(1)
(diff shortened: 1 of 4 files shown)


demisto_sdk/commands/common/update_id_set.py
```diff
@@ -1900,8 +1895,8 @@ def has_duplicate(id_set_subset_list, id_to_check, object_type=None, print_logs=
             dict2_from_version < dict1_to_version <= dict2_to_version,  # will catch (C, B), (C, A)
         ]):
             print_warning('The following {} have the same ID ({}) and their versions overlap: '
-                          '"1.{}-{}", "2.{}-{}".'.format(object_type, id_to_check, dict1_from_version, dict1_to_version,
-                                                         dict2_from_version, dict2_to_version))
+                          '1) "{}-{}", 2) "{}-{}".'.format(object_type, id_to_check, dict1_from_version,
+                                                           dict1_to_version, dict2_from_version, dict2_to_version))
```
Bargenish, line 1899: let's change to `f string` instead of the `format` usage for readability purpose.

demisto_sdk/__main__.py
```diff
@@ -1235,19 +1235,26 @@ def create_id_set(**kwargs):
...
+    fail_duplicates = kwargs['fail_duplicates']
 
     _, duplicates = merge_id_sets_from_files(
         first_id_set_path=first,
         second_id_set_path=second,
         output_id_set_path=output
     )
-    if duplicates:
+    if duplicates and fail_duplicates:
         print_error(f'Failed to merge ID sets: {first} with {second}, '
```
Bargenish, line 1258: Let's keep the duplicates print even if the flag is set to False for logging purposes.

@yaakovi following our conv.

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -1881,14 +1881,9 @@ def has_duplicate(id_set_subset_list, id_to_check, object_type=None, print_logs=
         if object_type == 'Layouts':
             if dict1.get('kind', '') != dict2.get('kind', ''):
                 return False
-            if dict1.get('typeId', '') != dict2.get('typeId', ''):
+            if dict1.get('typeID', '') != dict2.get('typeID', ''):
                 return False
 
-        # If they have the same pack name they actually the same entity.
-        # Added to support merge between two id-sets that contain the same pack.
```
yaakovi, line 1888: will this be possible as a legitimate case? or should it always be considered as a duplicate?

PA-Rweins: This comment refers to the original condition created by Chanoch. We had that meeting bach then when we decided that it should be considered as a duplicate.

yaakovi (approved):

PR #1421, opened 2021-06-29
Add filters/transformers to dependencies check

adi88d:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/25623

## Description
- Added the methods `get_filters_and_transformers_from_complex_value` and `get_filters_and_transformers_from_playbook`.
- Updated the following methods to calculate filters and transformers in the dependencies data output:
  - `get_playbook_data`
  - `get_classifier_data`
  - `get_mapper_data`
- Updated the following methods to take into account the filters an the transformers as script dependencies:
  - `_collect_playbooks_dependencies`
  - `_collect_classifiers_dependencies`
  - `_collect_mappers_dependencies`
- Add tests
## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


yaakovi (commented):
very nice

demisto_sdk/commands/find_dependencies/tests/find_dependencies_test.py
```diff
@@ -192,6 +192,7 @@ def create_content_repo():
     create_a_pack_entity(common_scripts, FileType.SCRIPT, 'EmailAskUser')
     create_a_pack_entity(common_scripts, FileType.SCRIPT, 'ScheduleCommand')
     create_a_pack_entity(common_scripts, FileType.SCRIPT, 'DeleteContext')
+    create_a_pack_entity(common_scripts, FileType.SCRIPT, 'IsInCidrRanges')
```
adi88d, line 195:
```suggestion
    create_a_pack_entity(common_scripts, FileType.SCRIPT, 'IsInCidrRanges')
```
@yaakovi do you think we need to add FileType.Transformer and FileType.filter?

yaakovi: don't think so, it will cause a lot of overhead since they should be treated as scripts in most places

demisto_sdk/commands/common/tests/update_id_set_test.py
```diff
@@ -501,6 +501,8 @@ class TestPlaybooks:
         "name": "Dummy Playbook",
         "file_path": TESTS_DIR + "/test_files/DummyPack/Playbooks/DummyPlaybook.yml",
         "fromversion": "4.5.0",
+        "filters": ["isEqualString"],
+        "transformers": ["'uniq'"],
```
yaakovi, line 505:
```suggestion
        "transformers": ["uniq"],
```

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -207,12 +207,26 @@ def get_filters_and_transformers_from_playbook(data_dict: dict) -> Tuple[list, l
...
 
+    # collect complex values from playbook tasks
     for task in tasks.values():
         # gets the task value
-        complex_value = task.get('scriptarguments', {}).get('value', {}).get('complex', {})
+        if task.get('type') == 'condition':
+            for condition in task.get('conditions', []):
+                for inner_condition in condition.get('condition', []):
+                    if inner_condition:
+                        complex_values.append(inner_condition[0].get('left', {}).get('value', {}).get('complex', {}))
```
yaakovi, line 224: are you sure only the first index is relevant?

adi88d: yes, I created a task for testing with one condition that contains 2 conditions inside, as you can see, each condition is a list that contains 1 element:
image
image

adi88d:
@yaakovi  I added also the part for collecting the filters and transformers from the right side of the condition

demisto_sdk/commands/common/update_id_set.py
yaakovi: try this settings:
image

yaakovi (approved):

PR #1428, opened 2021-06-30
Add Object items (types, fields, modules, definitions) to id set 

sabaghadar:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/38699

## Description
Add Object related items (types, fields, modules, definitions)  to id set (through create-id-set command)

## Must have
- [x] Tests
- [ ] Documentation


demisto_sdk/commands/common/update_id_set.py
```diff
@@ -1076,10 +1103,124 @@ def get_general_paths(path, pack_to_create):
...
+    field_calculations_scripts = json_data.get('fieldCalcScript')
+    if field_calculations_scripts:
+        all_scripts = all_scripts.union({field_calculations_scripts})
+
+    data = create_common_entity_data(path=path, name=name, to_version=toversion, from_version=fromversion, pack=pack)
+
+    if all_associated_types:
+        data['object_types'] = list(all_associated_types)
+    if all_scripts:
+        data['scripts'] = list(all_scripts)
```
roysagi, line 1194: weird are we sure fields (incident fields included) can be related to a script? @Itay4 am I missing something?

demisto_sdk/commands/create_id_set/create_id_set.py
```diff
@@ -70,6 +70,11 @@ def create_command_to_implemented_integration_map(self):
                     command_name_to_implemented_integration_map[command] = [integration_name]
         return command_name_to_implemented_integration_map
 
+    def default_json_serialize(self, obj):
+        if isinstance(obj, set):
+            return list(obj)
+        raise TypeError
```
roysagi, line 76: not sure what is the point of this?

roysagi (changes requested):

demisto_sdk/commands/common/update_id_set.py
Itay4: yea we have:
- on field change script
- field display script
- sla breach script

demisto_sdk/commands/create_id_set/create_id_set.py
sabaghadar: for converting sets to list before json serialising. Handled that inside the parsing func eventually and didn't delete that by mistake

roysagi (commented):
@sabaghadar - this looks awesome.
Few final things:
1. let's run upload flow test and private build to see this change does not break anything in the flow, it shouldn't but we should double check.
2. Add a changelog entry.

roysagi (approved):

PR #1435, opened 2021-07-05
Increased speed ot writing the json logs

jochman:
## Status
- [ ] In Progress
- [x] Ready

Lots of ignores because of a mypy bug with lru_cache function


demisto_sdk/commands/common/tools.py
```diff
@@ -965,6 +965,7 @@ def get_dict_from_file(path: str, use_ryaml: bool = False) -> Tuple[Dict, Union[
 
 
 # flake8: noqa: C901
+@lru_cache()
```
roysagi, line 968: what happens when I run on several files?

demisto_sdk/commands/common/tools.py
```diff
@@ -1638,6 +1639,7 @@ def find_file(root_path, file_name):
     return ''
 
 
+@lru_cache()
```
roysagi, line 1642: same question as above

demisto_sdk/commands/lint/lint_manager.py
```diff
@@ -739,6 +740,7 @@ def create_json_output(self):
         logger.info('Collecting results to write to file')
         # format all linters to JSON format -
         # if any additional linters are added, please add a formatting function here
+        start = timeit.default_timer()
```
roysagi, line 743: seems redundant no?

demisto_sdk/commands/common/tools.py
jochman: it'll cache the result in the memory. if it's for different files, it'll store a map of input -> output. 
Least Recently Used - Will delete things that are not in use.

demisto_sdk/commands/lint/lint_manager.py
jochman: seems like a leftover

roysagi (approved):

PR #1438, opened 2021-07-06
fixing rn validation pack path

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/38946

## Description
Fixed an issue where the release notes validation attempted to open a pack metadata file using an incorrect path.

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/release_notes.py
+++ b/demisto_sdk/commands/common/hook_validations/release_notes.py
@@ -1,9 +1,11 @@
 from __future__ import print_function
 
 import itertools
+import os
 import re
 
-from demisto_sdk.commands.common.constants import (RN_HEADER_BY_FILE_TYPE,
+from demisto_sdk.commands.common.constants import (PACKS_DIR,
+                                                   RN_HEADER_BY_FILE_TYPE,
                                                    FileType)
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/hook_validations/release_notes.py
```diff
@@ -50,7 +53,7 @@ def are_release_notes_complete(self):
                 elif self.pack_name + '/' in file:
                     # Refer image and description file paths to the corresponding yml files
                     file = UpdateRN.check_for_release_notes_valid_file_path(file)
-                    update_rn_util = UpdateRN(pack_path=self.release_notes_file_path, modified_files_in_pack=set(),
+                    update_rn_util = UpdateRN(pack_path=self.pack_path, modified_files_in_pack=set(),
```
Itay4, line 56: can we add test?

roysagi: we can but this fixes just an incorrect print that happened not an incorrect behavior so the test seems kinda redundant and will probably not look good. The changed lines are covered in other existing tests.

Itay4 (approved):

PR #1450, opened 2021-07-07
removed dict from find_type (speed up)

jochman:
## Status
- [ ] In Progress
- [ ] Ready
- [x] In Hold - Waiting for content build

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -1,4 +1,5 @@
 import argparse
+from demisto_sdk.commands.format.format_module import logger
 import glob
 import io
 import json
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -1010,98 +1009,106 @@ def find_type_by_path(path: str = '') -> Optional[FileType]:
...
+        path - a path to the file
+
+    Returns:
+        string representing the content file type
+    """
+    type_by_path = find_type_by_path(path)
+    if type_by_path:
+        return type_by_path
+
+    return find_file_by_content(path, ignore_sub_categories)
```
roysagi, line 1111: I don't like to split this into 2 separate ones.
This method has one function and it should be kept in the same place to ease the process of following it.

This adds steps to debugging which are not necessary

roysagi (changes requested):
Let's test if this change is worthwhile, can we please check how much does it improve nightly run for validate and lint, please?

If it is not a noticeable improvement, it might be wiser to leave this as is since this method is used everywhere in the sdk and content repo.

demisto_sdk/commands/common/tools.py
jochman: The improvement is already inside.
The only thing that changes here is the Download in the content repo.

PR #1482, opened 2021-07-19
validate array for reputation default value

barchen1:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/39099

## Description
Added a validation to ensure reputation command's default argument is set as an array input.

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/integration.py
+++ b/demisto_sdk/commands/common/hook_validations/integration.py
@@ -308,7 +308,7 @@ class IntegrationValidator(ContentEntityValidator):
     def is_valid_default_argument_in_reputation_command(self):
         # type: () -> bool
         """Check if a reputation command (domain/email/file/ip/url/cve)
-            has a default non required argument.
+            has a default non required argument which support array.
 
         Returns:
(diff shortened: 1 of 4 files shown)


JasBeilin:
Carbon Black currently has no instance to use to validate the fix.

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -336,6 +336,12 @@ def is_valid_default_argument_in_reputation_command(self):
                             if self.handle_error(error_message, error_code, file_path=self.file_path):
                                 self.is_valid = False
                                 flag = False
+                        if arg.get('isArray') is False:
```
guykeller, line 339: Seems like a boolean field, what other falsy value could it have?
```suggestion
                        if not arg.get('isArray'):
```

guykeller (approved):

bakatzir (approved):
nice.
minor comment, please rename `is_valid_default_argument_in_reputation_command` to reflect the addition of isarray.

PR #1490, opened 2021-07-22
Demisto SDK OpenAPI Command Str Has No Attribute Items Fix

tomneeman151293:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/39351

## Description
Added a fix to check if the props is dict object. Because some swaggers have example entries containing string indicating the type of the object and not the structure of the object itself
## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/demisto_sdk/commands/openapi_codegen/openapi_codegen.py
+++ b/demisto_sdk/commands/openapi_codegen/openapi_codegen.py
@@ -670,6 +670,7 @@ class OpenAPIIntegration:
                     for ref in refs:
                         ref = ref.split('/')[-1]
                         ref_props = self.extract_values(self.reference.get(ref, {}), 'properties')
+                        # ref_props = [ref_prop for ref_prop in ref_props if isinstance(ref_prop, dict)]
                         if ref_props:
                             for k, prop in ref_props[0].items():
                                 if k in self.root_objects:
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/openapi_codegen/openapi_codegen.py
```diff
@@ -670,6 +670,9 @@ def extract_outputs(self, data: dict) -> tuple:
                     for ref in refs:
                         ref = ref.split('/')[-1]
                         ref_props = self.extract_values(self.reference.get(ref, {}), 'properties')
+                        # Addition of filtering dicts only was added because some swaggers contain example files
+                        # Which are written in string and caused errors on ref_props[0].items()
+                        ref_props = [ref_prop for ref_prop in ref_props if isinstance(ref_prop, dict)]
```
dantavori, line 675: is this the only area in the code where this could happen? please have a look at line 720 and other similar areas in the code, and make sure we cover all these cases

tomneeman151293: I added to line 720 too as it seems relevant. The code is very complex to be confident I handled all those cases, but I did double check now and have not found any other place which seems relevant for this specific fix

dantavori (approved):

PR #1499, opened 2021-07-26
fix validate wrong error 

abaumgarten:
## Status
- [] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/39271

## Description
Fixed an issue when validate suggested running format on the yml file when mandatory fields were missing. 
Foramt does not handle adding those fields so the print was wrong. 
My solution is an indicative message to the user which field to add and the path to add it to.

## Screenshots

image


bakatzir (commented):
nice.
please consider adding an integration test, e.g here:
https://github.com/demisto/demisto-sdk/blob/master/demisto_sdk/tests/integration_tests/validate_integration_test.py#L534

that the proper message is printed in stdout in case of a faulty yml validation

bakatzir (approved):

PR #1516, opened 2021-07-27
update_rn fix Force and text  flags

iyeshaya:
## Status
- [x] In Progress
- [x] Ready

## Description
Fixed an issue where --text flag is ignored when using -f flag


jochman (approved):
add integration test

PR #1524, opened 2021-07-28
pwsh related improvements in lint

roysagi:
## Description
 - Retry pwsh on timeout.
 - Excluding pwsh from `get_python_version_from_image` method.
 - giving the option for timeout increase in the `get_python_version_from_image` method.

### DIFF
--- a/demisto_sdk/commands/lint/helpers.py
+++ b/demisto_sdk/commands/lint/helpers.py
@@ -281,7 +281,7 @@ def add_tmp_lint_files(content_repo: git.Repo, pack_path: Path, lint_files: List
 
 
 @lru_cache(maxsize=100)
-def get_python_version_from_image(image: str, timeout: int) -> float:
+def get_python_version_from_image(image: str, timeout: int = 60) -> float:
     """ Get python version from docker image
 
     Args:
--- a/demisto_sdk/commands/lint/helpers.py
+++ b/demisto_sdk/commands/lint/helpers.py
@@ -280,18 +280,25 @@ def add_tmp_lint_files(content_repo: git.Repo, pack_path: Path, lint_files: List
 
 
 @lru_cache(maxsize=100)
-def get_python_version_from_image(image: str) -> float:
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/lint/helpers.py
```diff
@@ -281,18 +281,26 @@ def add_tmp_lint_files(content_repo: git.Repo, pack_path: Path, lint_files: List
...
     Args:
         image(str): Docker image id or name
+        timeout(int): Docker client request timeout
 
     Returns:
         float: Python version X.Y (3.7, 3.6, ..)
     """
+
+    excluded_images = {'demisto/powershell', 'demisto/powershell-ubuntu', 'demisto/pwsh-exchange',
+                       'demisto/pwsh-infocyte', 'demisto/powershell-teams'}
```
yaakovi, line 296: maybe it is best to check if the name includes `powershell`/`pwsh` instead of a fixed set of images (in case we add a new PowerShell-based image, it will require a new SDK version.

demisto_sdk/commands/lint/helpers.py
```diff
@@ -281,18 +281,26 @@ def add_tmp_lint_files(content_repo: git.Repo, pack_path: Path, lint_files: List
...
+    is_pwsh_image = any([image_name in image for image_name in excluded_images])
+    if is_pwsh_image:
+        return 3.8
```
yaakovi, line 299: why 3.8?

demisto_sdk/commands/lint/helpers.py
roysagi: thought about it, decided against it for the possibility of excluding images that happen to have `pwsh` in the name - what do you say?

demisto_sdk/commands/lint/helpers.py
roysagi: seems to be the default, really does not matter when running on pwsh

yaakovi (approved):

PR #1525, opened 2021-07-29
Update RN APIModules bug fix

dansterenson:
There 
## Status
- [ ] In Progress
- [x] Ready

## Description
The update release notes command had a bug when running update-release-notes when some Api module has changed.
since we did not pass the full pack paths, some functions failed to find the files. In addition, I changed the condition for updating Api Modules: 
We want to handle ApiModules changes when:
(1) The user gave a path to the api module which was changed.
(2) The user did not give a specific path at all (is_all = True) but some ApiModules were changed.


demisto_sdk/commands/update_release_notes/update_rn.py
```diff
@@ -728,7 +728,7 @@ def update_api_modules_dependents_rn(pre_release: bool, update_type: Union[str,
     for integration in integrations:
         integration_path = integration.get('file_path')
         integration_pack = integration.get('pack')
-        update_pack_rn = UpdateRN(pack_path=integration_pack, update_type=update_type,
+        update_pack_rn = UpdateRN(pack_path=f'Packs/{integration_pack}', update_type=update_type,
```
yaakovi, line 731: use `pack_name_to_path` from tools.py

yaakovi (approved):

demisto_sdk/commands/update_release_notes/update_rn.py
dansterenson: Done

PR #1538, opened 2021-08-05
specifying files by given pack in update_rn_manager

roysagi:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/39422

## Description
Using the prebuilt mechanism for validate's -i pack_path -g mode to get specified file only in update-rn command.


demisto_sdk/commands/update_release_notes/update_rn_manager.py
```diff
@@ -62,29 +62,65 @@ def manage_rn_update(self):
...
+                else:
+                    file_path = str(file)
+
+                file_pack_name = get_pack_name(file_path)
+                if file_pack_name not in self.given_pack:
+                    continue
+
+                filtered_set.add(file)
+
+        return validate_manager.filter_to_relevant_files(filtered_set)
```
tomneeman151293, line 84: I think we can change this function to use validate_manager, it seems like it does some duplicate code. Or maybe even add a general function for filtering files by given pack so it will be generalized and can be called from other places in the future.

Also, this function having same name as validate_manager function is a little confusing, maybe filter_relevant_files_in_pack or something like that.
```suggestion
        filtered_set = validate_manager.filter_to_relevant_files(filtered_set)
        if self.given_pack:
            filtered_set = {file_path for file_path in filtered_set if get_pack_name(file_path) in self.given_pack}
        return filtered_set
```

demisto_sdk/commands/update_release_notes/update_rn_manager.py
```diff
@@ -62,29 +62,65 @@ def manage_rn_update(self):
...
             validate_manager.setup_git_params()
-            if self.given_pack:
-                # The Validator prints errors which are related to all changed files that were changed against prev
-                # version. When the user is giving a specific pack to update, we want to suppress the error messages
-                # which are related to other packs.
-                with suppress_stdout():
-                    return validate_manager.get_changed_files_from_git()
-            return validate_manager.get_changed_files_from_git()
+
+            with suppress_stdout():
```
tomneeman151293, line 117: But what about when pack input was not given? We wouldn't want to suppress STDOUT then wouldn't we? Or maybe im not familiar with the logic here

demisto_sdk/commands/update_release_notes/update_rn_manager.py
roysagi: Here is the issue with this idea (which was my original direction) - this does not solve the "bug" element of this issue: https://github.com/demisto/etc/issues/39422

if we use validate_manager's filter command out the back it will try and look at the file_type of all files and only then filter the relevant files by path - hence it could fail on unrelated files.
to avoid this, I've moved the path filtration before the VM's filtration - might look not as clean but it is correct.

demisto_sdk/commands/update_release_notes/update_rn_manager.py
roysagi: good point fixing

demisto_sdk/commands/update_release_notes/update_rn_manager.py
tomneeman151293: Got it. Resolving

tomneeman151293 (approved):

PR #1540, opened 2021-08-05
ignored slack errors in test-content command

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Description
ignore slack errors in the test-content command (we are using it as a best-effort - it shouldn't fail the build)

### DIFF
--- a/demisto_sdk/commands/test_content/TestContentClasses.py
+++ b/demisto_sdk/commands/test_content/TestContentClasses.py
@@ -731,10 +731,7 @@ class BuildContext:
         """
         user_id = ''
         try:
-            if os.environ.get('GITLAB_USER_LOGIN'):
-                user_name = os.environ['GITLAB_USER_LOGIN']
-            else:
-                user_name = self._get_user_name_from_circle()
+            user_name = os.getenv('GITLAB_USER_LOGIN') or self._get_user_name_from_circle()
             res = self.slack_client.api_call('users.list')
 
             user_list = res.get('members', [])
--- a/demisto_sdk/commands/test_content/TestContentClasses.py
+++ b/demisto_sdk/commands/test_content/TestContentClasses.py
@@ -708,45 +708,43 @@ class BuildContext:
         return conf, secret_conf
 
     def _get_user_name_from_circle(self):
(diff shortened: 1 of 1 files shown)


roysagi (approved):

demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -708,45 +708,43 @@ def _load_conf_files(conf_path, secret_conf_path):
...
-        for user in user_list:
-            profile = user.get('profile', {})
-            name = profile.get('real_name_normalized', '')
-            if name == user_name:
-                user_id = user.get('id', '')
+        try:
+            if os.environ.get('GITLAB_USER_LOGIN'):
+                user_name = os.environ['GITLAB_USER_LOGIN']
+            else:
+                user_name = self._get_user_name_from_circle()
```
avidan-H, line 737:
```suggestion
            user_name = os.getenv('GITLAB_USER_LOGIN') or self._get_user_name_from_circle()
```

avidan-H (changes requested):
Shouldn't we just catch the specific exception?

yaakovi:
> Shouldn't we just catch the specific exception?

I would like to ignore any kind of error in this flow as it is not mandatory for the rest of the command
WDYT?

avidan-H:
> > Shouldn't we just catch the specific exception?
> 
> I would like to ignore any kind of error in this flow as it is not mandatory for the rest of the command
> WDYT?

👍🏼  I guess in that case though, not sure if the debug message makes sense since it's assuming what the error is that is being caught and it could very well be something else...

yaakovi:
> > > Shouldn't we just catch the specific exception?
> > 
> > 
> > I would like to ignore any kind of error in this flow as it is not mandatory for the rest of the command
> > WDYT?
> 
> 👍🏼 I guess in that case though, not sure if the debug message makes sense since it's assuming what the error is that is being caught and it could very well be something else...

the function name is `retrieve slack user id` so the message is relevant for the whole block in my opinion.
it may fail to grab the GitHub user but it also means that we can't find the slack user...

avidan-H (approved):

PR #1544, opened 2021-08-08
Adding generate-test-playbook enhancements

yaakovpraisler:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/39488
fixes: https://github.com/demisto/etc/issues/36567

## Description
- Adding the following arguments to the generate-test-playbook command:
   - `--examples`. path for the file containing command examples (like in `generate-docs` command) to generate also command arguments.
   - `--commands`. Comma-separated command names to generate playbook tasks for.
   - `--upload`. Whether to upload the test playbook after generating.
- Generate output conditions by types.

## Must have
- [x] Tests
- [x] Documentation


demisto_sdk/commands/generate_test_playbook/test_playbook_generator.py
```diff
@@ -271,21 +278,57 @@ def create_automation_task_and_verify_outputs_task(test_playbook, command, item_
...
+    Args:
+        commands_file_path: command file or the content of such file.
+        specific_commands: commands specified by the user.
+
+    Return:
+        dict. Arguments separated by the commands.
+    """
+    command_examples = []  # type: list
+
+    if os.path.isfile(commands_file_path):
```
roysagi, line 294: If I recall correctly `isfile` does not only check the validity of the path it also checks its existence.
so in case you are given an integration and the command examples file path is not real, it will go to the else. and this can be weird...

might be wiser to check the file type and by that chose which flow should run.

demisto_sdk/commands/generate_test_playbook/test_playbook_generator.py
```diff
@@ -271,21 +278,57 @@ def create_automation_task_and_verify_outputs_task(test_playbook, command, item_
...
-                 verbose: bool = False, use_all_brands: bool = False):
+                 verbose: bool = False, use_all_brands: bool = False, commands: str = None, examples: str = None,
+                 upload: bool = False):
         self.integration_yml_path = input
-        self.output = output
-        if output:
-            self.test_playbook_yml_path = os.path.join(output, name + '.yml')
+
+        if not output or not os.path.isdir(output):
+            self.output = f'Packs/{get_pack_name(self.integration_yml_path)}/TestPlaybooks'
```
roysagi, line 318: this is not good cause we might not be running from content repo here.
you can either:
1) check if we are running from content and then use this path. (use `is_external_repository`)
2) if we are not in content just put the file locally.

Also what if I'm running in content in a pack that does not have TestPlaybooks dir? shouldn't we need to create it before putting the full path?

demisto_sdk/commands/generate_test_playbook/test_playbook_generator.py
```diff
@@ -353,4 +405,9 @@ def run(self):
         with open(self.test_playbook_yml_path, 'w') as yf:
             ryaml.dump(test_playbook.to_dict(), yf)
 
-            print_color(f'Test playbook yml was saved at:\n{self.test_playbook_yml_path}', LOG_COLORS.GREEN)
+            print_color(f'Test playbook yml was saved at:\n{self.test_playbook_yml_path}\n', LOG_COLORS.GREEN)
```
roysagi, line 408: use click instead.

roysagi (changes requested):
@yaakovpraisler - Overall a good job, I do have some notes.
Also please update the README file for generate test playbook command.

demisto_sdk/commands/generate_test_playbook/test_playbook_generator.py
yaakovpraisler: fixed.

demisto_sdk/commands/generate_test_playbook/test_playbook_generator.py
yaakovpraisler: done.

roysagi (approved):
Great job!

PR #1548, opened 2021-08-09
fixing the description demisto check

roysagi:
## Status
- [x] In Progress
- [ ] Ready

## Description
The validation that 'demisto' is not in the description file did not occur when running straight on the description file itself.
The validation only occurred when running on the connected integration - resulting in the validation passing like here: https://github.com/demisto/content/pull/13920

This fix will make sure all validations run when running both on the file itself and on the adjacent integration.

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/description.py
+++ b/demisto_sdk/commands/common/hook_validations/description.py
@@ -3,13 +3,13 @@ from typing import Optional
 
 from demisto_sdk.commands.common.constants import (
     BETA_INTEGRATION_DISCLAIMER, PACKS_INTEGRATION_NON_SPLIT_YML_REGEX,
-    PACKS_INTEGRATION_YML_REGEX)
-from demisto_sdk.commands.common.errors import Errors
+    PACKS_INTEGRATION_YML_REGEX, FileType)
+from demisto_sdk.commands.common.errors import FOUND_FILES_AND_ERRORS, Errors
 from demisto_sdk.commands.common.hook_validations.base_validator import \
     BaseValidator
 from demisto_sdk.commands.common.hook_validations.structure import \
     StructureValidator
-from demisto_sdk.commands.common.tools import get_yaml, os, re
+from demisto_sdk.commands.common.tools import find_type, get_yaml, os, re
 
 CONTRIBUTOR_DETAILED_DESC = 'Contributed Integration'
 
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/common/hook_validations/description.py
```diff
@@ -147,28 +149,43 @@ def verify_demisto_in_description_content(self):
...
-        data_dictionary = get_yaml(self.file_path)
-        is_unified_integration = data_dictionary.get('script', {}).get('script', '') not in {'-', ''}
+        description_path = ''
         yml_line_num = 0
+        yml_or_file = ''
 
-        if is_unified_integration:
-            description_content = data_dictionary.get('detaileddescription', '')
+        # case 1 the file path is for an integration
+        if find_type(self.file_path) == FileType.INTEGRATION:
```
yaakovi, line 157: what about beta integrations? use the ignore sub-categories

demisto_sdk/commands/common/hook_validations/description.py
```diff
@@ -147,28 +149,43 @@ def verify_demisto_in_description_content(self):
...
 
-                for line_n, line in enumerate(f.read().split('\n')):
-                    if 'detaileddescription:' in line:
-                        yml_line_num = line_n + 1
-        else:
-            try:
-                description_path = glob.glob(os.path.join(os.path.dirname(self.file_path), '*_description.md'))[0]
+                # find in which line the description begins in the yml
+                with open(self.file_path, 'r') as f:
+                    for line_n, line in enumerate(f.read().split('\n')):
```
yaakovi, line 168:
```suggestion
                    for line_n, line in enumerate(f.readlines()):
```

demisto_sdk/commands/common/hook_validations/description.py
```diff
@@ -147,28 +149,43 @@ def verify_demisto_in_description_content(self):
...
 
-            except IndexError:
-                error_message, error_code = Errors.no_description_file_warning()
-                self.handle_error(error_message, error_code, file_path=self.file_path, warning=True)
-                return True
+            # if not found try and look for the description file path
+            else:
+                try:
+                    yml_or_file = 'in the description file'
+                    description_path = glob.glob(os.path.join(os.path.dirname(self.file_path), '*_description.md'))[0]
```
yaakovi, line 176: no reason to use glob
```suggestion
                    description_path = self.file_path.replace('.yml', '_description.md')
```
or
```suggestion
                    description_path = f'{os.path.splitext(self.file_path)}_description.md'
```
then you just need to make sure the file exists using `os.path.isfile` instead of the try-except.

demisto_sdk/commands/common/hook_validations/description.py
roysagi: right

yaakovi (approved):

PR #1566, opened 2021-08-15
added openEnded to incidentfield schema

jochman:
## Status
- [ ] In Progress
- [x] Ready

## Description
Will be available fromversion 6.5 of the server. Will do nothing below it

## Must have
- [x] Tests


tomneeman151293 (changes requested):
Need to add format to bump version to 6.5.0 if field exists and validation that if this field exists then the version is 6.5.0 and above

jochman:
it's fine, it's backward compatible (you can try it, just upload an incident field with `openEnded`).

tomneeman151293 (approved):

PR #1568, opened 2021-08-15
Json-to-outputs - detect timestamp "zero" as Date-type

dorschw:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40195

## Description
When checking whether `0001-01-01T00:00:00` is of type `Date`, `dateparser` tried using the local timezone, which caused an overflow. Force-using GMT as the timezone solves the issue.

## Must have
- [x] Tests
- [x] Documentation


DeanArbel (approved):

dorschw:
closed as failed tests cause too much hassle for this size of bug. using a solution that isn't as elegant, see https://github.com/demisto/etc/issues/40195

PR #1585, opened 2021-08-23
Iron bank - add validation+format for latest image

JasBeilin:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/39543

## Description
Added a validation in the **validation** command that makes sure that if a pack has "Iron Bank" tag in metadata, the image is the same as the latest in Iron Bank.
Also added to **format** command the ability to take the docker version from Iron Bank.

## Must have
- [x] Tests
- [x] Documentation


demisto_sdk/commands/common/tools.py
```diff
@@ -2004,3 +2004,7 @@ def suppress_stdout():
             yield
         finally:
             sys.stdout = old_stdout
+
+def is_iron_bank_pack(file_path):
+    metadata = get_pack_metadata(file_path)
+    return 'Iron Bank' in metadata.get('tags', [])
```
guyfreund, line 2010: Don't we have a constants file for a pack's tags?

demisto_sdk/commands/common/hook_validations/pack_unique_files.py
```diff
@@ -268,7 +268,6 @@ def validate_pack_meta_file(self):
             self._is_approved_usecases(),
             self._is_approved_tags(),
             self._is_price_changed(),
-            self._is_approved_tags(),
```
guyfreund, line 271: Did we intentionally delete this?

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -337,3 +342,69 @@ def parse_docker_image(self, docker_image):
...
+
+        try:
+            last_commit = DockerImageValidator._get_latest_commit(commits_url, docker_image_name)
+            if not last_commit:
+                return ''
+            manifest_file_content = DockerImageValidator._get_manifest_from_commit(manifest_url, last_commit)
+            if not manifest_file_content:
+                return ''
+        except Exception as e:
+            raise(e)
```
guyfreund, line 369: This has no much impact, better to use logging for debugging purposes.

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -337,3 +342,69 @@ def parse_docker_image(self, docker_image):
...
+        if res.status_code != 200:
+            raise Exception("Missing manifest file in the latest successful commit.")
+
+        return res.text
+
+    @staticmethod
+    def _get_latest_commit(commits_url, docker_image_name):
+        # Get latest commit in master which passed the pipeline of the project in Iron Bank:
+        res = requests.get(url=commits_url, params={'ref': 'master', 'status': 'success'}, verify=False,
+                           timeout=TIMEOUT)
```
guyfreund, line 395: How do you handle pagination? See: https://docs.gitlab.com/ee/api/pipelines.html#pipelines-pagination

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -337,3 +342,69 @@ def parse_docker_image(self, docker_image):
...
+
+        # Project may not be existing and needs to be created.
+        if res.status_code != 200:
+            raise Exception('The docker image in your integration/script cannot be found in Iron Bank.'
+                            f' Please create image {docker_image_name} In Iron Bank.')
+
+        list_of_commits = res.json()
```
guyfreund, line 402:
```suggestion
        list_of_pipelines= res.json()
```

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -337,3 +342,69 @@ def parse_docker_image(self, docker_image):
...
+
+        # Project seems to have no succeed pipeline for master branch, meaning the image is not in Iron Bank.
+        if not list_of_commits:
+            raise Exception('The docker image in your integration/script does not have a tag in Iron Bank.'
+                            f' Please create or update to an updated versioned image In Iron Bank.')
+
+        list_of_commits = sorted(list_of_commits, key=lambda x: x['updated_at'], reverse=True)
```
guyfreund, line 409: Please use the `order by` and `sort` fields in the API call, you can sort by the `ID` of the pipeline. See: https://docs.gitlab.com/ee/api/pipelines.html#list-project-pipelines

guyfreund (changes requested):

demisto_sdk/commands/common/tools.py
JasBeilin: No, but I'll add a constant instead.

demisto_sdk/commands/common/hook_validations/pack_unique_files.py
JasBeilin: Yes, It runs twice-
self._is_approved_usecases(),
**self._is_approved_tags(),**
self._is_price_changed(),
**self._is_approved_tags()**

demisto_sdk/commands/common/hook_validations/docker.py
JasBeilin: we cannot actually add logging (correct me if i'm wrong please) except for handle_error, which we cannot use because of the format command using the same functions. The Exception allows us to see the reason for fail in both format and validate command.

guyfreund: So what should the user see in a case of an exception?

JasBeilin: The Format will show the error I am raising.  The validate will show it cannot fetch the tag (a more generic error, same as the regular docker hub fetching error). I can edit the current generic error to include the exception I raised if you think this is the best solution.

guyfreund: I believe it's better to let the user know why the validation is failing in that case and what should he do in order to solve it.

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -292,11 +294,14 @@ def get_docker_image_latest_tag(self, docker_image_name, yml_docker_image):
                     return ''
                 return "no-tag-required"
         try:
-            return self.get_docker_image_latest_tag_request(docker_image_name)
-        except (requests.exceptions.RequestException, Exception):
+            if is_iron_bank:
+                return self.get_docker_image_latest_tag_from_iron_bank_request(docker_image_name)
+            else:
+                return self.get_docker_image_latest_tag_request(docker_image_name)
```
bakatzir, line 300:
```suggestion
            return self.get_docker_image_latest_tag_request(docker_image_name)
```

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -337,3 +342,69 @@ def parse_docker_image(self, docker_image):
...
+        """
+        Get the latest tag for a docker image by request to Iron Bank Repo.
+        Args:
+            docker_image_name: The docker image name.
+
+        Returns:
+            The latest tag for the docker image.
+        """
+        project_name = docker_image_name.replace('demisto/', '')
+        api_url = 'https://repo1.dso.mil/api/v4/projects/dsop%2Fopensource%2Fpalo-alto-networks%2Fdemisto%2F'
```
bakatzir, line 357: does this URL is bound to chagne/constant? (v4), also - might we use it in another future validation? I suggest in any case move it to `constnats`

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -337,3 +342,69 @@ def parse_docker_image(self, docker_image):
...
+        commits_url = api_url + f'{project_name}/pipelines'
+        manifest_url = api_url + f'{project_name}/repository/files/hardening_manifest.yaml/raw'
```
bakatzir, line 359: please conform to f notation fully.

bakatzir (commented):
nice work. let's create a PR in the content repo and check 2 things:
1. how much time does it take to validate the entire repo when there are ±10 packs with this tag, as I see that we are running some regexes and using the `TIMEOUT` arg in some of the static functions.
2. how does the errors will look like, feel free to comment them in the main PR so we can see.

JasBeilin:
Case of validate when not updated tag in Iron Bank pack:
image

Case of validate when using docker which we don't have an image for in Iron Bank:
image

Case of format when we using docker which we don't have an image for in Iron Bank:
image

Validate a **non** Iron Bank pack (The "normal" pack has not changed):
image

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -337,3 +342,68 @@ def parse_docker_image(self, docker_image):
...
+        if res.status_code != 200:
+            raise Exception('The docker image in your integration/script cannot be found in Iron Bank.'
+                            f' Please create image {docker_image_name} In Iron Bank.')
+
+        last_successful_pipelines = res.json()
+
+        # Project seems to have no succeed pipeline for master branch, meaning the image is not in Iron Bank.
+        if not last_successful_pipelines:
+            raise Exception('The docker image in your integration/script does not have a tag in Iron Bank.'
+                            ' Please create or update to an updated versioned image In Iron Bank.')
```
bakatzir, line 407:
```suggestion
                            ' Please use only images that are already in Iron Bank, or upload your image to it.')
```

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -337,3 +342,68 @@ def parse_docker_image(self, docker_image):
...
+    def _get_latest_commit(commits_url, docker_image_name):
+        # Get latest commit in master which passed the pipeline of the project in Iron Bank:
+        res = requests.get(url=commits_url, params={'ref': 'master', 'status': 'success',
+                                                    'order_by': 'updated_at', 'per_page': '1'},
+                           verify=False, timeout=TIMEOUT)
+
+        # Project may not be existing and needs to be created.
+        if res.status_code != 200:
+            raise Exception('The docker image in your integration/script cannot be found in Iron Bank.'
+                            f' Please create image {docker_image_name} In Iron Bank.')
```
bakatzir, line 400:
```suggestion
                            f' Please create the image: {docker_image_name} in Iron Bank.')
```

guyfreund (commented):
looks good from my side, nice work.

wissamg (changes requested):

guyfreund (approved):

wissamg (approved):

bakatzir (approved):
good work!

bakatzir (commented):
Please add a line to the changelog

PR #1588, opened 2021-08-23
Added the hidden field to the integration argument schema

dansterenson:
## Status
- [ ] In Progress
- [x] Ready

Use case: https://github.com/demisto/content/pull/14494

## Description
Add 'hidden' option to the integration schema so the validation won't fail.


Bargenish:
Let's add a link to the XDR PR as a ref example for the use case.
Please also add this field to one of the test data integrations.

Bargenish (approved):

PR #1602, opened 2021-08-29
Init non content

ostolero:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
https://github.com/demisto/etc/issues/40607

## Description
Fix a bug in Init command in SDK - When command is called from a repo that is not content we fail on error due to white-list file missing. Continue run without white list in case it doesn't exist.


demisto_sdk/commands/init/initiator.py
```diff
@@ -415,10 +415,14 @@ def find_secrets(self):
...
-                f.write(secret)
-                f.write('\n')
+        try:
+            with open(f'Packs/{pack_dir}/.secrets-ignore', 'a') as f:
+                for secret in secrets:
+                    f.write(secret)
+                    f.write('\n')
+        except FileNotFoundError:
+            print_warning("Could not find the .secrets-ignore file - make sure your path is correct and is of format "
+                          "Packs/my_pack_name")
```
roysagi, line 425: can the user really make sure? isn't this an automatic thing that happens in the init process?

roysagi (commented):
See my comment, also please add a changelog entry.

demisto_sdk/commands/init/initiator.py
ostolero: Your'e right it's redundant. will remove

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Added validation to **validate** for release notes config file.
 * Added **isoversize** and **isautoswitchedtoquietmode** fields to the playbook schema.
 * Added to the **update-release-notes** command `-bc` flag to generate template for breaking changes version.
+* Added support for **init** command to run from non-content repo.
```
roysagi, line 15: move this to the changelog header please

roysagi (approved):
Seems good to me - just move the changelog entry please

CHANGELOG.md
ostolero: done

PR #1608, opened 2021-08-31
Split command

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40485

## Description
refactoring the existing `split-yml` command to `split` now supporting generic modules


demisto_sdk/commands/split/tests/jsonsplitter_test.py
```diff
@@ -0,0 +1,43 @@
...
+        - Running split on it.
+
+    Then
+        - Ensure dashboard is extracted to the requested location.
+        - Ensure the generic module file is edited properly in place.
+    """
+    pack = repo.create_pack('PackName')
+    pack.create_generic_module("generic-module", UNIFIED_GENERIC_MODULE)
+    generic_module_path = pack.generic_modules[0].path
+    json_splitter = JsonSplitter(input=generic_module_path, output=pack.path)
```
tomneeman151293, line 27:
```suggestion
    generic_module = pack.create_generic_module("generic-module", UNIFIED_GENERIC_MODULE)
    json_splitter = JsonSplitter(input=generic_module.path, output=pack.path)
```

demisto_sdk/commands/split/tests/jsonsplitter_test.py
```diff
@@ -0,0 +1,43 @@
...
+        res = json_splitter.split_json()
+        assert res == 0
+        assert os.path.isfile(expected_dashboard_path)
+
+        with open(expected_dashboard_path, 'r') as f:
+            result_dashboard = json.load(f)
+
+        assert result_dashboard == EXTRACTED_DASHBOARD
+
+        with open(generic_module_path, 'r') as f:
```
tomneeman151293, line 40:
```suggestion
        with open(generic_module.path, 'r') as f:
```

demisto_sdk/commands/split/jsonsplitter.py
```diff
@@ -0,0 +1,125 @@
...
+    def create_module(self):
+        if not self.new_module_file:
+            if self.logging:
+                click.echo(f"Updating module file {self.input}")
+            module_file_path = self.input
+
+        else:
+            if self.logging:
+                click.echo("Creating new module file")
+            file_name = os.path.split(self.input)[1]
```
tomneeman151293, line 106: I think it will be more clear this way
```suggestion
            file_name = os.path.basename(self.input)
```

demisto_sdk/commands/split/jsonsplitter.py
```diff
@@ -0,0 +1,125 @@
...
+            click.secho(f"Starting dashboard extraction from generic module {self.module_json_data.get('name')}",
+                        fg='bright_cyan')
+        self.create_output_dirs()
+        self.create_dashboards()
+        self.create_module()
+        return 0
+
+    def create_output_dirs(self):
+        # no output given
+        if not self.dashboard_dir:
```
tomneeman151293, line 49: I think this is a bit dangerous, prefer to have a field of self.output and check if self.output here. if someone changes in the future the way dashboard_dir or module_dir are initallized it will cause bugs

demisto_sdk/commands/split/jsonsplitter.py
```diff
@@ -0,0 +1,125 @@
...
+                try:
+                    os.mkdir(self.dashboard_dir)
+                except FileExistsError:
+                    pass
+
+                try:
+                    os.mkdir(self.module_dir)
+                except FileExistsError:
+                    pass
+
```
tomneeman151293, line 69: This can be generalized to one function of create_output_dir(dir_entity_type_name) or something like that.

demisto_sdk/commands/split/jsonsplitter.py
```diff
@@ -0,0 +1,125 @@
...
+        else:
+            if self.logging:
+                click.echo("Creating new module file")
+            file_name = os.path.split(self.input)[1]
+            module_file_path = os.path.join(self.module_dir, file_name)
+
+            if os.path.exists(module_file_path):
+                given_name = False
+                while not given_name:
+                    file_name = str(input("\nPlease enter a new file name: "))
```
tomneeman151293, line 112: Perhaps add a more clear message? that file already exists, please enter a new file name or something like that. WDYT?

tomneeman151293 (changes requested):

demisto_sdk/commands/split/jsonsplitter.py
```diff
@@ -0,0 +1,125 @@
...
+
+                try:
+                    os.mkdir(self.module_dir)
+                except FileExistsError:
+                    pass
+
+            # if not in content create the files locally
+            else:
+                if self.logging:
+                    click.echo("No output path given and not running in content repo, creating files locally")
```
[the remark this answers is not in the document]
roysagi: dont you think printing the cwd might be a bit much? maybe the same suggestion just without the os.getcwd?

demisto_sdk/commands/split/jsonsplitter.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def create_dashboards(self):
+        if self.logging:
+            click.echo("Starting dashboard creation")
+
+        for view in self.module_json_data.get('views', []):
+            for tab in view.get('tabs', []):
+                dashboard_data = tab.get('dashboard')
+
+                if dashboard_data:
```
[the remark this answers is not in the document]
roysagi: supposedly that is possible if you might not want to put a dashboard in a given tab (not sure why but it is possible as far as I know)

demisto_sdk/commands/split/jsonsplitter.py
roysagi: not sure what you mean - in the init I do get the `output` field and if it exists both the `dashboard_dir` and `module_dir` hold it.
They have to be different for the case where you'd like to split the files to different dirs - that is the better use case IMO since that is what the user would actually need to do.

demisto_sdk/commands/split/jsonsplitter.py
roysagi: Well, it does print `Creating new module file` just before this print so I think it is clear.
Come to think of it my UX thinking here is off, if you have requested to create a new file I should just ask the user for the new file name, I'm currently assuming he'd like to keep the same name and that is not necessarily the case.

demisto_sdk/commands/split/jsonsplitter.py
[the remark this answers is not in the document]
tomneeman151293: Agreed

tomneeman151293 (approved):
Great

PR #1613, opened 2021-09-01
skipping tests validation for api modules

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/40160

## Description
skipping tests validation for api modules


tomneeman151293 (approved):
What about NonSupported? Think we should handle them too? @roysagi

roysagi:
> What about NonSupported? Think we should handle them too? @roysagi

@tomneeman151293 - I am not sure as some of them do seems to have a test playbook - I think it is best to leave it for now and see if any complaints on the matter come up in the future.

tomneeman151293:
@roysagi Leave as is then

Issue #1620, opened 2021-09-03
Upload command fails behind proxy


dylan8902:
An upload to a server behind a proxy fails as no configuration is passed through to the rest client and the default is `None`

https://github.com/demisto/demisto-sdk/blob/141de2acfcc4de136fabfb01f501016e65ff7ed3/demisto_sdk/commands/upload/uploader.py#L99

dorschw:
The issue is [solved](https://github.com/demisto/demisto-py/pull/91) in the `demisto-py` package (that the demisto-SDK uses), please update it using `pip install demisto-py --upgrade`. Please let us know if there are any other issues.

dylan8902:
Brilliant thanks

PR #1621, opened 2021-09-03
Adds proxy configuration from env variables

dylan8902:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/demisto-sdk/issues/1620

## Description
Adds proxy when configuring rest client based on environment variables

## Must have
- [ ] Tests
- [ ] Documentation


tomneeman151293:
Thanks for the contribution. After viewing this PR, this support is indeed needed, but it is better to be implemented in the demisto client, to be supported by all cases using the client. We will open a dedicated issue for adding proxy support using env variable to demisto client

PR #1641, opened 2021-09-13
adding txt file support for --whitelist flag

roysagi:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/24148

## Description
You can now pass a txt file to the *--whitelist* flag in the **secrets** command.

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -566,7 +566,7 @@ def create_content_artifacts(**kwargs) -> int:
     '-wl', '--whitelist',
     default='./Tests/secrets_white_list.json',
     show_default=True,
-    help='Full path to whitelist file, file name should be "secrets_white_list.json"'
+    help='Full path to whitelist file. Can be either json or txt file.'
 )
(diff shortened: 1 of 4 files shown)


demisto_sdk/__main__.py
```diff
@@ -566,7 +566,7 @@ def create_content_artifacts(**kwargs) -> int:
     '-wl', '--whitelist',
     default='./Tests/secrets_white_list.json',
     show_default=True,
-    help='Full path to whitelist file, file name should be "secrets_white_list.json"'
+    help='Full path to whitelist file. Can be either json or txt file.'
```
tomneeman151293, line 569: I think we should add here an explanation of the expected format from txt/json file perhaps

tomneeman151293 (changes requested):
Small note about the help for the whitelist path argument.
Also need to add a unit test for the txt file

roysagi:
Closing this PR as it became stale

PR #1643, opened 2021-09-15
Validation Fix When Tried To Delete Non Existence Item In List

tomneeman151293:


tomneeman151293:
@merit-maita FYI - this happens because we have two TPB in ID set, where one is from version 5 upto 5.5, and other one is 5.5 and above. this causes the iteration to go 2 times over PB with same ID, the first delete works, but arriving to the second PB with same ID causes this error because remove fails if item does not exist in list

dansterenson (approved):

PR #1645, opened 2021-09-19
Fix `validate` crashes  - skip files, validate `pack_metadata` exists 

dorschw:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/41424
fixes: https://github.com/demisto/etc/issues/41425

## Description
- skip non-folders when calling `validate -a`
- validate packs have `pack_metadata.json` 

both prevent crashes when running `validate`.

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/pack_unique_files.py
+++ b/demisto_sdk/commands/common/hook_validations/pack_unique_files.py
@@ -53,8 +53,10 @@ INCORRECT_PACK_NAME_PATTERN = '[^a-zA-Z]pack[^a-z]|^pack$|^pack[^a-z]|[^a-zA-Z]p
 
 
 class BlockingValidationFailureException(BaseException):
-    # This should block other validations from being run.
-    # For example, raise it when a required file is missing, and there's no reason to keep validating other files.
+    """
+    Used for blocking other validations from being run, force-stopping the validation process.
+    For example, when a required file is missing. Raise it after adding a suitable error to self._errors.
+    """
     pass
 
 
--- a/demisto_sdk/commands/common/hook_validations/pack_unique_files.py
+++ b/demisto_sdk/commands/common/hook_validations/pack_unique_files.py
@@ -114,8 +114,8 @@ class PackUniqueFilesValidator(BaseValidator):
         """Returns the full file path to pack's file"""
         return os.path.join(self.pack_path, file_name)
 
-    def _is_pack_file_exists(self, file_name):
-        """Check if .secrets-ignore exists"""
+    def _is_pack_file_exists(self, file_name: str):
+        """Check if a file with given name exists in pack root"""
         if not os.path.isfile(self._get_pack_file_path(file_name)):
             if self._add_error(Errors.pack_file_does_not_exist(file_name), file_name):
                 return False
@@ -582,6 +582,7 @@ class PackUniqueFilesValidator(BaseValidator):
 
     def are_valid_files(self, id_set_validations) -> str:
         """Main Execution Method"""
+        self._is_pack_file_exists(self.pack_meta_file)
         self.validate_secrets_file()
         self.validate_pack_ignore_file()
         # We don't want to check the metadata file for this pack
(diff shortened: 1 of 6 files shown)


demisto_sdk/commands/common/hook_validations/pack_unique_files.py
```diff
@@ -582,6 +582,7 @@ def _is_price_changed(self) -> bool:
 
     def are_valid_files(self, id_set_validations) -> str:
         """Main Execution Method"""
+        self._is_pack_file_exists(self.pack_meta_file)
```
tomneeman151293, line 585: Have you tested it? it will continue to other checks anyway this way and crash, shouldn't it ?

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -296,13 +296,13 @@ def run_validation_on_all_packs(self):
             all_packs_valid.add(self.conf_json_validator.is_valid_conf_json())
 
         count = 1
-        all_packs = os.listdir(PACKS_DIR) if os.listdir(PACKS_DIR) else []
+        # Ignores local files (non-packs), such as .DS_STORE on MacOS
```
tomneeman151293, line 299:
```suggestion
        # Filter non-pack files that might exist locally (e.g, .DS_STORE on MacOS)
```

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -296,13 +296,13 @@ def run_validation_on_all_packs(self):
...
+        all_packs = list(filter(os.path.isdir, [os.path.join(PACKS_DIR, p) for p in os.listdir(PACKS_DIR)]))
```
tomneeman151293, line 300: I find list comprehension a little easier to read, but in this case it isn't a big deal, leaving this suggestion for you to decide
```suggestion
        all_packs = [os.path.join(PACKS_DIR, pack) for pack in os.listdir(PACKS_DIR) if os.path.isdir(pack)]
```

tomneeman151293 (changes requested):
Small request in addition, in the errors.py func
    def pack_file_does_not_exist(file_name):

please add an if that if file_name has pack_metadata... print an appropriate error message of
f'Pack metadata {file_name} expected to exist. Cannot proceed to further validations. Please add the pack metadata, and re-run validations.'

or something like that, to indicate to the user that there may be more problems with validations, but they cant be checked until file is supplied.

demisto_sdk/commands/validate/validate_manager.py
dorschw: here it'd be 
```
 all_packs = [os.path.join(PACKS_DIR, pack) for pack in os.listdir(PACKS_DIR) if os.path.isdir(os.path.join(PACKS_DIR, pack))]
```
which isn't elegant imo, so i'll leave it with`filter`

demisto_sdk/commands/common/hook_validations/pack_unique_files.py
dorschw: Yes. see the new mechanism. It works but I'd rather use an exception instead, to ensure validation halts (currently, it's up to the developers maintaining SDK to prevent other tests from running). wdyt?

tomneeman151293 (approved):

demisto_sdk/commands/common/hook_validations/pack_unique_files.py
```diff
@@ -52,6 +52,12 @@
                               '^A-Z]|[^A-Z]SCRIPT$|[^A-Z]Script[^a-z]|^Script$|^Script[^a-z]|[^A-Z]Script$ '
 
 
+class BlockingValidationFailureException(BaseException):
+    # This should block other validations from being run.
+    # For example, raise it when a required file is missing, and there's no reason to keep validating other files.
```
DeanArbel, line 57: use docstring, i.e. """

DeanArbel (approved):
Looks good. I've got a small question I'll ask you offline.

PR #1661, opened 2021-10-05
Fixing Azure Devops and other repositories which are not github nor gitlab

ilaner:
## Status
- [x] Ready

## Description
Sometimes `giturlparse` thinks the url is a gitlab url, even though it's `Azure Devops` URL or other provider. In this case the gitlab API url is not available and exception is being raised.
Changed to default to `demisto/content` in a case like this


tomneeman151293 (commented):
Awesome!
1. Please add to changelog the fix
2. Please verify (not sure if you already did but just making sure) i prefer you to wrap res.json() with error catch just to be even more safe. WDYT?

ilaner:
> Awesome!
> 
> 1. Please add to changelog the fix
> 2. Please verify (not sure if you already did but just making sure) i prefer you to wrap res.json() with error catch just to be even more safe. WDYT?

1. added
2. I check if the response is ok, and if it's ok it should return a correct json

tomneeman151293:
@ilaner Please assure this works for the customer. He can install specific version using 
git+https://github.com/demisto/demisto-sdk.git@<branch_name>

ilaner:
> Any update?

The customer didn't respond. I pinged him again today, and we will update about this. I think we should merge this one soon regardless.

tomneeman151293:
Customer responded and confirmed this solves his problem. Approving

tomneeman151293 (approved):

PR #1669, opened 2021-10-11
fix comparison methods

dorschw:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


TestSuite/utils.py
```diff
@@ -28,28 +31,12 @@ def is_dicts_equal(dict1: dict, dict2: dict):
...
-            return False
-
-        for item in list1:
-            if item not in list2:
-                return False
-
-        return True
+        return isinstance(list1, list) \
+            and isinstance(list2, list) \
+            and Counter(list1) == Counter(list2)  # regardless of order
```
tkatzir, line 36: Missing check for the actual values?

TestSuite/utils.py
```diff
@@ -1,55 +1,28 @@
...
-
-        if sorted(list(dict1.keys())) != sorted(list(dict2.keys())):
-            return False
-
-        for key, dict1_value in dict1.items():
-            dict2_value = dict2.get(key)
-
-            if not isinstance(dict2_value, type(dict1_value)):
-                return False
+    def is_dicts_equal(dict1: dict, dict2: dict, simple_comparison: bool):
```
tkatzir, line 3: Can `simple_comparison` have a default value?

tkatzir, line 3: I would first check whether `dict1` and `dict2` are instances of `dict`
Or will the `: dict` notation throw a runtime exception?

TestSuite/utils.py
```diff
@@ -1,55 +1,28 @@
...
+        dict1 = dict(dict1)
+        dict2 = dict(dict2)
 
-            if isinstance(dict1_value, dict):
-                if not IsEqualFunctions.is_dicts_equal(dict1_value, dict2_value):
-                    return False
+        if simple_comparison:
```
tkatzir, line 7: What does it mean? Comparing the pointers?
Is it different than below where all values are not `dict`/`list`/`set`?
If it's the same, then there's a potential for extra work...

demisto_sdk/commands/common/tests/update_id_set_test.py
```diff
@@ -674,8 +676,8 @@ class TestPlaybooks:
             "StopScheduledTask",
         ],
         "implementing_playbooks": [
-            "Calculate Severity - Standard",
```
tkatzir, line 677: Is this the edge case? Order of the items?

TestSuite/utils.py
dorschw: i'll add this test, as the notation doesn't affect runtime, thanks.

demisto_sdk/commands/common/tests/update_id_set_test.py
dorschw: yes, it tested for a subset (as set) instead of list equality.

TestSuite/utils.py
jochman: dict1==dict2 might never work. it'll work only if it is the same instance (like `dict is dict`)

demisto_sdk/commands/common/tests/dependencies_test.py
```diff
@@ -982,7 +982,7 @@ def test_dependencies(mocker, repo, test_number):
     if f'pack_{pack_to_verify}' in dependencies:
         dependencies.remove(f'pack_{pack_to_verify}')
 
-    assert IsEqualFunctions.is_lists_equal(list(dependencies), list(dependencies_from_pack_metadata))
+    assert IsEqualFunctions.is_sets_equal(dependencies, dependencies_from_pack_metadata)
```
jochman, line 985: why do we need that function? set comprehension is just `set == set`.

TestSuite/utils.py
jochman: for dict diff, we can use this tool:
https://github.com/seperman/deepdiff

it has only subdependency (ordered-dict) which is fine to add to the requirements.

jochman: also, no need to use it as Pytest magic works on dict equality: https://stackoverflow.com/questions/47201358/how-can-you-test-that-two-dictionaries-are-equal-with-pytest-in-python

demisto_sdk/commands/common/tests/dependencies_test.py
dorschw: correct, done

TestSuite/utils.py
dorschw: removed it, see updated function

dorschw: see updated function

TestSuite/utils.py
dorschw: @ilaner and I checked the variety of functions using this one, and found out the important part is the ability to compare lists as sets in specific cases (mostly when comparing a list that is created from a set), so we updated the function. Please check out the new version, it's pretty straightforward and doesn't require external tools.

jochman (approved):
You can distinguish a good fix when it's reducing the number of lines:
image

tkatzir (approved):

PR #1677, opened 2021-10-20
uncomment caplog UT in coverage-analyze command

esharf:
uncomment caplog UT in coverage-analyze command

the problem was that we need to use the `logging_setup` method before using logging actions per test (not per module).

FYI @yaakovi @tkatzir @jochman


jochman (approved):
Well done 🚀

PR #1696, opened 2021-11-04
added support for .env file

jochman:
Customer can now run several repositories using [.env file](https://pypi.org/project/python-dotenv/).

example:

.env file
```bash
DEMISTO_BASE_URL=http://127.0.0.1:8080
DEMISTO_API_KEY=XXXXXXXXXXXXXXXXXXX
```
the demisto-sdk auto loads the environment variable from cwd so the user don't need to update it manually each time he moves to another repo.

asked in https://xsoar.ideas.aha.io/ideas/FR-I-1438

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -52,6 +52,7 @@ PyPDF2 = "*"
 PyYAML = "*"
 giturlparse = "*"
 pytest-freezegun = "*"
+python-dotenv = "*"
 
 [dev-packages]
 mitmproxy = "==5.1.0"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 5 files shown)


jochman:
FYI @bakatzir

tomneeman151293:
Need to think what should be the order of precedence. Probably .env should be a fallback if the BASE_URL/API_KEY aren't exported. Think of when we work on multiple servers and we want to upload easily from content. We shouldn't be entering the .env file everytime to edit it

jochman:
> Need to think what should be the order of precedence. Probably .env should be a fallback if the BASE_URL/API_KEY aren't exported. Think of when we work on multiple servers and we want to upload easily from content. We shouldn't be entering the .env file everytime to edit it

The other way. If there's a .env file you should always use it.
local is stronger than global.

tomneeman151293 (approved):

PR #1699, opened 2021-11-08
SDK - read dependencies from meta_data in find-dependency command

ilappe:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/42348

### DIFF
--- a/demisto_sdk/commands/find_dependencies/find_dependencies.py
+++ b/demisto_sdk/commands/find_dependencies/find_dependencies.py
@@ -1471,7 +1471,7 @@ class PackDependencies:
             pack_name: str,
             id_set_path: str = '',
             exclude_ignored_dependencies: bool = True,
-            update_pack_metadata: bool = True,
+            update_pack_metadata: bool = False,
             silent_mode: bool = False,
             verbose: bool = False,
             debug_file_path: str = '',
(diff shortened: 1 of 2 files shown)


Pipfile
```diff
@@ -30,7 +30,7 @@ mergedeep = "*"
 mypy = "==0.782"
 networkx = "*"
 nltk = "*"
-pandas = "*"
+pandas = "<1.3.4"
```
yaakovi, line 33: why?

demisto_sdk/commands/find_dependencies/find_dependencies.py
```diff
@@ -1496,14 +1496,16 @@ def find_dependencies(
...
 
         """
-        if not id_set_path or not os.path.isfile(id_set_path):
-            if not skip_id_set_creation:
-                id_set = IDSetCreator(print_logs=False).create_id_set()
-            else:
-                return {}
-        else:
+
+        if id_set_path and os.path.isfile(id_set_path):
```
yaakovi, line 1500: nice

Pipfile
ilappe: Because pandas new version are depended some lib that required python 3.10 at minimum
i discussed it with @jochman and this should be the solution
may be should be in separate PR?

jochman: it should be in a separate pr

yaakovi (approved):
move the Pipfile and lock changes to another PR.

ilappe:
> move the Pipfile and lock changes to another PR.

done in https://github.com/demisto/demisto-sdk/pull/1706

PR #1703, opened 2021-11-10
Verify associated tasks

PA-Rweins:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/34343

## Description
- Added validation to playbooks and test playbooks which checks that all the playbook's scripts have a valid brand name
- A valid brand is a brand which is non-empty and also appears as an entry in the integration list of the id_set
- Created an error code for the new error

## Must have
- [x] Tests
- [x] Documentation


idovandijk:
Can you please elaborate on this check?
Is it checking that every command in a task which has a brand name in parenthesis is a brand that exists in content?

PA-Rweins:
> Can you please elaborate on this check? Is it checking that every command in a task which has a brand name in parenthesis is a brand that exists in content?

Yes, exactly what you wrote, and also if it doesn't have a brand name, for example "|||some-command", it will also fail the validation.

So it goes over each task and if that task has a script entry, it will check that the brand is valid.
A valid brand means it's not empty and that there's an existing integration with this name.

idovandijk:
What about things like this:
image


image

ChanochShayner:
@PA-Rweins @idovandijk This should be just For TestPB that test a specific integration/use-case,
In PB we want all the sources to bring information (for example All !ip for malicious IP, and not just from VT).

idovandijk:
> @PA-Rweins @idovandijk This should be just For TestPB that test a specific integration/use-case, In PB we want all the sources to bring information (for example All `!ip` for malicious IP, and not just from VT).

In specific integration tests - yes.
When testing use-cases - we may use general commands that don't call a specific brand, and also we use scripts which of course don't have a brand. Please confirm whether this behavior will still work normally so I can approve

idovandijk:
Updating with what was discussed with @PA-Rweins and @ChanochShayner:
Currently many test-playbooks are failing with this validation. Next steps:
1. Add a check for `Builtin|||` and don't fail it (because it's valid)/
2. Find which testplaybooks are for an **integration** by checking if the pack contains an integration.
3. Update the commands with ||| to use the brand of that integration.
4. Find which testplaybooks are for a **use-case** or a **script**.  Not many results should be found here.
5. Determine the correct solution individually.

ChanochShayner:
> > Updating with what was discussed with @PA-Rweins and @ChanochShayner: Currently many test-playbooks are failing with this validation. Next steps:
> > 
> > 1. Add a check for `Builtin|||` and don't fail it (because it's valid)/
> > 2. Find which testplaybooks are for an **integration** by checking if the pack contains an integration.
> > 3. Update the commands with ||| to use the brand of that integration.
> > 4. Find which testplaybooks are for a **use-case** or a **script**.  Not many results should be found here.
> > 5. Determine the correct solution individually.
> 
> @idovandijk @ChanochShayner Just talked with @Bargenish about this. We decided that since those are tpbs it's better to just ignore them all, so I'm working on creating a script that will ignore all the problematic tpbs

Please fix at least the reputations command -
IP, Domain, URL, File
That's why I opened this issue in the first place.
In addition, a correction to this validation should be added in the format command.

Bargenish:
@ChanochShayner please open a new issue for the format part.
we're starting bugs shift and it can be handled separately.

Let's discuss the fix playbooks part.

PR #1708, opened 2021-11-15
Handled ignorable files

PA-Rweins:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/43694)

## Description
Some files that were supposed to be ignored were not ignored. 
In order to ignore them I created a list of those files and created a function where I check for ignorable files so they wont be validated.

## Must have
- [x] Tests
- [ ] Documentation


Bargenish (changes requested):
Let's discuss the tests offline.

tomneeman151293:
@PA-Rweins Please add a test before your changes, showing that the test fails with same error msg in bug, and after that see that test passes after your changes.

demisto_sdk/commands/validate/tests/validators_test.py
```diff
@@ -1124,31 +1124,33 @@ def test_ignore_test_doc_non_pack_file_should_not_ignore(self, file_path: str):
...
             - File is ignored and True is returned
         """
         validate_manager = ValidateManager(check_is_unskipped=False)
-        assert validate_manager.ignore_test_doc_non_pack_file(file_path)
+        assert validate_manager.ignore_files_irrelevant_for_validation(file_path)
 
     @pytest.mark.parametrize('file_path',
                              ['OtherDir/Integration/file.json',
                               'TestData/file.json',
                               'TestPlaybooks/file.yml',
```
yaakovi, line 1151: in general, this folder shouldn't be ignored but let's leave it as is (better to deprecate the tests in it).

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1404,9 +1404,10 @@ def format_file_path(self, file_path, old_path, old_format_files):
         else:
             return file_path
 
-    def ignore_test_doc_non_pack_file(self, file_path: str) -> bool:
+    def ignore_files_irrelevant_for_validation(self, file_path: str) -> bool:
         # ignore doc data, test_data and non pack files
-        if any(test_dir in str(file_path) for test_dir in TESTS_AND_DOC_DIRECTORIES) or PACKS_DIR not in file_path:
+        if any(name in str(file_path) for name in VALIDATION_USING_GIT_IGNORABLE_DATA) \
+                or PACKS_DIR not in file_path:
```
yaakovi, line 1410: I would replace the order of the conditions (it is better performance-wise)

PA-Rweins: Fixed

demisto_sdk/commands/validate/tests/validators_test.py
PA-Rweins: I'm not sure which folder you're talking about

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1392,7 +1392,7 @@ def format_file_path(self, file_path, old_path, old_format_files):
                 file_path = file_path.replace('.py', '.yml').replace('.ps1', '.yml').replace('.js', '.yml')
 
                 if old_path:
-                    old_path = old_path.replace('.py', '.yml').replace('.ps1', ',yml').replace('.js', '.yml')
+                    old_path = old_path.replace('.py', '.yml').replace('.ps1', '.yml').replace('.js', '.yml')
```
Bargenish, line 1395: Good catch!

demisto_sdk/commands/common/constants.py
```diff
@@ -858,6 +858,19 @@ class FileType(Enum):
     'doc_imgs',
 ]
 
+VALIDATION_USING_GIT_IGNORABLE_DATA = (
+    'Pipfile',
+    'Pipfile.lock',
+    'command_examples',
```
Bargenish, line 864: Im pretty worried about  this one. we do not enforce the name of the example file and we have various names on the repo.
see the following for examples:
Packs/SentinelOne/Integrations/SentinelOne-V2/examples.txt
Packs/BitSight/Integrations/BitSightForSecurityPerformanceManagement/command-examples.txt

Bargenish (changes requested):
great job with the testing

Bargenish (approved):

yaakovi (approved):

PR #1717, opened 2021-11-21
add importlib-resources

dorschw:
## Related Issues
fixes: https://github.com/demisto/etc/issues/43962

## Description
Calling `demisto-sdk download` fails when using python <3.9 **and** `importlib-resources` is missing


jochman (changes requested):
Generate Pipfile.lock (`pipenv lock`) and add it to the pr.

See if the install works on python 3.7+

jochman (approved):

PR #1718, opened 2021-11-22
unused constants

dorschw:
## Status
- [x] In Progress
- [ ] Ready

## Description
Removing unused constants


tomneeman151293:
Just a small note: As content (and i think private + gold might too, need to check) imports the whole sdk project, it might be that the constants are used in content repo but not in SDK repo and deleting them can cause issues. so need to double check on our other repo's as well that they aren't being used.

PR #1733, opened 2021-12-01
Fix get_playground_id - filter the playground for the current user

ilaner:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/44272

## Description
When getting the playground id, we search investigations. We could be the admin user with several users. In this case we will get several playgrounds, but we are only interested on the current user playground.
Here we filter the playgrounds to the current user.

## Must have
- [ ] Tests
- [ ] Documentation


demisto_sdk/commands/run_cmd/runner.py
```diff
@@ -91,14 +92,21 @@ def _get_playground_id(self):
...
-        if answer.total != 1:
+        user_data, response, _ = self.client.generic_request(path='/user',
+                                                             method='GET',
+                                                             content_type='application/json',
+                                                             response_type=object)
+        if response != 200:
+            raise RuntimeError('Cannot find username')
+        username = user_data.get('username')
+        playgrounds = [playground for playground in answer.data if playground.creating_user_id == username]
+        if len(playgrounds) != 1:
```
DeanArbel, line 103: Let's put this flow after the `len(plagrounds) != 1`. If there's only one playground, there's no need to do the user filter.

DeanArbel (changes requested):
Nicely done, see my comment.

DeanArbel (approved):

PR #1735, opened 2021-12-02
Merge id_set in find_dependencies will not fail on duplicates

ilaner:
## Status
- [x] Ready
## Related Issues
fixes: https://github.com/demisto/etc/issues/43460

## Description
In validate, when finding dependencies, we merge the local id_sets with content id_sets. If there are duplicates, i.e the id_sets were already merged, the build fails.
In this fix we use the old id_sets in this case


DeanArbel (approved):
Good job

PR #1737, opened 2021-12-05
Add fix cj105

JasBeilin:
## Status
- [x] In Progress
- [x] Ready

## Related Issues
fixes: (https://github.com/demisto/etc/issues/44306)

## Description
Added a check so the validation only fails if a `<integration_yml_path>_test.py` file is not found. 

## Must have
- [ ] Tests


demisto_sdk/commands/common/hook_validations/conf_json.py
```diff
@@ -144,3 +145,18 @@ def integration_has_unskipped_test_playbook(self, integration_data, integration_
                     test_playbook_ids.append(test['playbookID'])
 
         return self.has_unskipped_test_playbook(integration_data, integration_id, file_path, test_playbook_ids)
+
+    def get_test_path(self, file_path):
+        test_path = Path(file_path)
+        test_file_name = test_path.parts[-1].replace('.yml', '_test.py')
```
DeanArbel, line 151: why is the function being called with `.yml` instead of `_test.py`>

DeanArbel (approved):

demisto_sdk/commands/common/hook_validations/conf_json.py
JasBeilin: Because this is the function that creates the test_path from the yml_path.

PR #1756, opened 2021-12-16
Update id according to old file in format

ilaner:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/43893

## Description
When running a format on a YML file, the `id` value should be changed to the `name` value. However, we don't change existing YMLs, because we have legacy YMLs which the `id` value is not equal to the `name`, but we don't set the `id` as the old `id`. 
This change will keep the existing `id` when running `format`

## Must have
- [x] Tests
- [x] Documentation

### DIFF
--- a/demisto_sdk/commands/format/update_generic_yml.py
+++ b/demisto_sdk/commands/format/update_generic_yml.py
@@ -86,11 +86,14 @@ class BaseUpdateYML(BaseUpdate):
                 updated_integration_id[self.id_and_version_location['id']] = self.data['name']
             self.id_and_version_location['id'] = self.data['name']
         else:
-            modified_yml_id = self.get_id_and_version_for_data(self.old_file).get('id')
-            modified_yml_id_from_master = self.get_id_and_version_for_data(self.old_file)['id']
-            if modified_yml_id_from_master != modified_yml_id:
-                click.secho(f'The modified YML file corresponding to the path: {self.path} ID does not match the ID in remote YML file. Changing the YML ID from {modified_yml_file} back to {modified_yml_id_from_master}.'
-                self.id_and_version_location['id'] = modified_yml_id_from_master
+            if self.verbose:
+                click.echo('It is a modified file, keeping the old ID')
+            current_id = self.id_and_version_location.get('id')
+            old_id = self.get_id_and_version_for_data(self.old_file).get('id')
+            if current_id != old_id:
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/format/update_generic_yml.py
```diff
@@ -82,6 +85,10 @@ def update_id_to_equal_name(self) -> None:
             if is_uuid(self.id_and_version_location['id']):
                 updated_integration_id[self.id_and_version_location['id']] = self.data['name']
             self.id_and_version_location['id'] = self.data['name']
+        else:
+            if self.verbose:
+                click.echo('It is a modified file, keeping the old ID')
+            self.id_and_version_location['id'] = self.get_id_and_version_for_data(self.old_file)['id']
```
tomneeman151293, line 91: I think we can add a message if ID in master does not equal to current ID, indicating we're changing it
```suggestion
            modified_yml_id = self.get_id_and_version_for_data(self.old_file).get('id')
            modified_yml_id_from_master = self.get_id_and_version_for_data(self.old_file)['id']
            if modified_yml_id_from_master != modified_yml_id:
                click.secho(f'The modified YML file corresponding to the path: {self.path} ID does not match the ID in remote YML file. Changing the YML ID from {modified_yml_file} back to {modified_yml_id_from_master}.'
                self.id_and_version_location['id'] = modified_yml_id_from_master
```

tomneeman151293 (changes requested):
Excellent. Add UTs please

ilaner:
> Excellent. Add UTs please

Added UTs

demisto_sdk/commands/format/update_generic_yml.py
```diff
@@ -81,6 +84,15 @@ def update_id_to_equal_name(self) -> None:
             if is_uuid(self.id_and_version_location['id']):
                 updated_integration_id[self.id_and_version_location['id']] = self.data['name']
             self.id_and_version_location['id'] = self.data['name']
+        else:
+            if self.verbose:
+                click.echo('It is a modified file, keeping the old ID')
```
DeanArbel, line 89: I don't see the purpose of this message as we're already printing in case of a modification.

DeanArbel (approved):
Good job!

tomneeman151293 (approved):
Great tests! Please before merge
1) run demisto-sdk format on modified/new file, and see result in your eyes that indeed the expected happens

PR #1757, opened 2021-12-19
replace ApiModule code in YmlSplitter back to import

ilaner:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/44714

## Description
When unifying YML, the API module code content is copied to the script. However, when splitting the YML filed. This code is not removed.
This PR removes the copied API module, and replaces the original `import` code.

## Must have
- [x] Tests
- [ ] Documentation


DeanArbel (approved):
Well done.
One note - no need to have the full EDL integration code in the sdk. Let's replace it with a mock integration instead, which only has an import APIModule line.

PR #1786, opened 2022-01-02
Validate deleted files

MLainer1:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/44703)

## Description
Adding validation for deleted integrations files.
Currently we allow to delete: 
1. 'testplaybook'
2. 'doc_image'
3. 'whitelist'

- We throw an error in case the validation finds a not allowed file that was deleted.
- It is not ignorable, in the cases that files has to be deleted, we'll force-merge.
- We get the deleted files from git, so if git deems a file as deleted, we treat it as deleted.

## Must have
- [x] Tests
- [ ] Documentation


tomneeman151293 (changes requested):
Awesome! Important note:
1) Some files can! ben deleted
- Test PBs
- Maybe more files
I suggest making a test for each type of content entity, and see that for each content entity, the validate on deleted files returns True/False if it can or cannot be deleted.

The issue itself talked only about integrations/scripts but I think we should expand it.
- add a NOT_ALLOWED_DELETE_FILETYPES of file types who cannot be deleted
- Only return error for every file whom file type is in NOT_ALLOWED_DELETE_FILETYPES
Also
- You should add a test for every deleted file if its file type is in NOT_ALLOWED_DELETE_FILETYPES

dorschw:
@MLainer1 what's our status on this?

idovandijk:
Can you please add what this validation does?

idovandijk:
@MLainer1 I'm not 100% certain about the logic for Git determining that a file is deleted. I do think that it has false-positives.
For example: playbooks goes through a major rework (lots of tasks renamed, moved, replaced), or a large enhancement to a playbook which can introduce a very different readme file.
OR - when a playbook is renamed in favor of a new playbook which replaces the old file name.
For example: `Phishing.yml` becomes `Phishing_6_0.yml` and then a new playbook takes the name of `Phishing.yml`. This happens from time to time and we don't want to have to force merge those

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1138,6 +1142,33 @@ def should_raise_pack_version(pack: str) -> bool:
...
+        Args:
+            file_path: The file path.
+
+        Returns: True if the file allowed to be deleted, else False.
+
+        """
+        file_types = [e for e in FileType]
+        file_types_not_allowed_to_be_deleted = [file_type for file_type in file_types if file_type not in FileType_ALLOWED_TO_DELETE]
+        file_type = find_type(file_path)
+        return file_type not in file_types_not_allowed_to_be_deleted
```
dantavori, line 1157: more simple:
```suggestion
        file_type = find_type(file_path)
        return file_type in FileType_ALLOWED_TO_DELETE or not file_type
```

demisto_sdk/tests/integration_tests/validate_integration_test.py
```diff
@@ -642,6 +642,9 @@ def test_invalid_modified_bc_deprecated_integration(self, mocker, repo):
         mocker.patch.object(GitUtil, '__init__', return_value=None)
         mocker.patch.object(GitUtil, 'get_current_working_branch', return_value='MyBranch')
 
+        mocker.patch.object(ValidateManager, 'validate_deleted_files', return_value=True)
```
dantavori, line 645: probably don't need this mock

dantavori (changes requested):
also make the change to address the edge case where git recognizes renamed files as deleted+added

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Improved the detection of errors in **doc-review** command.
 * The **validate** command now checks if a readme file is empty, only for packs that contain playbooks or were written by a partner
 * The **validate** command now makes sure common contextPath values (e.g. `DBotScore.Score`) have a non-empty description, and **format** populates them automatically.
+* Added to **validate** a validation that prevents the user to delete importent files.
```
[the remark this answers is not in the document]
dorschw:
```suggestion
* Added a **validate** check to prevent deletion of files whose deletion is not supported by the XSOAR marketplace.
```

dantavori (approved):

PR #1758, opened 2021-12-19
primary branch can be main

tkatzir:
## Status
- [ ] In Progress
- [x] Ready
- [ ] In Hold - (Reason for hold)

## Related Issues
fixes: https://github.com/demisto/etc/issues/39616

PR #1807, opened 2022-01-10
Revert "primary branch can be main"

yaakovi:
Reverts demisto/demisto-sdk#1758


dorschw (approved):

dorschw (approved):
is necessary after all. 
@tkatzir will handle the fix before re-applying this feature.

PR #1810, opened 2022-01-10
Fix an issue with list validation

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/45179

## Description
Fixed an issue where validation failed with no error message for lists (when no `fromVersion` field was found).

## Screenshots
Before:

image

After: 

image

## Must have
- [x] Tests


demisto_sdk/commands/common/hook_validations/lists.py
```diff
@@ -52,4 +51,8 @@ def is_valid_from_server_version(self) -> bool:
                                      file_path=None):
                     return False
             return True
-        return False
+        else:
```
yaakovi, line 54: no need for the `else` statement after a return

please also fix the call to handle_error in line 50 (why `file_path` is None?)

demisto_sdk/commands/common/tests/lists_test.py
```diff
@@ -14,6 +14,7 @@
 LIST_GOOD = TEST_CONTENT_REPO / PACKS_DIR / 'Sample01' / LISTS_DIR / 'list-checked_integrations.json'
 LIST_BAD_FROM_VERSION = TEST_CONTENT_REPO / PACKS_DIR / 'Sample01' / LISTS_DIR / 'bad_from_version.json'
 LIST_BAD_VERSION = TEST_CONTENT_REPO / PACKS_DIR / 'Sample01' / LISTS_DIR / 'bad_version.json'
+LIST_BAD_FORMAT = TEST_CONTENT_REPO / PACKS_DIR / 'Sample01' / LISTS_DIR / 'list-bad_format.json'
```
yaakovi, line 17: do not mix the test files of the `content` objects with the validate command

demisto_sdk/commands/common/hook_validations/lists.py
Shellyber: Didn't touch this code (handle_error) in that line, will change it

yaakovi (approved):

PR #1811, opened 2022-01-10
Update release_guide.md

dorschw:
Clearer release guide


dorschw:
will update and merge once for @ilaner 's update of the scripts is applied

PR #1819, opened 2022-01-11
fixed deps, send them loose

jochman:
send deps loose, will resolve locking.

also, created a requirements.txt file from pipenv that will help installing demisto-sdk and running tox locally.


dorschw:
handled in https://github.com/demisto/demisto-sdk/pull/1895

PR #1820, opened 2022-01-11
Loading and dumping  all YAML files with XSOAR_YAML, which uses ruamel.yaml

ilaner:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/45171

## Description
Inconsistency between `pyyaml` and `ruamel.yaml` cause a lot of problems.

## Must have
- [ ] Tests
- [ ] Documentation

@barchen1 FYI


DeanArbel (approved):
@ilaner What are the implications of the switch to ruamel?

dorschw:
wdyt about calling it `yaml_wrapper` or `yaml_util`, rather than `xsoar_yaml`? I think it would be clearer

ilaner:
> @ilaner What are the implications of the switch to ruamel?

It should solve inconsistency problems like in https://github.com/demisto/etc/issues/45652, `ruamel` is better, and suppose to support most of the `pyyaml` features.
If we have to use `pyyaml` for some reason, we can add a `use_pyyaml` argument to the `XSOAR_YAML` object.

demisto_sdk/commands/common/xsoar_yaml.py
```diff
@@ -3,29 +3,29 @@
 
 class XSOAR_YAML:
     def __init__(self, preserve_quotes=True, allow_duplicate_keys=True, default_flow_style=None, width=None, typ=None):
-        self.xsoar_yaml = YAML()
-        self.xsoar_yaml.preserve_quotes = preserve_quotes
-        self.xsoar_yaml.allow_duplicate_keys = allow_duplicate_keys
-        self.xsoar_yaml.default_flow_style = default_flow_style
-        self.xsoar_yaml.width = width
-        self.xsoar_yaml.typ = typ
+        self.ryaml = YAML()
```
dorschw: Perhaps name it `_ryaml`, discouraging anyone from accessing `xsoar_yaml.ryaml.command_or_attribute`, and allowing easier maintenance?

This will make customizing things much harder -- for better or worse.
wdyt? @ilaner @jochman

demisto_sdk/commands/common/xsoar_yaml.py
```diff
@@ -0,0 +1,27 @@
...
+    def _order_dict(data):
+        return {k: XSOAR_YAML._order_dict(v) if isinstance(v, dict) else v
+                for k, v in sorted(data.items())}
+
+    def load(self, stream):
+        return self._ryaml.load(stream)
+
+    def dump(self, data, stream=None, sort_keys=False, **kwargs):
+        if not stream:
+            from yaml import dump
```
jochman, line 23: we want to remove the use of pyyaml. Do you have a better solution that does not includes pyyaml?

demisto_sdk/commands/common/xsoar_yaml.py
```diff
@@ -0,0 +1,27 @@
+from ruamel.yaml import YAML
```
jochman, line 1: add docstring/stubs.

jochman (approved):
Generally fine (except the failing 47 uts).
Added some comments on the XSOAR_YAML object. Thanks!

demisto_sdk/commands/common/xsoar_yaml.py
```diff
@@ -0,0 +1,49 @@
+from io import StringIO
+
+from ruamel.yaml import YAML
+
+
+class XSOAR_YAML:
```
jochman, line 6: create a abstract class (interface) with dump/s and load/s

jochman (changes requested):

jochman (approved):
👍

PR #1821, opened 2022-01-11
End to end test - instance's classifier/ mapper

JasBeilin:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/38628)

## Description
Changed Classifier/ Mapper to be given in the conf.json in content instead of test-conf.
*Can be only tested live while build is running, as the classifier and mapper configured are erased from the instance once it is deactivated. *

## Must have
- [x] Tests


DeanArbel (commented):
Very nice job. What about tests?

DeanArbel (approved):
Approved.

JasBeilin:
> Very nice job. What about tests?

Writing here as well, tests are already in master as this was already merged, the source of configuration was changed and tests are irrelevant for that.

PR #1836, opened 2022-01-19
Default get remote file from git instead of API

ilaner:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/45718

## Description
Handle cases remote url recognized as gitlab incorrectly
First try to reach out gitlab to get the project id from there, but if failing treating it as github project if the remote url is parsed and `owner` and `name` of the repo exists.

## Must have
- [ ] Tests
- [ ] Documentation


demisto_sdk/commands/common/GitContentConfig.py
```diff
@@ -32,7 +32,7 @@ class GitContentConfig:
         CONTENT_GITHUB_MASTER_LINK: Link to the content git repository's master branch
         Credentials: Credentials to the git.
     """
-    BASE_RAW_GITHUB_LINK = r'https://raw.githubusercontent.com/'
+    BASE_RAW_GITHUB_LINK = r'https://raw.{}.com/'
```
dorschw, line 35: note this does not support custom domains such as `gitfoo.local`

ilaner: this is a typo, the `com` is there by mistake. fixed

demisto_sdk/commands/common/GitContentConfig.py
```diff
@@ -12,60 +18,77 @@
...
 
-    def __init__(self, repo_name: Optional[str] = None):
-        self.Credentials = GitCredentials()
-        if not repo_name:
+    ENV_REPO_HOSTNAME_NAME = 'DEMISTO_SDK_REPO_HOSTNAME'
+
+    def __init__(self, repo_name: Optional[str] = None, git_provider: Optional[GitProvider] = None, repo_hostname: Optional[str] = None):
+        self.credentials = GitCredentials()
+        self.repo_hostname = repo_hostname or os.getenv(GitContentConfig.ENV_REPO_HOSTNAME_NAME)
+        if not git_provider:
```
jochman, line 55: Why not just give it as a default?

demisto_sdk/commands/common/GitContentConfig.py
```diff
@@ -12,60 +18,77 @@
...
-    GITLAB_HOST: Optional[str] = None
-    GITLAB_ID: Optional[int] = None
     BASE_RAW_GITLAB_LINK = "https://{GITLAB_HOST}/api/v4/projects/{GITLAB_ID}/repository"
 
-    def __init__(self, repo_name: Optional[str] = None):
-        self.Credentials = GitCredentials()
-        if not repo_name:
+    ENV_REPO_HOSTNAME_NAME = 'DEMISTO_SDK_REPO_HOSTNAME'
+
+    def __init__(self, repo_name: Optional[str] = None, git_provider: Optional[GitProvider] = None, repo_hostname: Optional[str] = None):
```
jochman, line 52:
```suggestion
    def __init__(self, repo_name: Optional[str] = None, git_provider: GitProvider = GitProvider.github, repo_hostname: Optional[str] = None):
```
Why not just give it as a default?
Then you can delete the fallback to github

demisto_sdk/commands/common/GitContentConfig.py
```diff
@@ -12,60 +18,77 @@
...
+        self.credentials = GitCredentials()
+        self.repo_hostname = repo_hostname or os.getenv(GitContentConfig.ENV_REPO_HOSTNAME_NAME)
+        if not git_provider:
+            git_provider = GitProvider.GitHub
+        self.git_provider = git_provider
+        if not self.repo_hostname:
+            self.repo_hostname = GitContentConfig.GITHUB_USER_CONTENT if git_provider == GitProvider.GitHub else "gitlab.com"
```
jochman, line 59: What about hosted gitlab? You even got it in the environment varibles

demisto_sdk/commands/common/GitContentConfig.py
```diff
@@ -12,60 +18,77 @@
...
+            self.base_api = GitContentConfig.BASE_RAW_GITLAB_LINK.format(GITLAB_HOST=self.repo_hostname,
+                                                                         GITLAB_ID=self.gitlab_id)
 
-    def _get_repository_properties(self, urls: Iterable) -> Optional[giturlparse.result.GitUrlParsed]:
+    @staticmethod
+    def _get_repository_properties() -> Optional[giturlparse.result.GitUrlParsed]:
         """Returns the git repository of the cwd.
         if not running in a git repository, will return an empty string
         """
+        urls = list(GitUtil().repo.remote().urls)
```
jochman, line 91: it is an iterator, why we need it to be a list?

demisto_sdk/commands/common/GitContentConfig.py
```diff
@@ -76,43 +99,57 @@ def _set_repo_config(self, parsed_git: Optional[giturlparse.result.GitUrlParsed]
...
-                click.secho('Could not find repository id on gitlab - defaulting to demisto/content', fg='yellow')
-                self.CURRENT_REPOSITORY = GitContentConfig.OFFICIAL_CONTENT_REPO_NAME
-                return
-            self.GITLAB_HOST = gitlab_host
-            self.GITLAB_ID = gitlab_id
-            self.CURRENT_REPOSITORY = parsed_git.repo
-            self.BASE_RAW_GITLAB_LINK = self.BASE_RAW_GITLAB_LINK.format(GITLAB_HOST=gitlab_host,
-                                                                         GITLAB_ID=gitlab_id)
+        hostname = parsed_git.host
+        if '@' in hostname:
```
jochman, line 106: as you did below, add a comment on this edge-case.

demisto_sdk/commands/common/tools.py
```diff
@@ -208,117 +208,143 @@ def get_core_pack_list() -> list:
...
+        file_details = ryaml.load(file_content)
     # if neither yml nor json then probably a CHANGELOG or README file.
     else:
-        details = {}
-    return details
+        file_details = {}
+    return file_details
+
+
+# @lru_cache(maxsize=64)
```
jochman, line 317: can we enable cache?

jochman (changes requested):
I am really impressed. Those changed may take awhile to review.

Please run any CI flow that we have.

Add those changes to the changelog and the README (we might want to create an environment variable sections).

Are those changes affecting the performance of the build or running commands?

demisto_sdk/commands/common/GitContentConfig.py
ilaner: No real reason, will change.

demisto_sdk/commands/common/GitContentConfig.py
ilaner: That's the case that the user didn't provide the `repo_hostname` (not with the argument and not with env variable), but did pass git_provider as gitlab.
The default gitlab hostname is `gitlab.com`.

demisto_sdk/commands/common/GitContentConfig.py
ilaner: No real reason, those are leftovers from old code. Will change.

demisto_sdk/commands/common/tools.py
ilaner: I don't see why we can't do that. That's line was commented before, will investigate why.

ilaner:
> I am really impressed. Those changed may take awhile to review.
> 
> Please run any CI flow that we have.
> 
> Add those changes to the changelog and the README (we might want to create an environment variable sections).
> 
> Are those changes affecting the performance of the build or running commands?

I think this should improve the performance.
Before the change - we got the remote file from the API, and if we cannot get it we did `git show origin/{tag}:{file_path}`.
Now we do the opposite, which may be more efficient.

The main improvement will be that it should support all git providers, and support cases in which the user for some reason changed his remote URL.

jochman:
> > I am really impressed. Those changed may take awhile to review.
> > Please run any CI flow that we have.
> > Add those changes to the changelog and the README (we might want to create an environment variable sections).
> > Are those changes affecting the performance of the build or running commands?
> 
> I think this should improve the performance. Before the change - we got the remote file from the API, and if we cannot get it we did `git show origin/{tag}:{file_path}`. Now we do the opposite, which may be more efficient.
> 
> The main improvement will be that it should support all git providers, and support cases in which the user for some reason changed his remote URL.

Why don't we check it with the local master? (not origin). That **might** be faster but may cause some bc checks to not run.

ilaner:
> > > I am really impressed. Those changed may take awhile to review.
> > > Please run any CI flow that we have.
> > > Add those changes to the changelog and the README (we might want to create an environment variable sections).
> > > Are those changes affecting the performance of the build or running commands?
> > 
> > 
> > I think this should improve the performance. Before the change - we got the remote file from the API, and if we cannot get it we did `git show origin/{tag}:{file_path}`. Now we do the opposite, which may be more efficient.
> > The main improvement will be that it should support all git providers, and support cases in which the user for some reason changed his remote URL.
> 
> Why don't we check it with the local master? (not origin). That **might** be faster but may cause some bc checks to not run.

I don't think it's a good idea. It might be faster, but the function tries to get the remote file in origin. If the local branch is not fetched, or is not up-to-date or doesn't exist at all locally (for instance, we run this function when `tag` is `contrib/<pr name>` which probably is not available).

demisto_sdk/commands/common/tools.py
```diff
@@ -208,117 +209,143 @@ def get_core_pack_list() -> list:
...
+def get_file_details(
+        file_content,
+        full_file_path: str,
+        return_content: bool = False,
+) -> Union[str, Dict]:
     if return_content:
-        return file_content
-    if full_file_path.endswith('json'):
-        details = res.json() if res.ok else json.loads(local_content)
+        file_details = file_content
```
jochman, line 307: If you don't want to parse, don't parse :|
How much effort is to change it?

yaakovi (dismissed):
why doesn't `demisto_sdk/commands/common/GitContentConfig.py` follow our file name conventions? should be `snake_case`

demisto_sdk/commands/common/tools.py
```diff
@@ -188,117 +190,143 @@ def get_core_pack_list() -> list:
...
-        tag: The branch name. default is 'master'
-        return_content: Determines whether to return the file's raw content or the dict representation of it.
-        suppress_print: whether to suppress the warning message in case the file was not found.
-        github_repo: The repository to grab the file from
-    Returns:
-        The file content in the required format.
+    repo = git.Repo(search_parent_directories=True)  # the full file path could be a git file path
+    repo_git_util = GitUtil(repo)
+    git_path = repo_git_util.get_local_remote_file_path(full_file_path, tag)
+    file_content = repo_git_util.get_local_remote_file_content(git_path)
```
jochman, line 209: Move exception to here.

demisto_sdk/commands/common/git_content_config.py
```diff
@@ -0,0 +1,244 @@
...
+            git_provider: The git provider to use (e.g GitProvider.GitHub, GitProvider.GitLab)
+            repo_hostname: The hostname to use (e.g "code.pan.run", "gitlab.com", "my-hostename.com")
+            project_id: The project id, relevant for gitlab.
+        """
+        self.current_repository = repo_name if repo_name else None
+        if project_id:
+            git_provider = GitProvider.GitLab
+
+        self.credentials = GitCredentials()
+        parsed_hostname = urlparse(repo_hostname).hostname
```
jochman, line 73: what about just "hostname"? why do I care if it parsed?

demisto_sdk/commands/common/git_content_config.py
```diff
@@ -0,0 +1,244 @@
...
+        self.repo_hostname = parsed_hostname or repo_hostname or os.getenv(GitContentConfig.ENV_REPO_HOSTNAME_NAME)
```
jochman, line 74: Don't we want that user-defined variables will be stronger than resolved ones?

demisto_sdk/commands/common/git_content_config.py
```diff
@@ -0,0 +1,244 @@
...
+            repo_name: repository name in this structure: "<org_name>/<repo_name>".
+
+        Returns:
+            If found -  a tuple of the github hostname and the repo name that was found.
+            If not found - 'None`
+        """
+        if not github_hostname or not repo_name:
+            return None
+        api_host = github_hostname if github_hostname != GitContentConfig.GITHUB_USER_CONTENT else 'github.com'
+        if api_host == 'github.com':
```
jochman, line 174: apply `.lower()` to `api_host`. You know the rules and so do I.

demisto_sdk/commands/common/tests/tools_test.py
```diff
@@ -55,6 +57,8 @@
 from TestSuite.repo import Repo
 from TestSuite.test_tools import ChangeCWD
 
+GIT_ROOT = "{}".format(git_path())
```
jochman, line 60:
```suggestion
GIT_ROOT = git_path()
```

jochman (approved):
k

PR #1844, opened 2022-01-25
Create CODEOWNERS

anara123:
## Description
Add @idovandijk  as code owner to validate

PR #1846, opened 2022-01-27
Create CODEOWNERS

dorschw:
code owners


dorschw:
will apply to #1844 instead

PR #1853, opened 2022-01-31
Fix id create artifacts

Noy-Maimon:
fix for nested id field for integrations

### DIFF
--- a/demisto_sdk/commands/common/content/objects/abstract_objects/dictionary_based_object.py
+++ b/demisto_sdk/commands/common/content/objects/abstract_objects/dictionary_based_object.py
@@ -63,5 +63,12 @@ class DictionaryBasedObject(GeneralObject):
 
         return value
 
+    def get_id(self):
+        as_dict = self.to_dict()
+        if 'commonfields' in as_dict.keys():
+            return as_dict.get('commonfields', {}).get('id')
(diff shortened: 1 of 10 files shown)


demisto_sdk/commands/common/content/objects/abstract_objects/dictionary_based_object.py
```diff
@@ -63,5 +63,12 @@ def get(self, key: str, default: Optional[Any] = None) -> Any:
 
         return value
 
+    def get_id(self):
+        as_dict = self.to_dict()
+        if 'commonfields' in as_dict.keys():
+            return as_dict.get('commonfields', {}).get('id')
+        else:
+            return as_dict.get('id')
```
yaakovi, line 71: use `get_entity_id_by_entity_type`

yaakovi (approved):

demisto_sdk/commands/common/content/objects/abstract_objects/dictionary_based_object.py
Noy-Maimon: As discussed, remaining as is since the 'type' used has value mismatch (Integrations vs. Integration) and the object's attribute is only of an inheriting object.

PR #1857, opened 2022-02-01
Use Caching for open files

ChanochShayner:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
Related: https://github.com/demisto/etc/issues/45825

## Description
Add caching to open files functions to improve validate run time.


ChanochShayner:
@guykeller 
This cache mechanism downgrades the run time of:
Create ID Set - from 5.20 m to 3.20 m
Validate all files - from ~45 m to ~28 m

demisto_sdk/commands/common/tools.py
```diff
@@ -470,7 +470,10 @@ def get_last_remote_release_version():
     return ''
 
 
-def get_file(file_path, type_of_file):
+@lru_cache()
```
guykeller, line 473: 1. Have you tried changing the max size? What is our cache miss ratio?
2. Would love to discuss the cache invalidation strategy

ChanochShayner:
Using the maxsize arg doesn't improve the runtime (even with maxsize=10000)
@guykeller FYI

ilappe:
@ChanochShayner you can get the cache info (including miss, current size etc)
by printing the get_file.cache_info() (cache_info is a function added by the decorator)

guykeller (approved):

PR #1870, opened 2022-02-08
IDSet - fix bug in script_executions calculation

ilappe:
## Status
- [x] In Progress
- [ ] Ready

## Description
script_executions wasn't parsed well for JS code


demisto_sdk/commands/common/tests/update_id_set_test.py
```diff
@@ -645,6 +645,26 @@ def test_get_script_data_with_alternative_fields_second_level():
...
+            - A script file which contains javascript code.
+
+        When
+            - parsing scripts files
+
+        Then
+            - make sure the script_executions was parsed successfully.
+        """
+        file_path = TESTS_DIR + '/test_files/Script-js-script-code.yml'
+        script_name = 'TestScriptWithJsCode'
```
yaakovi, line 661: Use the `TestSuite` object instead of creating a test data file.
see the recent tests of `process_layoutscontainer`

yaakovi (approved):

PR #1877, opened 2022-02-10
Fixing readme docs automation bug

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/45822)

## Description
Edited several functions to not add the section automatically to the table unless those sections contain some data.
Also added tests of cases similar to the one in the issue, tested (and failed) before the functions additions(added to screenshots), tested (and succeed) after the function addition

## Screenshots
Screen Shot 2022-02-10 at 9 59 37


demisto_sdk/commands/generate_docs/generate_script_doc.py
```diff
@@ -139,9 +139,11 @@ def get_script_info(script_path):
...
 
     from_version = get_from_version(script_path)
-
-    res = [{'Name': 'Script Type', 'Description': script_type},
-           {'Name': 'Tags', 'Description': tags}]
+    res = []
+    if script_type:
+        res.append({'Name': 'Script Type', 'Description': script_type})
+    if tags:
+        res.append({'Name': 'Tags', 'Description': tags})
```
ShahafBenYakir, line 146: Why are we adding the fix both here and in the generate_table_section function?

demisto_sdk/commands/generate_docs/common.py
```diff
@@ -117,7 +117,10 @@ def generate_table_section(data: list, title: str, empty_message: str = '', text
         section.append('---')
 
     if not data:
-        section.extend([empty_message, ''])
+        if empty_message:
+            section.extend([empty_message, ''])
+        else:
+            section = ['']
```
ShahafBenYakir, line 123: Let's check if the same bug occurs for integration and playbooks.

ShahafBenYakir (commented):
Can we improve the scenarios in the unit tests to be more clear to an average developer? For example what is empty metadata? empty message?

demisto_sdk/commands/generate_docs/generate_script_doc.py
YuvHayun: Because first we don't want to add a row with no description to the table.
But then in case nothing was added to the table, we wouldn't want to create an empty table.

ShahafBenYakir: 👍

ShahafBenYakir (approved):

ShahafBenYakir (changes requested):
Add your changes to changelog.json before merging

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 * Removed PYYAML library (using only ruamel instead).
 * Fixed an issue when filtering items using the ID set in the **create-content-artifacts** command.
+* Fixed an issue where a table was generated with empty description column in the **generate_table_section** command.
```
ShahafBenYakir, line 4: generate_table_section is not the command (it is the function name) the command is generate-docs
```suggestion
* Fixed an issue in the **generate-docs** command where tables were generated with an empty description column.
```

ShahafBenYakir (approved):
See my changelog comment

PR #1879, opened 2022-02-11
Fix create content artifact filter by id set

yaakovi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
related issue: https://github.com/demisto/etc/issues/46570
see `CommonScripts` pack v1.6.35 in MPv2

## Description
Fixed an issue when filtering items using the ID set in the **create-content-artifacts** command.

## Screenshots
failing test without the fix: There is a warning about the skipped item due to the bug.
image

### DIFF
--- a/demisto_sdk/commands/common/content/objects/pack_objects/pack.py
+++ b/demisto_sdk/commands/common/content/objects/pack_objects/pack.py
@@ -76,7 +76,7 @@ class Pack:
                 if is_object_in_id_set(object_id, self._pack_info_from_id_set):
                     yield content_object
                 else:
-                    logging.warning(f'Skipping object {object_path} with id "{object_id}" since its missing from '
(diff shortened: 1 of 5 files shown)


demisto_sdk/commands/common/content/objects/pack_objects/pack.py
```diff
@@ -76,7 +76,7 @@ def _content_files_list_generator_factory(self, dir_name: str, suffix: str) -> I
                 if is_object_in_id_set(object_id, self._pack_info_from_id_set):
                     yield content_object
                 else:
-                    logging.warning(f'Skipping object {object_path} with id {object_id} since its missing from '
+                    logging.warning(f'Skipping object {object_path} with id "{object_id}" since its missing from '
```
tkatzir, line 79:
```suggestion
                    logging.warning(f'Skipping object {object_path} with id "{object_id}" since it\'s missing from '
```

PR #1888, opened 2022-02-14
Auto generate YML from python initial design

DinaMeylakh:
**WARNING** The following files are only for the design and they should *not* be merged to master.

### Design
**YMLMetadataCollector**: The class that will hold the wrappers that collect details on functions that are wrapped with them. 
**Python source code with functions**: The file which initializes the YMLMetadataCollector and uses its' wrappers, providing details explicitly and implicitly.
**YMLGenerator**: The class which is called when generating the yml. It obtains the relevant YMLMetadataCollector and the functions using it from the source code file path. It then makes the YMLMetadataCollector collect the details and writes yml file accordingly.

The following files serve as a POC and a way to document the heart of the design.

#### Where
The logic will sit inside `demisto_sdk/commands/unify` and will be called whenever unify is being called, creating the yml file only when needed. Not as part of the source code.

Running the files results in:
Screen Shot 2022-02-14 at 17 06 15

### DIFF
--- a/demisto_sdk/commands/generate_yml/yml_metadata_collector.py
+++ b/demisto_sdk/commands/generate_yml/yml_metadata_collector.py
@@ -1,8 +1,8 @@
 """This file is a part of the generating yml design. Generating a yml file from a python file."""
 
 
-class DetailsCollector:
-    """The DetailsCollector class provides decorators for integration
+class YMLMetadataCollector:
+    """The YMLMetadataCollector class provides decorators for integration
     functions which contain details relevant to yml generation.
 
     If collect_data is set to true, calling the decorated functions will result in
--- a/demisto_sdk/commands/generate_yml/details_collector.py
+++ b/demisto_sdk/commands/generate_yml/details_collector.py
@@ -1,0 +1,46 @@
+"""This file is a part of the generating yml design. Generating a yml file from a python file."""
+
+
+class DetailsCollector:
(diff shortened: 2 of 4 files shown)


Itay4 (commented):
overall looks good, i think we can improve the naming

demisto_sdk/commands/generate_yml/yml_metadata_collector.py
```diff
@@ -0,0 +1,50 @@
...
+        self.docs = {}
+        self.commands = []
+        self.collect_data = False
+        self.conf = conf
+
+    def set_collect_data(self, value):
+        """A setter for collect_data."""
+        self.collect_data = value
+
+    def add_command(self, command_name=None):
```
tkatzir, line 23: I would simply call it "command"
It's a decorator for a command.

demisto_sdk/commands/generate_yml/yml_metadata_collector.py
```diff
@@ -0,0 +1,50 @@
...
+        """A decorator example. Its' arguments are extra inputs specified in the call.
+        It returns a wrapper which collects details of a command functions
+        if collect_data is set to True.
+        """
+        print("adding command")
+
+        def add_command_wrapper(func):
```
tkatzir, line 30: Similarly - "command_wrapper"

tkatzir (changes requested):
Looks good!

PR #1890, opened 2022-02-15
Add more info to validation error

YuvHayun:
## Related Issues
fixes: https://github.com/demisto/etc/issues/46091

## Description
Added more informative messages to differ between the case where a param related to *INCIDENT_FETCH_REQUIRED_PARAMS* is malformed and the case where it's completely missing.


demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -911,10 +911,16 @@ def is_valid_fetch(self) -> bool:
             for param in params:
                 if 'defaultvalue' in param:
                     param.pop('defaultvalue')
-            for param in fetch_required_params:
-                if param not in params:
-                    error_message, error_code = Errors.parameter_missing_from_yml(param.get('name'),
-                                                                                  yaml.dumps(param))
+            for fetch_required_param in fetch_required_params:
+                if fetch_required_param not in params:
```
ShahafBenYakir, line 915:
```suggestion
                # If this condition returns true, we will go over the params dict and we'll check if there is a param that does match the fetch_required_param name. If there is one, we know that in the params dict there is a matching param to the fetch_required_param but it has a malformed structure.
                if fetch_required_param not in params:
```

ShahafBenYakir (dismissed):
Just add the comment to the code and we are ok

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Fixed an issue in the **split** command where splitting failed when using relative input/output paths.
 * Added warning when inferred files are missing.
 * Added to **validate** a validation for integration image dimensions, which should be 120x50px.
+* Added to **validate** a validation that differ between the case when a fetch required param is malformed or missing.
```
ShahafBenYakir, line 9:
```suggestion
* Improved an error in the **validate** command to better differentiate between the case where a required fetch parameter is malformed or missing.
```

ShahafBenYakir (approved):
See my changelog comment

PR #1891, opened 2022-02-15
install from pipfile.lock

dorschw:
## Status
- [x] In Progress
- [ ] Ready

## Description
install from pipfile.lock

fixes: https://github.com/demisto/etc/issues/46750


dorschw:
irrelevant since #2007

PR #1894, opened 2022-02-16
validate image size

ilaner:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/45608)

## Description
Validate that the integration image dimensions in the correct size.

## Must have
- [ ] Tests
- [ ] Documentation


Pipfile
```diff
@@ -54,6 +54,7 @@ python-dotenv = "*"
 importlib-resources = "*"
 dictdiffer = "*"
 setuptools = "*"
+pillow = "*"
```
dorschw, line 57: consider [this solution](https://stackoverflow.com/a/19035508) to avoid adding dependencies

ilaner: Trying this library:
https://pypi.org/project/imagesize/

demisto_sdk/commands/common/errors.py
```diff
@@ -1045,6 +1047,11 @@ def invalid_image_name_or_location():
                "For more info: https://xsoar.pan.dev/docs/packs/packs-format#author_imagepng\n" \
                "Otherwise, any other image should be located under the 'Doc_files' dir."
 
+    @staticmethod
+    @error_code_decorator
+    def invalid_image_dimensions(width: int, height: int):
+        return f'The image dimensions are {width}x{height}. The requirements are 120x50.'
```
dorschw, line 1053: is it strict 120x50, or up to 120x50?

demisto_sdk/commands/common/hook_validations/image.py
```diff
@@ -51,7 +55,7 @@ def is_valid(self):
             return self._is_valid
 
         is_existing_image = False
-        self.validate_size(allow_empty_image_file=True)
+        self.validate_size(allow_empty_image_file=True, should_validate_dimensions=True)
```
dorschw, line 58: when do we not validate dimensions?

demisto_sdk/commands/common/errors.py
ilaner: strict

demisto_sdk/commands/common/hook_validations/image.py
ilaner: We don't care about dimensions when validating AuthorImage, only in Integration image.

dorschw: Please add a comment mentioning it

demisto_sdk/commands/common/hook_validations/image.py
```diff
@@ -61,21 +65,34 @@ def is_valid(self):
...
         """
         if re.match(IMAGE_REGEX, self.file_path, re.IGNORECASE):
             image_size = os.path.getsize(self.file_path)
             if image_size > maximum_size:  # disable-secrets-detection
                 error_message, error_code = Errors.image_too_large()
                 if self.handle_error(error_message, error_code, file_path=self.file_path):
                     self._is_valid = False
-
+            if should_validate_dimensions:
+                width, height = imagesize.get(self.file_path)
```
DeanArbel, line 91: What happens if the file is not really an image, but just with an image suffix?

DeanArbel (approved):
Nice job, see my comments.

demisto_sdk/commands/common/hook_validations/image.py
ilaner: @DeanArbel 
If there is no image or corrupted one the library will return (-1, -1), therefore it will fail the validation.
Anyway, there is already a check for `is_existing_image` in the same class.

demisto_sdk/commands/common/hook_validations/image.py
```diff
@@ -61,21 +65,34 @@ def is_valid(self):
...
+                      should_validate_dimensions: bool = False,
+                      allowed_width: int = IMAGE_WIDTH,
+                      allowed_height: int = IMAGE_HEIGHT) -> None:
         """
         Checks if image has a valid size.
         if 'allow_empty_image_file' is true, checks that the image file is not empty.
         Args:
             allow_empty_image_file (bool): Whether empty image file is an error.
             maximum_size (int): Maximum allowed size.
+            should_validate_dimensions (bool): Should validate the image dimensions (Not being validated on AuthorImages)
```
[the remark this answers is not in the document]
ilaner: fixed

Issue #1893, opened 2022-02-15
YMLSplitter does not work for relative paths


panw-pduncan:
```bash
$ python3 -m demisto_sdk split -i ./integration.yml
You are using demisto-sdk 1.6.0.
Starting migration of: integration.yml to dir: .
Traceback (most recent call last):
  File "/Users/pduncan/.pyenv/versions/3.8.12/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/pduncan/.pyenv/versions/3.8.12/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/pduncan/.pyenv/versions/3.8.12/lib/python3.8/site-packages/demisto_sdk/__main__.py", line 2032, in <module>
    main()
  File "/Users/pduncan/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/pduncan/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/pduncan/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/pduncan/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/pduncan/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/pduncan/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/decorators.py", line 73, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/Users/pduncan/.pyenv/versions/3.8.12/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/pduncan/.pyenv/versions/3.8.12/lib/python3.8/site-packages/demisto_sdk/__main__.py", line 207, in split
    return yml_splitter.extract_to_package_format()
[...]

PR #1897, opened 2022-02-17
fix relative split output resolving

dorschw:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/46777
fixes: https://github.com/demisto/demisto-sdk/issues/1893

## Description
yml-splitter no longer fails on relative paths.

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/split/ymlsplitter.py
+++ b/demisto_sdk/commands/split/ymlsplitter.py
@@ -76,7 +76,7 @@ class YmlSplitter:
     def get_output_path(self):
         """Get processed output path
         """
-        output_path = self.output
+        output_path = Path(self.output).resolve()
         if self.autocreate_dir and output_path.name in {'Integrations', 'Scripts'}:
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/split/ymlsplitter.py
```diff
@@ -76,7 +76,7 @@ def __init__(self, input: str, output: str = '', file_type: str = '', no_demisto
     def get_output_path(self):
         """Get processed output path
         """
-        output_path = self.output
+        output_path = Path(self.output).resolve()
```
ilaner, line 79: Maybe better to add this in the ctor with both input and output?

dorschw: true, will do, thanks

ilaner (approved):
Amazing

PR #1899, opened 2022-02-17
warn of missing inferred files

dorschw:
## Status
- [x] Ready

## Description
Added a warning when a file is inferred (for example, `foo_description.md` implies the existance of `foo.yml`), but does not exist.


DeanArbel (approved):
Nice

PR #1902, opened 2022-02-20
Support validating and formatting for alternative fields

Noy-Maimon:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/45831

## Description
1. Add the id set the fields 'id_x2' and 'name_x2' for when the item has these alternative fields.
2. Set validate and format to run the needed validations and formatting for items yml\json that mention an item that has an alternative field, for playbooks, incident types, and incident fields (not for scripts, see issue https://github.com/demisto/demisto-sdk/pull/1902#issue-1145093486).


demisto_sdk/commands/format/format_module.py
```diff
@@ -284,9 +283,9 @@ def run_format_on_file(input: str, file_type: str, from_version: str, interactiv
     if file_type not in ('integration', 'playbook', 'script') and 'add_tests' in kwargs:
         # adding tests is relevant only for integrations, playbooks and scripts.
         del kwargs['add_tests']
-    if file_type != FileType.INCIDENT_FIELD.value and 'id_set_path' in kwargs:
-        # relevant only for incidentfield
-        del kwargs['id_set_path']
+    # if file_type != FileType.INCIDENT_FIELD.value and 'id_set_path' in kwargs:
+    #     # relevant only for incidentfield
+    #     del kwargs['id_set_path']
```
Noy-Maimon, line 288: This should be removed but left as comment for now in case there will be any related errors before merging

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1715,3 +1730,46 @@ def check_and_validate_deprecated(self, file_type, file_path, current_file, is_m
...
+        alternative_id, alternative_name = get_alternative_id_and_name_from_id_set(item_id, file_type, self.id_set_file)
+        if alternative_id or alternative_name:
+            items_using_current_item_paths = get_all_using_paths(item_id, file_type, self.id_set_file)
+            for path in items_using_current_item_paths:
+                with open(path, 'r') as file:
+                    using_data = json.loads(file.read()) if path.endswith('json') else yaml.load(file)
+                    file_type = find_type(path)
+                    if check_and_add_missing_alternative_fields(using_data, file_type, self.id_set_file):
+                        # TODO: in the above condition function, there is duplicity with checking if the items have an
+                        #  alternative field (since we got here cause we already checked that...)
```
Noy-Maimon, line 1769: @yaakovi for now i left it as is - with the duplicity- since it is working properly and this PR has been dragged for a while. I can edit it to be more efficient but it will require some more changes  and effort. what do you think?

demisto_sdk/commands/common/errors.py
```diff
@@ -2172,3 +2176,15 @@ def aliases_with_inner_alias(invalid_aliases: List[str]):
...
+    @error_code_decorator
+    def missing_alternative_fields_nested_items():
+        return 'Alternative fields of nested items were found and are missing from this file, ' \
+               'please add them to the file.'  # TODO: improve msg
+
+    @staticmethod
+    @error_code_decorator
+    def missing_alternative_fields_in_using_items(using_id, item_id):
+        return f'Item with id {using_id} is using the item with id {item_id} and it is missing its alternative ' \
+               f'fields. Please add the missing alternative fields to the file.'  # TODO: improve msg
```
Noy-Maimon, line 2190: @yaakovi could use your input with a better error msg in these too errors.
plus - this will suggest a solution for running format on each of the 'using' files (meaning many different calls of format). this is the desired behavior, right? or do we want to allow format to automatically run on all of the using files when running on the single file? (the one that have the alternative field)

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1953,3 +1969,46 @@ def check_and_validate_deprecated(self, file_type, file_path, current_file, is_m
 
             return is_valid_as_deprecated
         return None
+
+    def validate_alternative_fields(self, item_data: dict, file_type: FileType):
```
dantavori, line 1973:
```suggestion
    @error_codes('AF100,AF101')
    def validate_alternative_fields(self, item_data: dict, file_type: FileType):
```

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1953,3 +1967,44 @@ def check_and_validate_deprecated(self, file_type, file_path, current_file, is_m
...
+        # validate alternative fields in items that are using this current item
+        item_id = get_id_from_item_data(item_data)
+        alternative_id, alternative_name = get_alternative_id_and_name_from_id_set(item_id, file_type, self.id_set_file)
+        if alternative_id or alternative_name:
+            items_using_current_item_paths = get_all_using_paths(item_id, file_type, self.id_set_file)
+            for path in items_using_current_item_paths:
+                with open(path, 'r') as file:
+                    using_data = json.loads(file.read()) if path.endswith('json') else yaml.load(file)
+                    file_type = find_type(path)
+                    if check_and_add_missing_alternative_fields(using_data, file_type, self.id_set_file):
```
Noy-Maimon, line 2004: @yaakovi for now i left it as is - with the duplicity- since it is working properly and this PR has been dragged for a while. I can edit it to be more efficient but it will require some more changes and effort. what do you think?

yaakovi: > (recommented this older comment as it was deleted since the file was modified)
?

yaakovi (commented):
haven't finished reviewing the tools.py file

demisto_sdk/commands/validate/validate_manager.py
Noy-Maimon: referring to this one: https://github.com/demisto/demisto-sdk/pull/1902#discussion_r818889318
Since the code was modified, it would not appear in the file when reviewing it. so I copied it and added it again.

demisto_sdk/commands/common/tools.py
```diff
@@ -2456,3 +2457,100 @@ def order_dict(data):
...
+    Args:
+        item_identifier: id or name of sought after item.
+        id_set_section: section of id set of the type of the sought after item.
+
+    Returns:
+        the item's info from the id set.
+    """
+
+    for item in id_set_section:
+        if item_identifier in [list(item.keys())[0], list(item.values())[0].get('name')]:
```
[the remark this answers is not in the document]
Noy-Maimon: But this is alwayer per item-section. As in its always checking only one content item type. Is this what you meant?

demisto_sdk/commands/common/tools.py
```diff
@@ -2456,3 +2457,100 @@ def order_dict(data):
...
+            return item
+    return None
+
+
+def add_missing_alternative_fields(item_data: dict, item_type: str, id_set_file: dict):
```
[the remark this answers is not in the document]
Noy-Maimon: This was already modified, and it is used both in validate and fomat

demisto_sdk/commands/common/schemas/script.yml
```diff
@@ -164,7 +161,5 @@ schema;important_schema:
     description:
       type: str
       required: True
-    description_x2:
```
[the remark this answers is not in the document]
Noy-Maimon: We did think at the start that we will need to change the description too (if it had incidents instead of alerts), but then decided to stay with the id and name.

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -135,6 +135,29 @@ def does_dict_have_alternative_key(data: dict) -> bool:
     return False
 
 
+def get_alternative_fields_from_data(data: dict) -> Union[bool, dict]:
```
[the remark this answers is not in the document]
Noy-Maimon: But this function should return also whether there is any alternative key in the file (that is not id/name). this modifies the `has_alternative_meta` field in the id set.
if we return an empty dict, this will not work.
if we want to return the exact fields (that are not the id/name) then this means we need to change the recursive function `does_dict_have_alternative_key`, and iterate over the nested dicts in a playbook and save the values to a new dict, and this could be a lot of work for data we don't actually need. 
WDYT?

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1953,3 +1967,45 @@ def check_and_validate_deprecated(self, file_type, file_path, current_file, is_m
...
+        Returns:
+            False if the file is invalid due to missing alternative fields, True otherwise.
+
+        """
+        is_valid = True
+
+        # validate the use of alternative fields in current item's file
+        if check_and_add_missing_alternative_fields(item_data, file_type, self.id_set_file):
+            error_message, error_code = Errors.missing_alternative_fields_nested_items()
+            self.handle_error(error_message, error_code, file_path=self.file_path,
```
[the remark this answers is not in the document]
Noy-Maimon: Iit should be ignored if an item on an external repo (that is not meant to be merged to our master) is for a user that is using XSOAR only and not XSIAM. Is this what you meant?

PR #1908, opened 2022-02-22
Fix same folders function

YuvHayun:
## Related Issues
fixes: https://github.com/demisto/etc/issues/45801

## Description
Fixed the **same_folder** function in ***conetent_artifacts_creator_test*** and the related unit tests that were affected by the changes.


.demisto-sdk-report
```diff
@@ -0,0 +1 @@
+[]
```
ShahafBenYakir, line 1: ?

ShahafBenYakir (commented):
I don't think the UT fix requested is just to make the test pass (by adding not to the assertion), the fix should be to go over the tests and check for the problematic logic and correct it to make the assertion correct. @Noy-Maimon Can you give Yuval some more context?

YuvHayun:
> I don't think the UT fix requested is just to make the test pass (by adding not to the assertion), the fix should be to go over the tests and check for the problematic logic and correct it to make the assertion correct. @Noy-Maimon Can you give Yuval some more context?

I didn't just add 'not' to the assertions, I fixed the logic of the **same_folders** function and by requiring a full fail-log of pytest I found out what should be the real value of the assertion and therefore added 'not'

Noy-Maimon:
@ShahafBenYakir @YuvHayun I agree with Shahaf - these are old UT that were probably written assuming that this function works properly, hence the inaccuracies were not detected. The UTs should be modified - check why it is not resulting as a same folder (my assumption - the folder we are comparing **to** is not 100% accurate, and that is what should be modified).

ShahafBenYakir (approved):

PR #1911, opened 2022-02-23
Fix incident field suggestion

JasBeilin:
## Status
- [x] Ready

## Description
Suggestion fix for incident field is not passing validation when pack is written with capital letters.

## Screenshots
Current suggestion:
image

Validate on the fix suggested:
image

After change:

image

## Must have
- [x] Tests


dorschw (changes requested):
Please update changelog.md

dorschw (approved):

PR #1920, opened 2022-03-01
Dev Bug - review release notes - failing on words like IPs, URLs, etc

MLainer1:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/45270

## Description
Review release notes fails on upper case words that are plural

## Screenshots
image

## Must have
- [ ] Tests
- [ ] Documentation


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * Fixed an issue in **update-release-notes** where suggestion fix failed in validation.
 * Fixed a bug where `.env` files didn't load. @nicolas-rdgs
+* Fixed an issue in **review-release-notes** where certain words failed in validation.
```
dantavori, line 5:
```suggestion
* Improved the detection of errors in **review-release-notes** command.
```

demisto_sdk/commands/doc_reviewer/doc_reviewer.py
```diff
@@ -124,6 +124,15 @@ def camel_case_split(camel):
 
         return camel.split()
 
+    @staticmethod
+    def is_upper_case_word_plural(word):
+        """check if a given word is an upper case word in plural, like: URLs, IPs, etc"""
+        if word[-1] == 's':
+            singular_word = word[:-1]
+            if singular_word == singular_word.upper():
+                return True
```
dantavori, line 133:
```suggestion
            return singular_word == singular_word.upper
```

demisto_sdk/commands/doc_reviewer/doc_reviewer.py
```diff
@@ -124,6 +124,15 @@ def camel_case_split(camel):
 
         return camel.split()
 
+    @staticmethod
+    def is_upper_case_word_plural(word):
+        """check if a given word is an upper case word in plural, like: URLs, IPs, etc"""
+        if word[-1] == 's':
```
dantavori, line 130:
```suggestion
        if len(word) > 2 and word[-1] == 's':
```
As, Is, Us...

demisto_sdk/commands/doc_reviewer/doc_reviewer.py
```diff
@@ -302,7 +311,8 @@ def remove_punctuation(word):
     def check_word(self, word):
         """Check if a word is legal"""
         # check camel cases
-        if not self.no_camel_case and self.is_camel_case(word):
+        if not self.no_camel_case and self.is_camel_case(word) and not self.is_upper_case_word_plural(self.remove_punctuation(word)):
```
dantavori, line 314: what if `self.no_camel_case` is True? we don't want to call `is_upper_case_word_plural()` in this case?

dantavori (changes requested):

demisto_sdk/commands/doc_reviewer/doc_reviewer.py
MLainer1: The function `is_upper_case_word_plural()` should be called in a case where the function `is_camel_case(word)` deems a word as camel case, but in fact it is not (like IPs, URLs, etc). Maybe it is better to move the logic in the function `is_upper_case_word_plural()` to `is_camel_case(word)`

dantavori (approved):
niiiice

PR #1921, opened 2022-03-01
review-docs -g functionality 

omerKarkKatz:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/45688

## Description
there was a prev change that removed the functionality of the -g arg in the doc review. 
fixed this bug. 

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/doc_reviewer/doc_reviewer.py
+++ b/demisto_sdk/commands/doc_reviewer/doc_reviewer.py
@@ -150,7 +150,7 @@ class DocReviewer:
             ) in self.SUPPORTED_FILE_TYPES:
                 self.files.append(file)
 
-    def get_files_to_run_on(self, file_path):
+    def get_files_to_run_on(self, file_path=None):
         """Get all the relevant files that the spell-check could work on"""
         if self.git_util:
             self.get_files_from_git()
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/doc_reviewer/doc_reviewer.py
```diff
@@ -150,7 +150,7 @@ def get_files_from_git(self):
             ) in self.SUPPORTED_FILE_TYPES:
                 self.files.append(file)
 
-    def get_files_to_run_on(self, file_path):
+    def get_files_to_run_on(self, file_path=None):
         """Get all the relevant files that the spell-check could work on"""
         if self.git_util:
             self.get_files_from_git()
```
dorschw, line 156: seems duplicate to what we have in `get_files_to_run_on()`, consider using the latter instead.

demisto_sdk/commands/doc_reviewer/doc_reviewer.py
```diff
@@ -198,8 +198,11 @@ def run_doc_review(self):
         if len(self.SUPPORTED_FILE_TYPES) == 1:
             click.secho('Running only on release notes', fg='bright_cyan')
 
-        for file_path in self.file_paths:
-            self.get_files_to_run_on(file_path)
+        if self.file_paths:
+            for file_path in self.file_paths:
+                self.get_files_to_run_on(file_path)
+        else:
+            self.get_files_to_run_on()
```
dorschw, line 205: 1. seems like this part should come before `if self.git_util:`in line 155 
2. self.get_files_to_run_on() may be called twice the way it is now
3. do we intentionally allow running on *both* git files and provided paths at the same time? shouldn't it be mutex?

yaakovi (approved):
update the CHANGELOG.

PR #1924, opened 2022-03-01
fixed image regex

esharf:
## Status
- [x] Ready


esharf:
this is not a code issue that will be resolved via IT ticket

PR #1931, opened 2022-03-07
Adding filter, to add test playbooks to bucket

daryakoval:
## Status
- [x] In Progress
- [ ] Ready


demisto_sdk/commands/common/content/objects/pack_objects/playbook/playbook.py
```diff
@@ -24,4 +24,6 @@ def upload(self, client: demisto_client):
         return client.import_playbook(file=self.path)
 
     def type(self):
+        if 'TestPlaybooks' in self.path.parts:
```
dantavori, line 27: let's use the constant `TEST_PLAYBOOKS_DIR`

demisto_sdk/commands/create_artifacts/content_artifacts_creator.py
```diff
@@ -699,8 +699,9 @@ def dump_pack(artifact_manager: ArtifactsManager, pack: Pack) -> ArtifactsReport
     for script in pack.scripts:
         content_items_handler.handle_content_item(script)
         pack_report += dump_pack_conditionally(artifact_manager, script)
-    for test_playbook in pack.test_playbooks:
-        pack_report += dump_pack_conditionally(artifact_manager, test_playbook)
+    if artifact_manager.remove_test_playbooks:
```
dantavori, line 702: shouldn't this be the opposite?

demisto_sdk/commands/common/content/objects/pack_objects/pack.py
```diff
@@ -71,7 +71,9 @@ def _content_files_list_generator_factory(self, dir_name: str, suffix: str) -> I
             content_object = path_to_pack_object(object_path)
             # skip content items that are not displayed in the id set, if the corresponding flag is used
             if self._filter_items_by_id_set and content_object.type().value not in [FileType.RELEASE_NOTES.value,
-                                                                                    FileType.RELEASE_NOTES_CONFIG.value]:
+                                                                                    FileType.RELEASE_NOTES_CONFIG.value,
+                                                                                    FileType.TEST_PLAYBOOK.value]:
```
Noy-Maimon, line 75: I suggest adding a comment describing why we exclude both ReleaseNotes and TestPlaybooks at this point (since they are missing from the id set but are needed in the pack's zip), for better readability of this code.

yaakovi (approved):

PR #1940, opened 2022-03-10
Run validate -a with Multiprocessing

ChanochShayner:
## Status
- [ ] In Progress
- [x] Ready

## Description
Run the validate all command with Multiprocessing,
This downgrade the run-time of the validate all from ~30 m' to ~10 m'.
Added a parameter to disable this for debugging.

The multiprocessing broke some prints (the percent for example), but IMO the run time is more important here,
And the important prints - the files with the errors - are still there.

## Screenshots

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -439,10 +439,15 @@ def zip_packs(**kwargs) -> int:
     "--allow-skipped",
     help="Don't fail on skipped integrations or when all test playbooks are skipped.",
     is_flag=True)
+@click.option(
+    "--without-multiprocessing",
+    help="run validate all without multiprocessing, for debugging purposes.",
+    is_flag=True, default=False)
 @pass_config
 def validate(config, **kwargs):
     """Validate your content files. If no additional flags are given, will validated only committed files."""
     from demisto_sdk.commands.validate.validate_manager import ValidateManager
+    run_with_mp = not kwargs.pop('without_multiprocessing')
     check_configuration_file('validate', kwargs)
     sys.path.append(config.configuration.env_dir)
 
@@ -479,6 +484,7 @@ def validate(config, **kwargs):
             debug_git=kwargs.get('debug_git'),
             include_untracked=kwargs.get('include_untracked'),
             quite_bc=kwargs.get('quite_bc_validation'),
+            run_with_multiprocessing=run_with_mp,
             check_is_unskipped=not kwargs.get('allow_skipped', False),
         )
         return validator.run_validation()
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -440,14 +440,14 @@ def zip_packs(**kwargs) -> int:
     help="Don't fail on skipped integrations or when all test playbooks are skipped.",
     is_flag=True)
 @click.option(
-    "--without-multiprocessing",
+    "--no-multiprocessing",
     help="run validate all without multiprocessing, for debugging purposes.",
     is_flag=True, default=False)
 @pass_config
 def validate(config, **kwargs):
     """Validate your content files. If no additional flags are given, will validated only committed files."""
     from demisto_sdk.commands.validate.validate_manager import ValidateManager
-    run_with_mp = not kwargs.pop('without_multiprocessing')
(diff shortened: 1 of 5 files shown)


demisto_sdk/__main__.py
```diff
@@ -439,10 +439,15 @@ def zip_packs(**kwargs) -> int:
     "--allow-skipped",
     help="Don't fail on skipped integrations or when all test playbooks are skipped.",
     is_flag=True)
+@click.option(
+    "--without-multiprocessing",
```
dorschw, line 443: perhaps `single-process` or `no-multiprocessing`? not that important anyway

demisto_sdk/__main__.py
```diff
@@ -479,6 +484,7 @@ def validate(config, **kwargs):
             debug_git=kwargs.get('debug_git'),
             include_untracked=kwargs.get('include_untracked'),
             quite_bc=kwargs.get('quite_bc_validation'),
+            run_with_multiprocessing=run_with_mp,
```
dorschw, line 487: `multiprocessing` or reversing into `single_process` would be better/shorter names

demisto_sdk/commands/common/hook_validations/readme.py
```diff
@@ -551,7 +552,14 @@ def start_mdx_server(handle_error: Optional[Callable] = None, file_path: Optiona
...
                     else:
                         raise Exception(error_message)
-        return True
+        yield True
+        ReadMeValidator.stop_mdx_server()
+
+    @staticmethod
+    def add_node_env_vars():
+        content_path = get_content_path()
+        node_modules_path = content_path / Path('node_modules')
```
dorschw, line 561:
```suggestion
        node_modules_path = content_path / 'node_modules'
```
if content_path is a `Path` object, no need to call `Path` on the other, `/` will work on strings too

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -201,7 +204,7 @@ def print_final_report(self, valid):
             return 0
 
         else:
-            all_failing_files = '\n'.join(FOUND_FILES_AND_ERRORS)
+            all_failing_files = '\n'.join(list(set(FOUND_FILES_AND_ERRORS)))
```
dorschw, line 207:
```suggestion
            all_failing_files = '\n'.join(set(FOUND_FILES_AND_ERRORS))
```

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -341,7 +370,7 @@ def run_validations_on_pack(self, pack_path):
             else:
                 self.ignored_files.add(content_entity_path)
 
-        return all(pack_entities_validation_results)
+        return all(pack_entities_validation_results), FOUND_FILES_AND_ERRORS
```
dorschw, line 373: due to the change in return value (bool to tuple), please make sure there are no other calls to this function as they may break.

dorschw (approved):
Nice

demisto_sdk/commands/common/hook_validations/readme.py
ChanochShayner: content_path is a str

demisto_sdk/__main__.py
ChanochShayner: done

demisto_sdk/__main__.py
ChanochShayner: done

demisto_sdk/commands/validate/validate_manager.py
ChanochShayner: Done

demisto_sdk/commands/validate/validate_manager.py
ChanochShayner: done

demisto_sdk/commands/common/hook_validations/readme.py
dorschw: cool, i didn't know you could do `str / Path` -> `Path`

ChanochShayner:
Thanks, @dorschw !!
The quickest CR ever!

PR #1948, opened 2022-03-15
speed up lint by using mount when posible

esharf:
## Status
- [x] Ready

### DIFF
--- a/demisto_sdk/commands/lint/docker_helper.py
+++ b/demisto_sdk/commands/lint/docker_helper.py
@@ -70,7 +70,7 @@ class Docker:
             try:
                 src = Path(src)
                 if src.exists():
-                    mounts.append(Mount(target, str(src.absolute()), 'bind'))
+                    mounts.append(Mount(target, str(src.absolute()), 'bind', read_only=True))
             except Exception:
                 logger.debug(f'Failed to mount {src} to {target}')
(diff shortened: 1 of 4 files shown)


demisto_sdk/commands/lint/linter.py
```diff
@@ -43,6 +43,7 @@
                                                stream_docker_container_output)
 
 logger = logging.getLogger('demisto-sdk')
+CAN_MOUNT_FILES = os.getenv('CIRCLECI', False)
```
dorschw:
```suggestion
SHOULD_MOUNT_FILES = os.getenv('CIRCLECI', False)
```

esharf: added a comment that explains the name

demisto_sdk/commands/lint/docker_helper.py
```diff
@@ -13,6 +13,8 @@
 DOCKER_CLIENT = None
 logger = logging.getLogger('demisto-sdk')
 PATH_OR_STR = Union[Path, str]
+# this will be used to determaiine if the system supports mounts
```
yaakovi, line 16:
```suggestion
# this will be used to determine if the system supports mounts
```

yaakovi (approved):

PR #1963, opened 2022-03-19
Remove Tox, update & add 3.9, 3.10 to CircleCI 

jochman:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/47119
https://github.com/demisto/etc/issues/46759
https://github.com/demisto/etc/issues/44514
https://github.com/demisto/etc/issues/46322

## Description
- Instead of tox - using the base image with node (cimg/python-<<version>>-node).
- Used CircleCI's python orb for installations using pipenv (we can use it for release as well).
- Using native python module (python -m demisto_sdk) instead of script call (demisto-sdk) as we should not install the script. I just added the demisto-sdk folder to PYTHONPATH.
- updated all images to the new cimg.

need to merge first: https://github.com/demisto/content/pull/18192

also in this effort: https://github.com/demisto/dockerfiles/pull/7271
## Must have
- [x] Tests


.circleci/config.yml
```diff
@@ -3,86 +3,75 @@
...
-              pip install tox
-        - run:
-            name: Tox build
-            no_output_timeout: 15m
-            command: |
-              tox -e py38 -v -- --cov=demisto_sdk --cov-report=html
+              mkdir test-results
+              python -m pytest -p no:warnings -v --cov=demisto_sdk --cov-report=html --junitxml=test-results/junit.xml
+        - store_test_results:
+            path: test-results
```
dorschw, line 74: does this allow/use parallelism? (splitting the tests between multiple machines of hte same image to speed the process up)

dorschw:
Does it solve https://github.com/demisto/etc/issues/47123 too?

.circleci/config.yml
jochman: We shouldn't split the test across containers. We should run pytest with one of those plugins: [pytest-concurrent](https://pypi.org/project/pytest-concurrent/) or [pytest-parallel](https://pypi.org/project/pytest-parallel/).

The blocker is [this](https://github.com/demisto/etc/issues/30012) issue.
We can not remove all file usage (which is used deeply, also within our TestSuite), but we surely should remove all `Path.write` from tests.

jochman:
> Does it solve [demisto/etc#47123](https://github.com/demisto/etc/issues/47123) too?

it'll remove 3.7 from the build but we still should change `setup.py`.

.circleci/config.yml
```diff
@@ -95,49 +85,27 @@ jobs:
...
-      - *save_tox_37_cache
-      - persist_to_workspace:
-          root: ~/project
-          paths:
-            - .tox
 
   precommit-checks:
       docker:
-        - image: circleci/python:3.8.3-buster-node
+        - image: cimg/python:3.9.10
```
tkatzir, line 91: Can this image tag be paameterized?

tkatzir (approved):

.circleci/config.yml
```diff
@@ -3,86 +3,79 @@
...
+    default: v3
+  dev-cache-version:
+    type: string
+    default: v3
+  node-cache-version:
+    type: string
+    default: v3
+  global-docker-image:
+    type: string
+    default: cimg/python:3.9.10-node
```
avidan-H, line 29: 3.10?

.circleci/config.yml
```diff
@@ -3,86 +3,79 @@
...
-            command: |
-              npm ci
-        - run:
-            name: Setup Tox
-            command: |
-              pip install tox
+        - python/install-packages:
+            pkg-manager: "pipenv"
+            args: "--dev --system"
+            cache-version: << pipeline.parameters.python-cache-version >>
```
avidan-H, line 69: I don't understand why the cache version is a parameter when it's value is `v3` - if it's relatively constant such as a major version then it doesn't need to be a parameter - and if it isn't, then it's value probably should look other than something like `v3`.

.circleci/config.yml
```diff
@@ -3,86 +3,79 @@
...
-            name: Setup Tox
-            command: |
-              pip install tox
-        - run:
-            name: Tox build
-            no_output_timeout: 15m
-            command: |
-              tox -e py38 -v -- --cov=demisto_sdk --cov-report=html
+              mkdir test-results
+              python -m pytest -p no:warnings -v --cov=demisto_sdk --cov-report=html --junitxml=test-results/junit.xml
```
avidan-H, line 76: Don't we want our code to _amazing_? Maybe we should fail on warnings.

.circleci/config.yml
```diff
@@ -153,49 +123,41 @@ jobs:
               - content
   validate-files:
       docker:
-        - image: circleci/python:3.8.3-buster-node
+        - image: cimg/python:3.9.10-node
```
avidan-H, line 126: why not cacha?
```suggestion
        - image: << pipeline.parameters.global-docker-image >>
```

.circleci/config.yml
```diff
@@ -153,49 +123,41 @@ jobs:
...
-              npm ci
-              export PYTHONPATH=".:${PYTHONPATH}"
-              demisto-sdk -v
+              python -m demisto_sdk -v
 
-              CI_COMMIT_BRANCH="master" demisto-sdk create-id-set -o ./Tests/id_set.json --fail-duplicates
+              CI_COMMIT_BRANCH="master" python -m demisto_sdk create-id-set -o ./Tests/id_set.json --fail-duplicates
+        - python/install-packages:
+            pkg-manager: "pip"
+            args: "gsutil"
```
avidan-H, line 142: Why are there `args` and `pip-dependency-file`?

.circleci/config.yml
```diff
@@ -153,49 +123,41 @@ jobs:
...
               cd content
-
-              export PYTHONPATH=".:${PYTHONPATH}"
-              demisto-sdk -v
+              python -m demisto_sdk -v
               export CIRCLE_ARTIFACTS="/home/circleci/project/artifacts"
 
               gsutil cp gs://marketplace-dist/content/private_id_set.json $CIRCLE_ARTIFACTS/unified_id_set.json
-              CI_COMMIT_BRANCH="master" demisto-sdk merge-id-sets -i1 ./Tests/id_set.json -i2 $CIRCLE_ARTIFACTS/unified_id_set.json -o $CIRCLE_ARTIFACTS/unified_id_set.json --fail-duplicates
+              CI_COMMIT_BRANCH="master" python -m demisto_sdk merge-id-sets -i1 ./Tests/id_set.json -i2 $CIRCLE_ARTIFACTS/unified_id_set.json -o $CIRCLE_ARTIFACTS/unified_id_set.json --fail-duplicates
```
avidan-H, line 153: Let's move `CI_COMMIT_BRANCH` out from being hardcoded into a parameter with a default?

.circleci/config.yml
```diff
@@ -207,6 +169,7 @@ jobs:
         - setup_remote_docker
```
avidan-H, line 169: I think we might need to take another look at this job and reconsider.

.circleci/config.yml
```diff
@@ -229,36 +192,35 @@ jobs:
...
             command: |
-              . .tox/py37/bin/activate
-
               cd content
-              demisto-sdk -v
+              python -m demisto_sdk -v
               mkdir ./tmp
 
-              CIRCLE_BRANCH="master" demisto-sdk create-content-artifacts -a ./tmp
+              CIRCLE_BRANCH="master" python -m demisto_sdk create-content-artifacts -a ./tmp
```
avidan-H, line 218: `CIRCLE_BRANCH` != `CI_COMMIT_BRANCH`? What are the uses of the two different environment variables? Do we need both of them?

avidan-H (changes requested):
Love the initiative. Just a few comments to look at it. I think the next big change is to parallelize the unit tests (and correct the improperly implemented unit tests that are blocking this).

.circleci/config.yml
dorschw: https://github.com/demisto/demisto-sdk/pull/1963/commits/ff8bf2595b484ac7d03294ebb4cefc0921ef82ad#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R23

.circleci/config.yml
dorschw: done

.circleci/config.yml
```diff
@@ -15,18 +15,12 @@ parameters:
...
-  python-cache-version:
-    type: string
-    default: v3
-  dev-cache-version:
-    type: string
-    default: v3
-  node-cache-version:
+  cache-version:
     type: string
     default: v3
```
avidan-H, line 20: I still don't understand this default value. Is there an option to simply cache where the key is the computed hash of the files we care about when they change?

.circleci/config.yml
```diff
@@ -41,13 +35,13 @@ references:
 
     install_node_ci: &install_node_ci
       - node/install-packages:
-          cache-version: << pipeline.parameters.node-cache-version >>
+          cache-version: << pipeline.parameters.cache-version >>
           
     install_demisto_sdk: &install_demisto_sdk
```
avidan-H, line 40: I don't love that this isn't explicit what it's installing in the global environment and is dependent upon where it is executed. If you are going to leave it this way please add a comment explaining the behavior you expect.

.circleci/config.yml
```diff
@@ -123,44 +117,49 @@ jobs:
...
             at: ~/project
         - <<: *install_demisto_sdk
+        - python/install-packages:
+            pkg-manager: "pip"
+            pip-dependency-file: "requirements-gsutil.txt"
         - run:
             name: Set Up & Create Id Set
             when: always
+            environment:
+              CI_COMMIT_BRANCH: "master"
```
avidan-H, line 133: I think the value should be set by a pipeline parameter where the default value is `master`.

.circleci/config.yml
```diff
@@ -123,44 +117,49 @@ jobs:
...
-        - python/install-packages:
-            pkg-manager: "pip"
-            args: "gsutil"
-            pip-dependency-file: "requirements-gsutil.txt"
+              python -m demisto_sdk create-id-set -o ./Tests/id_set.json --fail-duplicates
         - run:
             name: Merge Id Set
             when: always
+            environment:
+              CI_COMMIT_BRANCH: "master"
```
avidan-H, line 143: Same comment here.

.circleci/config.yml
```diff
@@ -123,44 +117,49 @@ jobs:
...
 
               gsutil cp gs://marketplace-dist/content/private_id_set.json $CIRCLE_ARTIFACTS/unified_id_set.json
-              CI_COMMIT_BRANCH="master" python -m demisto_sdk merge-id-sets -i1 ./Tests/id_set.json -i2 $CIRCLE_ARTIFACTS/unified_id_set.json -o $CIRCLE_ARTIFACTS/unified_id_set.json --fail-duplicates
+              python -m demisto_sdk merge-id-sets -i1 ./Tests/id_set.json -i2 $CIRCLE_ARTIFACTS/unified_id_set.json -o $CIRCLE_ARTIFACTS/unified_id_set.json --fail-duplicates
         - <<: *install_node_ci
         - run:
             name: Test validate files and yaml
             when: always
+            environment:
+              CI_COMMIT_BRANCH: "master"
```
avidan-H, line 156: Same comment here.

avidan-H (changes requested):

.circleci/config.yml
```diff
@@ -3,86 +3,73 @@
...
 jobs:
-  tox3-7-unit-tests:
-      # docker image circleci/python:3.8.3-buster-node also contains python 3.7
+  pytest:
+      parameters:
+        pythonversion:
+          type: string
       docker:
-        - image: circleci/python:3.8.3-buster-node
+        - image: << parameters.pythonversion >>
```
avidan-H, line 57:
```suggestion
        - image: cimg/python:<< parameters.pythonversion >>-node
```

.circleci/config.yml
jochman: `CIRCLE_BRANCH` is used only in `deploy.sh`
`CI_COMMIT_BRANCH` is used only in `execute_test_content.py`. We might want to change the CIRCLE_BRANCH to CI_COMMIT_BRANCH and remove the other.

But it is for another pr.

.circleci/config.yml
jochman: It's just which algorithm to use (v3).

.circleci/config.yml
```diff
@@ -270,17 +233,15 @@ workflows:
...
-          <<: *tag_filter
       - checkout-content:
           <<: *tag_filter
       - precommit-checks:
           <<: *tag_filter
-          <<: *reqs_tox
+      - pytest:
+          matrix:
+            parameters:
+              pythonversion: [cimg/python:3.8.12-node, cimg/python:3.9.10-node, cimg/python:3.10.2-node]
```
avidan-H, line 243:
```suggestion
              pythonversion: [3.8.12, 3.9.10, 3.10.2]   
```

.circleci/config.yml
jochman: it's always using the files it needs. npm ci - package-lock.json. pipenv - Pipfile and Pipfile.lock. It's always on the root of the demisto-sdk.

avidan-H (changes requested):
Little odd to have dedicated dependency files for `gsutil` and `pre-commit`. Not sure what benefit you are trying to get from doing that.

.circleci/config.yml
jochman: It is wrong doing so. Run the SDK but the branch., We are running "master" checks on non-master branches,

.circleci/config.yml
jochman: checking if it applicable

.circleci/config.yml
avidan-H: @dorschw, he was next to me as I was reviewing so he committed some CR changes before I technically submitted my review.

.circleci/config.yml
avidan-H: So if I understand correctly, then I don't think we need to make this a pipeline parameter. Can be hardcoded. If we want to update the caching version then it should be a PR.

.circleci/config.yml
avidan-H: I don't follow.

.circleci/config.yml
```diff
@@ -126,6 +126,7 @@ jobs:
         - python/install-packages:
             pkg-manager: "pip"
             pip-dependency-file: "requirements-gsutil.txt"
+        - <<: *install_node_ci
```
avidan-H, line 129: Why is this here? The `install_demisto_sdk` executed on line 125 includes this reference configuration.

avidan-H:
@jochman - what about [this](https://github.com/demisto/demisto-sdk/pull/1963#pullrequestreview-915130187)?

avidan-H (approved):

PR #1964, opened 2022-03-19
Unify - YAML Unifier Interface

Itay4:
## Status
- [x] Ready

## Related Issues
groundwork for https://github.com/demisto/etc/issues/47159

## Description
Implemented the `YAMLUnifier` interface

Did not make any actual code changes, just rearranging the existing code before implementing the fix for https://github.com/demisto/etc/issues/47159

## Must have
- [x] Tests

### DIFF
--- a/demisto_sdk/commands/common/content/objects/pack_objects/abstract_pack_objects/yaml_unify_content_object.py
+++ b/demisto_sdk/commands/common/content/objects/pack_objects/abstract_pack_objects/yaml_unify_content_object.py
@@ -6,7 +6,8 @@ from wcmatch.pathlib import EXTMATCH, Path
 import demisto_sdk.commands.common.content.errors as exc
 from demisto_sdk.commands.common.constants import (INTEGRATIONS_DIR,
                                                    SCRIPTS_DIR, FileType)
-from demisto_sdk.commands.unify.yml_unifier import YmlUnifier
+from demisto_sdk.commands.unify.integration_script_unifier import \
+    IntegrationScriptUnifier
 
 from .yaml_content_object import YAMLContentObject
 
@@ -107,9 +108,9 @@ class YAMLContentUnifiedObject(YAMLContentObject):
         # Directory configuration - Integrations or Scripts
         unify_dir = SCRIPTS_DIR if self._content_type == FileType.SCRIPT else INTEGRATIONS_DIR
         # Unify step
-        unifier = YmlUnifier(input=str(self.path.parent), dir_name=unify_dir, output=dest_dir, force=True,
-                             yml_modified_data=self.to_dict())
-        created_files: List[str] = unifier.merge_script_package_to_yml()
+        unifier = IntegrationScriptUnifier(input=str(self.path.parent), dir_name=unify_dir, output=dest_dir, force=True,
+                                           yml_modified_data=self.to_dict())
+        created_files: List[str] = unifier.unify()
         # Validate that unify succeed - there is not exception raised in unify module.
         if not created_files:
             raise exc.ContentDumpError(self, self.path, "Unable to unify object")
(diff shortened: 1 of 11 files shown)


demisto_sdk/commands/unify/yaml_unifier.py
```diff
@@ -0,0 +1,81 @@
...
+from demisto_sdk.commands.common.constants import (DIR_TO_PREFIX,
+                                                   INTEGRATIONS_DIR,
+                                                   SCRIPTS_DIR)
+from demisto_sdk.commands.common.errors import Errors
+from demisto_sdk.commands.common.handlers import YAML_Handler
+from demisto_sdk.commands.common.tools import get_yml_paths_in_dir, print_error
+
+UNSUPPORTED_INPUT_ERR_MSG = 'Unsupported input. Please provide either: ' \
+                            '1. a directory of an integration or a script. ' \
+                            '2. a path of a GenericModule file.'
```
yaakovi, line 16:
```suggestion
UNSUPPORTED_INPUT_ERR_MSG = '''Unsupported input. Please provide either:
1. a directory of an integration or a script.
2. a path of a GenericModule file.'''
```

demisto_sdk/commands/unify/yaml_unifier.py
```diff
@@ -0,0 +1,81 @@
...
+            input = os.path.abspath(input)
+        if not os.path.isdir(input):
+            print_error(UNSUPPORTED_INPUT_ERR_MSG)
+            sys.exit(1)
+        for optional_dir_name in DIR_TO_PREFIX:
+            if optional_dir_name in input:
+                directory_name = optional_dir_name
+
+        if not directory_name:
+            print_error(UNSUPPORTED_INPUT_ERR_MSG)
```
yaakovi, line 49: Shouldn't we exit here too?

demisto_sdk/commands/unify/yaml_unifier.py
```diff
@@ -0,0 +1,81 @@
...
+
+        self.package_path = input
+        if self.package_path.endswith(os.sep):
+            self.package_path = self.package_path.rstrip(os.sep)
```
yaakovi, line 53: The condition is redundant
```suggestion
        self.package_path = self.package_path.rstrip(os.sep)
```

demisto_sdk/commands/common/content/objects/pack_objects/abstract_pack_objects/yaml_unify_content_object.py
```diff
@@ -107,9 +108,9 @@ def _unify(self, dest_dir: Path) -> List[Path]:
         # Directory configuration - Integrations or Scripts
         unify_dir = SCRIPTS_DIR if self._content_type == FileType.SCRIPT else INTEGRATIONS_DIR
         # Unify step
-        unifier = YmlUnifier(input=str(self.path.parent), dir_name=unify_dir, output=dest_dir, force=True,
-                             yml_modified_data=self.to_dict())
-        created_files: List[str] = unifier.merge_script_package_to_yml()
+        unifier = IntegrationScriptUnifier(input=str(self.path.parent), dir_name=unify_dir, output=dest_dir, force=True,
```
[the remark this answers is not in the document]
Itay4: noted

demisto_sdk/commands/unify/yaml_unifier.py
Itay4: i dont think so, because if we get here from the command run (and not from the content object) then we wont have a dir name

lets discuss - i think we can remove this condition all together

Itay4: @yaakovi fixed as discussed to always use abs path

yaakovi (approved):

PR #1972, opened 2022-03-21
Parallelizing SDK UT

dorschw:
## Status
- [x] In Progress

## Description
Unit tests can be run in parallel

### DIFF
--- a/demisto_sdk/tests/integration_tests/validate_integration_test.py
+++ b/demisto_sdk/tests/integration_tests/validate_integration_test.py
@@ -1,4 +1,5 @@
 import copy
+from copy import deepcopy
 from os.path import join
 
 import pytest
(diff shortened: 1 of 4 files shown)


demisto_sdk/tests/integration_tests/validate_integration_test.py
```diff
@@ -1,4 +1,5 @@
 import copy
+from copy import deepcopy
```
Itay4, line 2: nice

avidan-H: I actually got rid of it now and fixed the cache clearing mechanism of the `get_file` function to work properly and used that instead (per @yaakovi's recommendation).

Itay4: that's disappointing

DeanArbel (approved):

PR #1977, opened 2022-03-22
fixed missing readme error message

dorschw:
showing file type name (`Playbook`) instead of `FileType.PLAYBOOK`


DeanArbel (approved):
Nice.

PR #1984, opened 2022-03-27
Testing pb inputs

JasBeilin:
## Status
- [x] Ready

## Related Issues
related: https://github.com/demisto/etc/issues/44364

## Description
Added logging in case of typo in the values of configuration.

## Screenshots

### DIFF
--- a/demisto_sdk/commands/test_content/TestContentClasses.py
+++ b/demisto_sdk/commands/test_content/TestContentClasses.py
@@ -2122,7 +2122,7 @@ def replace_external_playbook_configuration(client: DefaultApi, external_playboo
 
             # If no value (simple or complex) was found, It is a typo
             if complex is None and simple is None:
-                raise Exception(f'Did not found neither simple nor complex value for field: {input_}. '
+                raise Exception(f'Could not found neither a `simple` nor `complex` value for field: {input_}. '
                                 'A valid configuration should be of the followng format: '
                                 '{<param name>: {"simple", <required value>}}')
 
--- a/demisto_sdk/commands/test_content/TestContentClasses.py
+++ b/demisto_sdk/commands/test_content/TestContentClasses.py
@@ -2123,7 +2123,7 @@ def replace_external_playbook_configuration(client: DefaultApi, external_playboo
             # If no value (simple or complex) was found, It is a typo
             if complex is None and simple is None:
                 raise Exception(f'Did not found neither simple nor complex value for field: {input_}. '
-                                'Please check configuration valid as following: '
+                                'A valid configuration should be of the followng format: '
                                 '{<param name>: {"simple", <required value>}}')
 
             existing_val['value']["simple"] = simple
--- a/demisto_sdk/commands/test_content/TestContentClasses.py
+++ b/demisto_sdk/commands/test_content/TestContentClasses.py
@@ -2113,8 +2113,14 @@ def replace_external_playbook_configuration(client: DefaultApi, external_playboo
         matching_record = list(filter(lambda x: x.get('key') == input_, inputs))
         if matching_record:
             existing_val = matching_record[0]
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -2074,66 +2074,81 @@ def execute_tests(self):
...
         if matching_record:
             existing_val = matching_record[0]
-            existing_val['value']["simple"] = external_playbook_configuration["input_parameters"][input_].get("simple")
-            existing_val['value']["complex"] = external_playbook_configuration["input_parameters"][input_].get("complex")
+            simple = external_playbook_configuration["input_parameters"][input_].get("simple")
+            complex = external_playbook_configuration["input_parameters"][input_].get("complex")
+
+            # If no value (simple or complex) was found, It is a typo
+            if complex is None and simple is None:
+                raise Exception(f'Did not found neither simple nor complex value for field: {input_}. '
```
dorschw, line 2125:
```suggestion
                raise Exception(f'Could not found neither a `simple` nor `complex` value for field: {input_}. '
```

demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -2074,66 +2074,81 @@ def execute_tests(self):
...
+                                'Please check configuration valid as following: '
```
dorschw, line 2126:
```suggestion
                                'A valid configuration should be of the followng format: '
```

demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -2074,66 +2074,81 @@ def execute_tests(self):
...
+                                '{<param name>: {"simple", <required value>}}')
```
dorschw, line 2127: do we not show a complex example on purpose?

demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -2074,66 +2074,81 @@ def execute_tests(self):
...
+
+            existing_val['value']["simple"] = simple
+            existing_val['value']["complex"] = complex
```
dorschw, line 2130: Why do we set it here? How is the behavior different than the way it used to be?

demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -2074,66 +2074,81 @@ def execute_tests(self):
...
             changed_keys.append(input_)
 
         else:
             failed_keys.append(input_)
+
     if failed_keys:
         raise Exception(f'Some input keys was not found in playbook: {",".join(failed_keys)}.')
-    logger_module.info(f"Changing keys: {changed_keys}.")
 
+    logger_module.info(f"Changing keys: {changed_keys}.")
```
dorschw, line 2139:
```suggestion
    logger_module.info(f"Changing missing playbook keys: {changed_keys}.")
```

dorschw (commented):
nice

demisto_sdk/commands/test_content/TestContentClasses.py
JasBeilin: I use the simple and complex in the condition above so instead of getting the value twice I get it once and use it. nothing changed in the behavior.

demisto_sdk/commands/test_content/TestContentClasses.py
JasBeilin: It will be the same but with the word "complex". They have documentation, this is only to make it easier to find misspellings.

demisto_sdk/commands/test_content/TestContentClasses.py
JasBeilin: This is not missing playbook keys. These are keys in the playbook that we change for testing.

dorschw (approved):

PR #1990, opened 2022-03-29
fix: removed type from yaml object

jochman:
- [x] Ready


jochman:
> Why do we remove it? (or why was it there before?) I suppose @ilaner added it to preserve some BC?

by mistake. It's unneeded and brakes the interface. He also asked for that removal.

dorschw (approved):

Issue #1978, opened 2022-03-22
doc-review spellcheck highlights words as mis-spelt even when in known_words 


ds-jdcockrill:
SDK version: 1.6.2

### Reproduction steps

* following the [Adopt-a-Pack instructions](https://xsoar.pan.dev/docs/partners/adopt#text-for-the-pack) I created a new release note using the command: `demisto-sdk update-release-notes -i Packs/DigitalShadows -f`
* I updated the generated release note to replace the placeholder text with `- Completed Adopt-a-Pack process (https://xsoar.pan.dev/docs/partners/adopt#process).`
* I ran the `doc-review` command: `demisto-sdk doc-review -i Packs/DigitalShadows/ReleaseNotes/1_0_3.md`
* I see some warnings, including:

```
 - Words that might be misspelled were found in Packs/DigitalShadows/ReleaseNotes/1_0_3.md:
  - Adopt-a-Pack - did you mean: ['ca', 'co', 'ci', 'cki', 'cu']
If these are not misspelled consider adding them to a known_words file:
  Pack related words: content/Packs/<PackName>/.pack-ignore under the [known_words] section.
  Not pack specific words: content/Tests/known_words.txt
To test locally add --use-packs-known-words or --known-words flags.
```

* I add 'Adopt-a-Pack' to either the `.pack-ignore` or `content/Tests/known_words.txt` and re-run with the appropriate switches.
* It still tells me that `Adopt-a-Pack` is misspelled.

Note: the same behaviour seems to occur with words like "IoC" (note the lowercase middle letter).

### Expected behaviour

* if I add 'Adopt-a-Pack' to either of the `known_words` mechanisms, I shouldn't be warned about it anymore.
* perhaps misspellings shouldn't be highlighted red by the SDK and instead just shown as warnings.

ds-jdcockrill:
See attached file for full log during the `doc-review` command.

[demisto-sdk-issue-1978_1.log](https://github.com/demisto/demisto-sdk/files/8324601/demisto-sdk-issue-1978_1.log)

dorschw:
Thanks for opening an issue, we'll take a look and update here / close this issue with a fix PR.

dorschw:
@ds-jdcockrill this fix will be available starting on demisto-sdk v1.6.4, planned to be released later this month.

PR #2014, opened 2022-04-06
Add support for kebab-case in doc-review

DeanArbel:
## Status
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/47536
fixes: https://github.com/demisto/demisto-sdk/issues/1978

## Description
Added support for sub words suggestion in kebab-case sentences when running the **doc-review** command.


demisto_sdk/commands/doc_reviewer/doc_reviewer.py
```diff
@@ -318,19 +318,23 @@ def remove_punctuation(word):
     def check_word(self, word):
         """Check if a word is legal"""
         # check camel cases
+        word = self.remove_punctuation(word)
+        sub_words = []
+        if '-' in word:
+            sub_words.extend(word.split('-'))
         if not self.no_camel_case and self.is_camel_case(word):
```
dorschw, line 325: `elif`?

demisto_sdk/commands/doc_reviewer/doc_reviewer.py
```diff
@@ -318,19 +318,23 @@ def remove_punctuation(word):
...
-                    self.unknown_words[word] = list(self.spellchecker.candidates(sub_word))[:5]
-
+            sub_words.extend(self.camel_case_split(word))
         else:
-            word = self.remove_punctuation(word)
-            if word.isalpha() and self.spellchecker.unknown([word]):
-                self.unknown_words[word] = list(self.spellchecker.candidates(word))[:5]
+            sub_words.append(word)
+
+        self.unknown_words[word] = []
```
dorschw, line 330: perhaps a set would be better here?

demisto_sdk/commands/doc_reviewer/doc_reviewer.py
DeanArbel: I wasn't too sure about the validity of this case, but let's go with elif.

demisto_sdk/commands/doc_reviewer/doc_reviewer.py
DeanArbel: Had to make sure - this'll only change a print, so it's ok to change it to set.

dorschw (approved):

PR #2020, opened 2022-04-07
Increase docker connection pool in CI systems

esharf:
## Status
- [x] Ready

## Description
- increased the docker SDK version to 4.4.0
- increased the connection pool size from default (10) to 20 inCI systems (using the `CI` environment variable).


yaakovi (approved):

.circleci/config.yml
```diff
@@ -237,6 +237,7 @@ jobs:
         - run:
             name: Test lint on python and ps files
             when: always
+            no_output_timeout: 20m
```
jochman, line 240: why do we need it?

PR #2021, opened 2022-04-08
Handle empty command arguments

darkushin:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/47517

## Description
- Fixes an issue where when command arguments were left empty, both `format` and `validate` would fail unexpectedly.
- After the fix, the `format` command will add `[]` to the empty arguments, and the `validate` command will show an indicative error.

### DIFF
--- a/demisto_sdk/commands/common/errors.py
+++ b/demisto_sdk/commands/common/errors.py
@@ -448,7 +448,8 @@ class Errors:
     @staticmethod
     @error_code_decorator
     def empty_command_arguments(command_name):
-        return f"The arguments of the integration command `{command_name}` are None."
+        return f"The arguments of the integration command `{command_name}` can not be None. If the command has no arguments, " \
+               f"use `arguments: []` or remove the `arguments` field."
 
     @staticmethod
     @error_code_decorator
--- a/demisto_sdk/commands/common/errors.py
+++ b/demisto_sdk/commands/common/errors.py
@@ -448,7 +448,7 @@ class Errors:
     @staticmethod
     @error_code_decorator
     def empty_command_arguments(command_name):
-        return f"The integration command: {command_name} has None as arguments."
+        return f"The arguments of the integration command `{command_name}` are None."
 
     @staticmethod
     @error_code_decorator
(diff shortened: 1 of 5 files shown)


demisto_sdk/commands/common/errors.py
```diff
@@ -224,6 +224,7 @@
     "changed_integration_yml_fields": {'code': "IN147", "ui_applicable": False, 'related_field': 'script'},
     "parameter_is_malformed": {'code': "IN148", 'ui_applicable': False, 'related_field': 'configuration'},
     'empty_outputs_common_paths': {'code': 'IN149', 'ui_applicable': False, 'related_field': 'contextOutput'},
+    "empty_command_arguments": {'code': 'IN150', 'ui_applicable': False, 'related_field': 'arguments'},
```
darkushin, line 227: Not sure whether the `ui_applicable` field should be true or false. Let me know what you think

demisto_sdk/commands/common/errors.py
```diff
@@ -444,6 +445,11 @@ def file_cannot_be_deleted(file_path: str):
     def suggest_fix(file_path: str, *args: Any, cmd: str = 'format') -> str:
         return f'To fix the problem, try running `demisto-sdk {cmd} -i {file_path} {" ".join(args)}`'
 
+    @staticmethod
+    @error_code_decorator
+    def empty_command_arguments(command_name):
+        return f"The arguments of the integration command `{command_name}` are None."
```
dorschw, line 451:
```suggestion
        return f"The arguments of the integration command `{command_name}` must be defined (not None). For an empty list, use `arguments: []`."
```

or whatever the correct syntax for the command is

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -385,6 +385,13 @@ def is_valid_default_argument(self):
 
         for command in commands:
             default_args = set()
+            if command.get('arguments', []) is None:
```
dorschw, line 388: what happens what `arguments` is not in `command`?

dorschw (commented):
nice job

demisto_sdk/commands/common/hook_validations/integration.py
darkushin: same as happened before - `command.get('arguments', [])` will return [], which is not None, so the `if` statement will be False and we continue to the old flow which iterates over the arguments list (empty list so it won't really iterate over anything).

yuvalbenshalom (approved):

PR #2026, opened 2022-04-10
fix get_file with relative paths & refactor 

dorschw:


dorschw:
https://github.com/demisto/etc/issues/47894 will be a quick fix, this refactoring will wait

PR #2027, opened 2022-04-12
dev-bug - Added validation for deleted non-content items

MLainer1:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/47939

## Description
Added validation for deleted non-content items

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/validate/tests/validators_test.py
+++ b/demisto_sdk/commands/validate/tests/validators_test.py
@@ -1823,11 +1823,12 @@ def test_job_unexpected_field_values_in_non_feed_job(repo, capsys,
 
 
 @pytest.mark.parametrize('file_set,expected_output,expected_result,added_files',
-                         (({'mock_file_description.md'}, "[BA115]", False, set()),
+                         (({'Packs/Integration/mock_file_description.md'}, "[BA115]", False, set()),
                           (set(), "", True, set()),
-                          ({'doc_files/image.png'}, "", True, set()),
-                          ({'mock_playbook.yml'}, "", True, {'renamed_mock_playbook.yml'}),
-                          ({Path('mock_playbook.yml')}, "", True, {Path('renamed_mock_playbook.yml')})))
+                          ({'Packs/Integration/doc_files/image.png'}, "", True, set()),
+                          ({'Packs/Integration/Playbooks/mock_playbook.yml'}, "", True, {'renamed_mock_playbook.yml'}),
+                          ({Path('Packs/Integration/Playbooks/mock_playbook.yml')}, "", True, {Path('renamed_mock_playbook.yml')}),
+                          ({Path('non_content_deleted_file.yml')}, "[BA115]", False, set())))
 def test_validate_deleted_files(capsys, file_set, expected_output, expected_result, added_files, mocker):
     """
     Given
--- a/demisto_sdk/commands/validate/tests/validators_test.py
+++ b/demisto_sdk/commands/validate/tests/validators_test.py
@@ -1827,8 +1827,7 @@ def test_job_unexpected_field_values_in_non_feed_job(repo, capsys,
                           (set(), "", True, set()),
                           ({'Packs/Integration/doc_files/image.png'}, "", True, set()),
                           ({'Packs/Integration/Playbooks/mock_playbook.yml'}, "", True, {'renamed_mock_playbook.yml'}),
-                          ({Path('Packs/Integration/Playbooks/mock_playbook.yml')}, "", True, {Path('renamed_mock_playbook.yml')}),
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1215,12 +1216,19 @@ def validate_deleted_files(self, deleted_files, added_files) -> bool:
...
         for file_path in deleted_files:
-            file_path = str(file_path)
-            if not self.was_file_renamed_but_labeled_as_deleted(file_path, added_files):
-                if not self.is_file_allowed_to_be_deleted(file_path):
-                    error_message, error_code = Errors.file_cannot_be_deleted(file_path)
-                    if self.handle_error(error_message, error_code, file_path):
-                        is_valid = False
+            if 'Packs' not in Path(file_path).absolute().parts:
+                # not allowed to delete non-content files
+                error_message, error_code = Errors.file_cannot_be_deleted(file_path)
```
dorschw, line 1221: file_path is passed here, is it a string or a `Path` object? Please add type hints to the method signature if possible.

Further down (row 1226) we call `str()` on it, but not here, why?

demisto_sdk/commands/validate/tests/validators_test.py
```diff
@@ -1823,11 +1823,11 @@ def test_job_unexpected_field_values_in_non_feed_job(repo, capsys,
...
 @pytest.mark.parametrize('file_set,expected_output,expected_result,added_files',
-                         (({'mock_file_description.md'}, "[BA115]", False, set()),
+                         (({'Packs/Integration/mock_file_description.md'}, "[BA115]", False, set()),
                           (set(), "", True, set()),
-                          ({'doc_files/image.png'}, "", True, set()),
-                          ({'mock_playbook.yml'}, "", True, {'renamed_mock_playbook.yml'}),
-                          ({Path('mock_playbook.yml')}, "", True, {Path('renamed_mock_playbook.yml')})))
+                          ({'Packs/Integration/doc_files/image.png'}, "", True, set()),
+                          ({'Packs/Integration/Playbooks/mock_playbook.yml'}, "", True, {'renamed_mock_playbook.yml'}),
+                          ({Path('Packs/Integration/Playbooks/mock_playbook.yml')}, "", True, {Path('renamed_mock_playbook.yml')})))
```
dorschw, line 1830: Please add a case for non-content file, here or in a separate test

dorschw (changes requested):

demisto_sdk/commands/validate/validate_manager.py
MLainer1: It is both. When using the -g flag we get a Path object. When using another flag we get a str

demisto_sdk/commands/validate/tests/validators_test.py
MLainer1: Is there a way to mock the `absolute()` method? when I added a test for it without the Packs in the path It still added it in the absolute path somehow

dorschw: try [this](https://stackoverflow.com/a/70961657)

demisto_sdk/commands/validate/validate_manager.py
dorschw: Oh boy. Please add type hints as `List[Union[str,Path]]` to this method, add `file_path = Path(file_path)` to the beginning of the method, then later call `str` before sending to `Errors.some_error(path)`.

MLainer1: Actually it is a Set, the precommit check fails when adding typing of `Set[Union[Path,str]]`, instead I added just the `set`. Let me know if you have a better solution

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1199,28 +1200,39 @@ def was_file_renamed_but_labeled_as_deleted(file_path, added_files):
...
-            deleted_file_dict = get_file(file_path, find_type(file_path))
-            deleted_file_id = _get_file_id(file_path, deleted_file_dict)
-            if deleted_file_id:
-                for file in added_files:
-                    file = str(file)
-                    file_dict = get_file(file, find_type(file))
-                    if deleted_file_id == _get_file_id(file, file_dict):
-                        return True
+            file_type = find_type(file_path)
+            if file_type:
```
dorschw, line 1204:
```suggestion
            if file_type := find_type(file_path):
```

dorschw (approved):

PR #2034, opened 2022-04-14
fix misnamed validation

dorschw:
a validation seems to not match its name
fixes https://github.com/demisto/etc/issues/48023


yaakovi (changes requested):
This is not the intention of the validation.
Its purpose is to force users to use the package format instead of a single unified YML.
We allow the "old format" (unified) only for old content (has `toversion`) or deprecated content.

PR is invalid.

PR #2044, opened 2022-04-20
fail validations on non-ignorable error codes

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/47631

## Description
fail validations on error codes which cannot be in ignored .pack-ignore. 

## Screenshots

example of how it looks like in the output (panorama xsoar supported):

image

while this is the panorama .pack-ignore 

image

## Must have
- [x] Tests
- [ ] Documentation


GuyAfik:
@idovandijk this is not ready, ill let you know once it will be, therefore you have no reason to review.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+* Added support in the **validate** command to fail validations on non-ignorable errors which are part of the .pack-ignore.
```
dorschw, line 4:
```suggestion
* Added a validation to the **validate** command, failing when non-ignorable errors are present in  .pack-ignore.
```

demisto_sdk/commands/common/hook_validations/base_validator.py
```diff
@@ -26,49 +27,103 @@ class BaseValidator:
 
     def __init__(self, ignored_errors=None, print_as_warnings=False, suppress_print: bool = False,
                  json_file_path: Optional[str] = None):
+        # these are the ignored errors from the .pack-ignore including un-allowed error codes
         self.ignored_errors = ignored_errors if ignored_errors else {}
```
dorschw, line 31:
```suggestion
        self.ignored_errors = ignored_errors or {}
```

demisto_sdk/commands/common/hook_validations/base_validator.py
```diff
@@ -26,49 +27,103 @@ class BaseValidator:
...
 
-        return False
+        return (
+            error_code in ignored_errors_pack_ignore or error_type in ignored_errors_pack_ignore
+        ) and (
+            error_code in ALLOWED_IGNORE_ERRORS
+        )
+
+    @staticmethod
+    def should_error_code_not_be_ignored_in_pack_ignore(error_code, ignored_errors_pack_ignore):
```
dorschw, line 80: what's the difference between this one and `should_ignore_error`?

demisto_sdk/commands/common/hook_validations/base_validator.py
```diff
@@ -97,7 +158,7 @@ def formatted_error_str(error_type):
 
         formatted_error = formatted_error_str('ERROR')
         if should_print and not self.suppress_print:
-            if suggested_fix:
+            if suggested_fix and 'can not be ignored in .pack-ignore' not in formatted_error:
```
dorschw, line 161: Not urgent, but using strings for conditionals is not the best practice, can we use some other flag instead?

demisto_sdk/commands/validate/tests/validators_test.py
```diff
@@ -660,7 +661,7 @@ def test_get_error_ignore_list(self, mocker):
 
         validate_manager = ValidateManager()
         ignore_errors_list = validate_manager.get_error_ignore_list("fake")
-        assert ignore_errors_list['file_name'] == ['BA101', 'SC101', 'BA106']
+        assert ignore_errors_list['file_name'] == ['BA101', 'SC101', 'IN117', 'BA106', 'IN100']
```
dorschw, line 664: why is this modified?

dorschw (commented):
looks great
scheduling us a mini demo before approval

demisto_sdk/commands/common/hook_validations/base_validator.py
GuyAfik: basically the ```should_error_code_not_be_ignored_in_pack_ignore``` method checks out if an error code that is **in** the .pack-ignore should be ignored or not, while the ```should_ignore_error``` method defines if an error should be ignored in general (including errors from .pack-ignore and errors that can be ignored in parter/community supported/deprecated packs

demisto_sdk/commands/common/hook_validations/base_validator.py
GuyAfik: you are right, ill handle it in a different way.

demisto_sdk/commands/validate/tests/validators_test.py
GuyAfik: pay attention that in the ```validator_manager``` until now we collected **only** all the errors which can be ignored (we simply ignored from un-ignorable errors).

I changed the functionality in the way that we would get all the errors from the .pack-ignore including those who can not be ignored.

This is a test which checks that an un-ignorable errors won't be added to the ```ignore_errors``` list. 
hence modified it to include error codes which can not be ignored.
The logic is being handled in the ```base_validator``` now.

demisto_sdk/commands/common/hook_validations/base_validator.py
dorschw: I see, thanks
perhaps a name like `is_error_allowed_in_pack_ignore` (keep in mind to revert the boolean result) will be clearer

GuyAfik: will be done!

dorschw (approved):

PR #2046, opened 2022-04-20
Validate breaking change json

AradCarmi:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://github.com/demisto/etc/issues/45943)

## Description
Added validation that ensuring an existence of a suitable json in case the words ```Breaking change``` appear in the release notes.

## Must have
- [x] Tests
- [x] Documentation


idovandijk:
I'm a bit confused on why we need to check specifically that the JSON exists if it's a breaking change. Don't we have a check that validates that all items from release notes were also modified/new in the pack?

AradCarmi:
When we are upgrade an integration to a new version, in case the new version has breaking change, the server will promote a message with the breaking change. This message extracted from that json.

idovandijk:
Oh I see, great :]

demisto_sdk/commands/common/hook_validations/release_notes.py
```diff
@@ -117,6 +117,21 @@ def is_docker_image_same_as_yml(self) -> bool:
 
         return True
 
+    def is_breaking_change(self) -> bool:
```
dantavori, line 120:
```suggestion
    error_codes('<ADD_HERE_THE_ERROR_CODE>')
    def is_breaking_change(self) -> bool:
```
please take a look at existing validations and see what you need to add - e.g. lines 113-115
do we need to create another error code or use an existing one?

demisto_sdk/commands/common/hook_validations/release_notes.py
```diff
@@ -175,6 +194,7 @@ def is_file_valid(self):
             bool. True if file's release notes are valid, False otherwise.
         """
         validations = [
+            self.is_breaking_change(),
```
dantavori, line 197: let's think of a better name for this method (someone who reads this can think that every release notes file must be with break changes)

demisto_sdk/commands/common/hook_validations/release_notes.py
```diff
@@ -120,6 +120,25 @@ def is_docker_image_same_as_yml(self) -> bool:
 
         return True
 
+    def is_breaking_change(self) -> bool:
+        """
+        In case of a breaking change in the release note, ensure the existence of a proper json file.
+        """
+        is_valid = True
+        if 'breaking change' in self.latest_release_notes.lower():
```
dantavori, line 128: what will happen in this line if there is no release notes file?

demisto_sdk/commands/common/hook_validations/release_notes.py
```diff
@@ -120,6 +120,25 @@ def is_docker_image_same_as_yml(self) -> bool:
...
+            json_path = self.release_notes_file_path[:-2] + 'json'
+            error_message, error_code = Errors.release_notes_bc_json_file_missing(json_path)
+            try:
+                js = get_dict_from_file(path=json_path)[0]  # extract only the dictionary
+                error_message, error_code = Errors.release_notes_bc_json_file_missing(json_path)
```
dantavori, line 133:
```suggestion
```

demisto_sdk/commands/common/hook_validations/release_notes.py
```diff
@@ -120,6 +120,25 @@ def is_docker_image_same_as_yml(self) -> bool:
...
+    def is_breaking_change(self) -> bool:
+        """
+        In case of a breaking change in the release note, ensure the existence of a proper json file.
+        """
+        is_valid = True
+        if 'breaking change' in self.latest_release_notes.lower():
+            json_path = self.release_notes_file_path[:-2] + 'json'
+            error_message, error_code = Errors.release_notes_bc_json_file_missing(json_path)
+            try:
+                js = get_dict_from_file(path=json_path)[0]  # extract only the dictionary
```
dantavori, line 132:
```suggestion
                json_file_content = get_dict_from_file(path=json_path)[0]  # extract only the dictionary
```

demisto_sdk/commands/common/tests/release_notes_test.py
```diff
@@ -475,3 +476,48 @@ def test_get_categories_from_rn(method_to_check, release_notes_content, filled_e
...
+def test_is_breaking_change(release_notes_content, has_json, change_json, expected_result, mocker, repo):
+    """
+    Given
+    - A release note.
+    When
+    - Run is_breaking_change validation.
+    Then
+    - Ensure that if the release note contains 'breaking change', there is also an appropriate json file.
+    """
+    mocker.patch.object(ReleaseNotesValidator, '__init__', lambda a, b: None)
```
dantavori, line 512: if we mock this, what will be the value of `self.latest_release_notes` in every test case?

dantavori (changes requested):
nice,
1. does this validation run only on new files, or can also fail for old RNs with "breaking changes" and no JSON file? (the expected behavior is the first option).
2. although not mentioned in the issue, it's possible to add an auto fix for this in format command. let's add it.

demisto_sdk/commands/common/hook_validations/release_notes.py
AradCarmi: What do you mean by *there is no release notes file*?
This section will run only if there are release note else it will fail by the main class of the SDK.

demisto_sdk/commands/common/hook_validations/release_notes.py
AradCarmi: Can you please explain the reason? In the [last CR](https://github.com/demisto/demisto-sdk/pull/2046#discussion_r869210826) you told me to add error handling.

demisto_sdk/commands/common/tests/release_notes_test.py
AradCarmi: The ***get_validator*** method will set the `self.latest_release_notes` with the `release_notes_content` argument for each test case.

demisto_sdk/commands/common/hook_validations/release_notes.py
dantavori: see line 130

demisto_sdk/commands/common/hook_validations/release_notes.py
dantavori: I mean, what happens if `self.latest_release_notes` is `None` - lower() fill fail.
If I understand correctly, `is_breaking_change()` runs before this check (`has_release_notes_been_filled_out()`).

demisto_sdk/commands/common/hook_validations/release_notes.py
```diff
@@ -120,6 +120,25 @@ def is_docker_image_same_as_yml(self) -> bool:
 
         return True
```
MLainer1, line 122: Please add the error_codes decorator with the corresponding error code of the validation:
`@error_codes('ERROR_CODE')`

AradCarmi:
After investigation, I found that the build run the validate command with the -g flag hence, It will not be a problem to run the command. On the other hand, the nightly will run the command validate with -a flag ie on all files. I ran it locally with the -a flag and did not see any issues regarding the release note. I also put a breakpoint and ran in debug and did not see it entered to the new validation I added. So In conclusion i think it wont fail for old RNs.

dantavori (approved):
great job :)
please before merging make this RN ignorable (consult with others from the team if you need)

PR #2116, opened 2022-05-23
Added the display_name field to the entities in id-set

yaakovpraisler:
## Status
- [x] In Progress
- [ ] Ready

## Description
Added the display_name field to all the entities in id-set.
This is part of a [task](https://github.com/demisto/etc/issues/48188) to format the RN based on the marketplace to compare the entity name in the RN we need to add the same display_name as it was taken when generating release notes.

## Must have
- [x] Tests
- [ ] Documentation


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * Improved logs when running test playbooks (in a build).
 * Fixed an issue in **upload** did not include list-type content items. @nicolas-rdgs
 * Reverted release notes to old format.
+* Added the new `display_name` field to entities in **create-id-set** command.
```
ShahafBenYakir, line 21:
```suggestion
* Added the new `display_name` field to relevant entities in the **create-id-set** command.
```

demisto_sdk/commands/common/hook_validations/release_notes.py
```diff
@@ -63,9 +63,8 @@ def are_release_notes_complete(self):
                     if file_name and file_type and file_type in RN_HEADER_BY_FILE_TYPE:
                         if (RN_HEADER_BY_FILE_TYPE[file_type] not in self.latest_release_notes) or \
                                 (file_name not in self.latest_release_notes):
-                            entity_name = update_rn_util.get_display_name(file)
```
ShahafBenYakir, line 66: How come this is not longer needed?

demisto_sdk/commands/common/tools.py
```diff
@@ -2773,6 +2773,50 @@ def remove_copy_and_dev_suffixes_from_str(field_name: str) -> str:
     return field_name
 
 
+def get_display_name(file_path, file_data={}) -> str:
```
ShahafBenYakir, line 2776: Have we checked this function on the new XSIAM entities as well?

demisto_sdk/commands/common/tests/update_id_set_test.py
```diff
@@ -680,7 +686,7 @@ def test_get_script_data_script_executions(repo, code):
             'commonfields': {
                 'id': 'DummyScript'
             }})
-        res, _ = process_script(script.path, packs={'DummyPack': {}}, marketplace=MarketplaceVersions.XSOAR.value, print_logs=False)
+        res, _ = process_script(script.yml.path, packs={'DummyPack': {}}, marketplace=MarketplaceVersions.XSOAR.value, print_logs=False)
```
ShahafBenYakir, line 689: How come this worked before?

ShahafBenYakir (commented):
Great work. Let's add a test as part of the `update_id_set` test module to explicitly check that we find the right display names. You can write a quick test that goes over all entity types and makes sure their display names appear correctly in the created id_Set.

demisto_sdk/commands/common/hook_validations/release_notes.py
yaakovpraisler: It was unnecessary before as we already got the same value from the same place (get_display_name) in the `file_name` variable (line 62).

demisto_sdk/commands/common/tests/update_id_set_test.py
yaakovpraisler: Reverting this.
It was supposed to be a test fix, but I fixed it in another way and forgot to revert.

demisto_sdk/commands/common/tools.py
yaakovpraisler: It's the same function as it was in the UpdateRN class, so it should behave the same as before.
Or maybe I didn't get it right?

CHANGELOG.md
yaakovpraisler: done.

demisto_sdk/commands/common/tools.py
ShahafBenYakir: You understood correctly but I am asking this since we might have missed this in our original effort and thus it will not work here as well.

ShahafBenYakir (approved):

PR #2120, opened 2022-05-24
fix non-existent-field issue

GuyAfik:
## Description
fixed a bug that if id_set_file was not provided the mapper format would fail.

### DIFF
--- a/demisto_sdk/commands/format/update_layout.py
+++ b/demisto_sdk/commands/format/update_layout.py
@@ -267,7 +267,10 @@ class LayoutBaseFormat(BaseUpdateJSON, ABC):
         Remove non-existent fields from a container layout.
         """
         if not self.id_set_file:
-            click.secho('Skipping non-existent-fields formatting as id_set_path argument is missing', fg='yellow')
+            click.secho(
+                f'Skipping formatting of non-existent-fields for {self.source_file} as id_set_path argument is missing',
+                fg='yellow'
+            )
             return
 
         content_fields = self.get_available_content_fields()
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/format/update_mapper.py
```diff
@@ -67,6 +67,9 @@ def remove_non_existent_fields(self):
         """
         Remove non-existent fields from a mapper.
         """
+        if not self.id_set_file:
```
dorschw, line 70: please add logging, explaining what happens what the behavior is with and id_set, and without it

JudahSchwartz (approved):
@dorschw can this be merged? Does it really need a changelog entry?

demisto_sdk/commands/format/update_layout.py
```diff
@@ -267,6 +267,7 @@ def remove_non_existent_fields_container_layout(self):
         Remove non-existent fields from a container layout.
         """
         if not self.id_set_file:
+            click.secho('Skipping non-existent-fields formatting as id_set_path argument is missing', fg='yellow')
```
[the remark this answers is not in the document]
GuyAfik: ill add the name of the file that's a good point

demisto_sdk/commands/format/update_layout.py
```diff
@@ -267,6 +267,10 @@ def remove_non_existent_fields_container_layout(self):
         Remove non-existent fields from a container layout.
         """
         if not self.id_set_file:
+            click.secho(
```
dorschw, line 270: please replace with `logger = logging.getLogger('demisto-sdk')`, and `logger.warning`

PR #2126, opened 2022-05-25
Add server UI url

daryakoval:
## Status
- [ ] In Progress
- [x] Ready

Adding xsiam server ui url


demisto_sdk/commands/test_content/tools.py
```diff
@@ -81,3 +81,16 @@ def is_redhat_instance(instance_ip: str) -> bool:
         return True
     except CalledProcessError:
         return False
+
+
+def get_server_url(client_host):
```
ShahafBenYakir, line 86:
```suggestion
def get_ui_url(client_host):
```

ShahafBenYakir (approved):

PR #2155, opened 2022-06-06
fix validate should fail on invalid versions in playbook's entities

DinaMeylakh:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-2963?filter=-1)


DinaMeylakh:
Failed TDD unittest for the bug in [8914662](https://github.com/demisto/demisto-sdk/pull/2155/commits/8914662efb3301fa88e21304cddaa0b75201b7bc)

dorschw:
is it related/similar/included in your PR, @israelpoli ?

DinaMeylakh:
@idovandijk I found this as part of the investigation of the recent inconsistent build failures, see the comments in the jira issue for more details: [link](https://jira-hq.paloaltonetworks.local/browse/CIAC-2963?filter=-1)
There is an example failure there that does not fail the build.
```
[ERROR]: Packs/AccentureCTI/Playbooks/playbook-ACTI_Create_Report-Indicator_Associations.yml: [PB110] - Playbook ACTI Create Report-Indicator Associations with version 6.0.0 uses ['CreateIndicatorRelationship'] with a version that does not match the main playbook version. The from version of ['CreateIndicatorRelationship'] should be 6.0.0 or lower.
```
Invalid versions in playbook's entities implies that if a playbook of fromversion x uses a subplaybook of fromversion y, and y>x this is invalid, because we cant use the parent playbook when in version x (and until version y).

Not sure the issue you mentioned is connected since its a different error (no subplaybook versions there)
And you are right, a simple unittest could have caught this.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Fixed an issue where **validate** did not fail on invalid playbook entities' versions.
```
dorschw, line 5: please add a couple words here too, explaining what made them invalid

PR #2158, opened 2022-06-06
Ignore site packages when running mypy

ilaner:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-3033

## Description
Add a flag to ignore site-packages to mypy

### DIFF
--- a/demisto_sdk/commands/lint/tests/command_builder_test.py
+++ b/demisto_sdk/commands/lint/tests/command_builder_test.py
@@ -89,7 +89,7 @@ def test_build_mypy_command(files, py_num, content_path):
     files = [str(file) for file in files]
     expected = f"python3 -m mypy --python-version {py_num} --check-untyped-defs --ignore-missing-imports " \
                f"--follow-imports=silent --show-column-numbers --show-error-codes --pretty --allow-redefinition " \
-               f"--show-absolute-path --cache-dir={expected_cache_dir} {' '.join(files)}"
+               f"--show-absolute-path --no-site-packages --cache-dir={expected_cache_dir} {' '.join(files)}"
     assert expected == output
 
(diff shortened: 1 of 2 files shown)


esharf (approved):

dorschw (approved):

demisto_sdk/commands/lint/tests/command_builder_test.py
```diff
@@ -87,9 +87,14 @@ def test_build_mypy_command(files, py_num, content_path):
...
     output = build_mypy_command(files, py_num, content_path)
     files = [str(file) for file in files]
-    expected = f"python3 -m mypy --python-version {py_num} --check-untyped-defs --ignore-missing-imports " \
-               f"--follow-imports=silent --show-column-numbers --show-error-codes --pretty --allow-redefinition " \
-               f"--show-absolute-path --cache-dir={expected_cache_dir} {' '.join(files)}"
+    if py_num == '3.7':
+        expected = f"python3 -m mypy --python-version {py_num} --check-untyped-defs --ignore-missing-imports " \
+            f"--follow-imports=silent --show-column-numbers --show-error-codes --pretty --allow-redefinition " \
+            f"--show-absolute-path --cache-dir={expected_cache_dir} {' '.join(files)}"
+    if py_num == '2.7':
```
tkatzir, line 94: Can you use an enum instead of `2.7`?

ilaner:
It's better to use `no-site-packages` on all python versions.
We run the `mypy` locally, therefore the package version we use locally is not necessarily the same version we use on docker image.
If we want to lint for correct usage of site packages, we should move the `mypy` to run on docker (which is not possible on python2 integrations)

PR #2168, opened 2022-06-09
Split new entities

Shellyber:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: https://github.com/demisto/etc/issues/47160
https://jira-hq.paloaltonetworks.local/browse/CIAC-593

## Description
Added support to split command for parsing and modeling rules.

## Screenshots
image

## Must have
- [ ] Tests
- [ ] Documentation


demisto_sdk/commands/split/ymlsplitter.py
```diff
@@ -97,14 +98,16 @@ def extract_to_package_format(self) -> int:
         except ValueError as ex:
             print_error(str(ex))
             return 1
-        self.print_logs("Starting migration of: {} to dir: {}".format(self.input, output_path), log_color=LOG_COLORS.NATIVE)
+        self.print_logs("Starting migration of: {} to dir: {}".format(self.input, output_path),
+                        log_color=LOG_COLORS.NATIVE)
         output_path.mkdir(parents=True, exist_ok=True)
         base_name = output_path.name if not self.base_name else self.base_name
         code_file = output_path / base_name
         self.extract_code(code_file)
```
ShahafBenYakir, line 106: Instead of having the try-except block on line 248 ([here](https://github.com/demisto/demisto-sdk/pull/2168/files#diff-913a538463e23a378949fb612aea23e4156a05d20f5922ca84e8e1ff412d0aedR248)) let's just add a condition to check if we are looking at addling or parsing

demisto_sdk/commands/split/ymlsplitter.py
```diff
@@ -113,108 +116,120 @@ def extract_to_package_format(self) -> int:
...
-                    self.print_logs("Detecting python version and setting up pipenv files ...", log_color=LOG_COLORS.NATIVE)
-                    docker = get_all_docker_images(script_obj)[0]
-                    py_ver = get_python_version(docker, self.config.log_verbose)
-                    pip_env_dir = get_pipenv_dir(py_ver, self.config.envs_dirs_base)
-                    self.print_logs("Copying pipenv files from: {}".format(pip_env_dir), log_color=LOG_COLORS.NATIVE)
-                    shutil.copy("{}/Pipfile".format(pip_env_dir), output_path)
-                    shutil.copy("{}/Pipfile.lock".format(pip_env_dir), output_path)
-                    env = os.environ.copy()
-                    env["PIPENV_IGNORE_VIRTUALENVS"] = "1"
+        if self.file_type == 'modelingrule' or self.file_type == 'parsingrule':
```
ShahafBenYakir, line 119:
```suggestion
        if self.file_type in ('modelingrule', 'parsingrule'):
```

demisto_sdk/commands/split/tests/ymlsplitter_test.py
```diff
@@ -29,6 +30,105 @@ def test_extract_long_description(tmpdir):
...
+    """
+    Given:
+        - A unified YML of a Modeling Rule
+    When:
+        - run extract_rules
+    Then:
+        - Ensure that the rules were extracted
+    """
+    extractor = YmlSplitter(input=f'{git_path()}/demisto_sdk/tests/test_files/modelingrule-OktaModelingRules.yml',
+                            output=str(tmpdir.join('temp_rules.xif')), file_type='modelingrule')
```
ShahafBenYakir, line 43: You don't really need the output here as you are not using it in the `extract_rules` function, you can just pass it straight to the function.
```suggestion
```

ShahafBenYakir: This is true for all the tests you are doing for both `extract_rules` and `extract_rule_schema`.

ShahafBenYakir (commented):
Nice.
1. Why are we adding "hardcoded" YMLs for testing and not using the test suite we have built for ad-hoc test files? It will also remove the necessity to use the `tmpdir` package.
2. I am missing one "big" test for each of the types - modeling and parsing rule. A complete use case where we have a unified yml and we call the `extract_to_package_format`once on each, then make sure the extraction worked.

demisto_sdk/commands/split/ymlsplitter.py
```diff
@@ -113,108 +116,120 @@ def extract_to_package_format(self) -> int:
...
+            self.extract_rules("{}/{}.xif".format(output_path, base_name))
+            self.extract_rule_schema("{}/{}.json".format(output_path, base_name))
+            yaml_obj['fromversion'] = "6.8.0"
```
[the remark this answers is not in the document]
Shellyber: We can remove it, but since we creating the yml file here, I thought it is better to create it properly. 
Also not sure about format for the rules.

demisto_sdk/commands/split/tests/ymlsplitter_test.py
Shellyber: What's the difference? if I pass it as part of the YMLSplitter or directly to the function

Shellyber:
@ShahafBenYakir Please review again

demisto_sdk/commands/split/tests/ymlsplitter_test.py
ShahafBenYakir: Cause it's confusing - when you do `YmlSplitter(output='some_path')` it makes me think this is being used inside the flow in some way where in reality both `extract_rules` and `extract_rule_schema` are getting it as an argument.

ShahafBenYakir (approved):

PR #2171, opened 2022-06-12
migrate to gcloud

esharf:
## Status
- [x] Ready

## Related Issues
related: https://github.com/demisto/content/pull/23690

## Description
change the logic to disable the tunnel usage.

### DIFF
--- a/demisto_sdk/commands/test_content/TestContentClasses.py
+++ b/demisto_sdk/commands/test_content/TestContentClasses.py
@@ -687,24 +687,23 @@ class BuildContext:
             self.instances_ips
         )
 
-    def _get_all_integration_config(self, instances_ips: dict) -> Optional[list]:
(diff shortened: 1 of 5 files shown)


demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -687,24 +683,23 @@ def __init__(self, kwargs: dict, logging_module: ParallelLoggingManager):
...
-            instances_ips: The mapping of the urls to the ports used to tunnel it's traffic
-
+            instances_ips: the instances urls
         Returns:
             A dict containing the configuration for the integration if found, else empty list
         """
         if not self.is_nightly:
             return []
-        url, port = list(instances_ips.items())[0]
+        url = instances_ips[0]
```
DeanArbel, line 698: Why are using a list if we always access the first item in the list?

DeanArbel (approved):

PR #2180, opened 2022-06-19
Add Global Marketplace Env

DeanArbel:
## Status
- [x] In Progress

## Description
Add global argument for marketplace version, to make it easier to set behavior in accordance to the currently set ENV variable.

## Notes
This change will adjust the ENV variable for subsequent runs without the marketplace arg.


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 * Added a new validation to the **validate** command to verify that the docker in use is not deprecated.
 * Added support for multiple ApiModules in the **unify** command
+* Added enviornment variable **DEMISTO_SDK_MARKETPLACE**.
```
dorschw, line 6: please add a couple more words about where it's populated and used

dorschw (approved):
Nice. Wdyt of adding UT for this? may prevent some mess in the future.

PR #2183, opened 2022-06-19
fix flaky PB110 validation

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-3227)

## Description
1) Fixed an issue in the ```PB110``` validation where if we have more than one playbook-id, the validation could fail based on the looping order of those playbook-ids
2) enhance the validation error message
3) refactor the validation logic to be not prone to errors (but the validation itself should behave the same)

error message example:

```
[ERROR]: Packs/CommonPlaybooks/Playbooks/playbook-Search_And_Delete_Emails_-_Generic_-_v2_6_1.yml: [PB110] - Playbook Phishing Investigation - Generic v2 with 'fromversion' 6.0.0 uses the following sub-playbooks along with their invalid 'fromversion': [Packs/CommonPlaybooks/Playbooks/playbook-Search_And_Delete_Emails_-_Generic_-_v2_6_1.yml: 6.1.0]. The 'fromversion' of the sub-playbooks should be 6.0.0 or lower.
```

## Must have
- [x] Tests

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/id.py
+++ b/demisto_sdk/commands/common/hook_validations/id.py
@@ -529,53 +529,42 @@ class IDSetValidations(BaseValidator):
         Returns:
             bool. Whether the playbook's version match playbook's entities.
         """
-        # the following dict holds the playbook names as keys and true/false whether the version is valid.
-        # it handles the case where multiple playbook IDs appear in the id_set and each one of them support different versions.
-        # for example:
-        # main_playbook_version = '5.0.0'
-        # id_set = [{'name': 'pb1', 'fromversion': '5.0.0', 'toversion': '5.4.9'}, {'name': 'pb1' - 'fromversion': '5.5.0'}]
-        # entity_status will look like that: { 'pb1': True}
-        entity_status: dict = {}
-        invalid_entities_path_and_version: dict = {}
-        implemented_entities = implemented_entity_list_from_playbook.copy()
+        implemented_entity_list_from_playbook_copy = set(implemented_entity_list_from_playbook.copy())
+
+        entity_ids_to_id_set = {}
+        for entity in entity_set_from_id_set:
+            entity_id = list(entity.keys())[0]
+            if entity_id in implemented_entity_list_from_playbook_copy:
+                if entity.get(entity_id).get('fromversion'):
+                    if entity_id not in entity_ids_to_id_set:
+                        entity_ids_to_id_set[entity_id] = []
+                    entity_ids_to_id_set[entity_id].append(entity.get(entity_id))
+                # ignore entities which do not have fromversion.
+                if entity_id in implemented_entity_list_from_playbook:
+                    implemented_entity_list_from_playbook.remove(entity_id)
+
+        invalid_sub_entities = {}
+        for _id, entities in entity_ids_to_id_set.items():
+            # get entity with the minimum version
+            min_version_entity = min(entities, key=lambda _entity: LooseVersion(_entity.get('fromversion')))
+            min_version = min_version_entity.get('fromversion')
+            is_version_valid = LooseVersion(min_version) <= LooseVersion(main_playbook_version)
+            tasks_data = get_script_or_sub_playbook_tasks_from_playbook(
+                searched_entity_name=_id,
+                main_playbook_data=main_playbook_data
+            )
+            skip_unavailable = all(task_data.get('skipunavailable', False) for task_data in tasks_data) \
+                if tasks_data and LooseVersion(main_playbook_version) >= LooseVersion('6.0.0') else False
+            if not (is_version_valid or skip_unavailable):
+                sub_entity_file_path = min_version_entity.get('file_path') or ''
+                invalid_sub_entities[sub_entity_file_path] = min_version_entity
+
         is_valid = True, None
-        for entity_data_dict in entity_set_from_id_set:
-            entity_id = list(entity_data_dict.keys())[0]
-            all_entity_fields = entity_data_dict[entity_id]
-            entity_name = entity_id if entity_id in implemented_entity_list_from_playbook else all_entity_fields.get(
-                "name")
-            is_entity_used_in_playbook = entity_name in implemented_entity_list_from_playbook
-
-            if is_entity_used_in_playbook:
-                tasks_data = get_script_or_sub_playbook_tasks_from_playbook(searched_entity_name=entity_name,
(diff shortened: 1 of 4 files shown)


GuyAfik:
failed unit-test - https://app.circleci.com/pipelines/github/demisto/demisto-sdk/16975/workflows/81d70075-809b-4f75-8050-e2e6801ec517/jobs/108248

GuyAfik:
failed unit-test - https://app.circleci.com/pipelines/github/demisto/demisto-sdk/16976/workflows/ce745594-6878-4f1a-9cc0-2fadafb3f525/jobs/108253

GuyAfik:
@JudahSchwartz i guess we can shorten the ```CHANGELOG.md``` here to a more concise string, let me know if you have got something better

demisto_sdk/commands/common/hook_validations/id.py
```diff
@@ -507,7 +507,7 @@ def get_commands_to_integration(self, file_name, file_path):
     @error_codes('PB110,PB117')
     def is_entity_version_match_playbook_version(self, implemented_entity_list_from_playbook,
```
JudahSchwartz, line 508: there are some logic issues here as discussed in person,
lets refactor this method. Broadly - group entities by id so we will have have a list of versions by id, and only compare to the min version

JudahSchwartz (changes requested):

demisto_sdk/commands/common/hook_validations/id.py
```diff
@@ -517,69 +525,80 @@ def is_entity_version_match_playbook_version(self, implemented_entity_list_from_
...
-        invalid_version_entities = [entity_name for entity_name, status in entity_status.items() if status is False]
-
-        if invalid_version_entities:
+        def get_skip_unavailable(_entity_name) -> bool:
+            tasks_data = get_script_or_sub_playbook_tasks_from_playbook(
+                searched_entity_name=_entity_name,
+                main_playbook_data=main_playbook_data
+            )
+            return all(task_data.get('skipunavailable', False) for task_data in tasks_data) \
+                if tasks_data and LooseVersion(main_playbook_version) >= LooseVersion('6.0.0') else False
```
dorschw, line 546: Please use `packaging.version.Version` instead of `LooseVersion`

demisto_sdk/commands/common/hook_validations/id.py
```diff
@@ -517,69 +525,80 @@ def is_entity_version_match_playbook_version(self, implemented_entity_list_from_
...
+
+        def is_minimum_version_valid(_min_version_entity) -> bool:
+            """
+            In case we have more than one entity with the same ID,
+            verify that the one with the minimum version is valid.
+            """
+            return LooseVersion(min_version_entity.get('fromversion')) <= LooseVersion(main_playbook_version)
```
dorschw, line 553: here too

demisto_sdk/commands/common/hook_validations/id.py
```diff
@@ -517,69 +525,80 @@ def is_entity_version_match_playbook_version(self, implemented_entity_list_from_
...
+
+        implemented_entity_list_from_playbook = set(implemented_entity_list_from_playbook)
+        implemented_ids_in_id_set = set()
+
+        implemented_entity_name_to_entities: Dict[str, list] = {}
+        for entity in entity_set_from_id_set:
+            entity_id = list(entity.keys())[0]
```
dorschw, line 560:
```suggestion
            entity_id = next(entity.keys())
```

dorschw (commented):
Nice

demisto_sdk/commands/common/hook_validations/id.py
GuyAfik: ```TypeError: 'dict_keys' object is not an iterator```

dorschw (approved):
Works for me, please wait for @JudahSchwartz to CR this too.

idovandijk:
@GuyAfik are we sure that the invalid fromversion is in the subplaybooks and not the main playbook? Also what is the issue with the loops? Can you please let me know before you merge since this PR seems to affect playbooks

GuyAfik:
> @GuyAfik are we sure that the invalid fromversion is in the subplaybooks and not the main playbook? Also what is the issue with the loops? Can you please let me know before you merge since this PR seems to affect playbooks

Basically this validation verifies that a main playbook ```fromversion``` must be higher or equal to the ```fromversion``` of its sub-playbooks/used scripts. So regarding the first part of your question, this was the way the validation was implemented in the first place.

This validation failed on the nightly build of the upload flow, it seemed to be flaky, sometimes it passed and sometimes it didn't.

After digging inside into it, I have found out that the validation fails on a specific case where if we have several sub-playbooks with the same ID, then there could be a possibility where by mistake we could fail the validation even in case that a sub-playbook ID with ```fromversion``` < main-playbook ```fromversion``` exist based on their looping order in the code. So we failed the validation even when it shouldn't have failed in the old implemented code.

Along with @JudahSchwartz we have decided to change to way this method is implemented since it was prone to unwanted errors.

The internal logic of the validation has not been changed! That means that the behaviour should be exactly the same as before. The only difference now is the way this method is implemented.

idovandijk:
> > @GuyAfik are we sure that the invalid fromversion is in the subplaybooks and not the main playbook? Also what is the issue with the loops? Can you please let me know before you merge since this PR seems to affect playbooks
> 
> Basically this validation verifies that a main playbook `fromversion` must be higher or equal to the `fromversion` of its sub-playbooks/used scripts. So regarding the first part of your question, this was the way the validation was implemented in the first place.
> 
> This validation failed on the nightly build of the upload flow, it seemed to be flaky, sometimes it passed and sometimes it didn't.
> 
> After digging inside into it, I have found out that the validation fails on a specific case where if we have several sub-playbooks with the same ID, then there could be a possibility where by mistake we could fail the validation even in case that a sub-playbook ID with `fromversion` <= main-playbook `fromversion` exist based on their looping order in the code. So we failed the validation even when it shouldn't have failed in the old implemented code.
> 
> Along with @JudahSchwartz we have decided to change to way this method is implemented since it was prone to unwanted errors.
> 
> The internal logic of the validation **has not been changed!** That means that the behaviour should be exactly the same as before. The only difference now is the way this method is implemented.

Got it, thanks :)

PR #2189, opened 2022-06-21
Verify package directory name is the same as the files contained in the package

Shellyber:
## Status
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-1871

## Description
Validating that the following files have the right prefix:
- yml
- png
- py
- md

## Must have
- [x] Tests

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/description.py
+++ b/demisto_sdk/commands/common/hook_validations/description.py
@@ -142,7 +142,7 @@ class DescriptionValidator(BaseValidator):
             integrations_folder = os.path.basename(os.path.dirname(path))
 
             # drop file extension
-            description_file_base_name, _ = description_file.split('_')
+            description_file_base_name = description_file.split('_')[0]
 
         # checking if there are any .md files only for description with a wrong name
         for path in md_paths:
--- a/demisto_sdk/commands/common/hook_validations/description.py
+++ b/demisto_sdk/commands/common/hook_validations/description.py
@@ -142,7 +142,7 @@ class DescriptionValidator(BaseValidator):
             integrations_folder = os.path.basename(os.path.dirname(path))
 
             # drop file extension
-            description_file_base_name = description_file.split('_')[0]
+            description_file_base_name = description_file.rsplit('_', 1)[0]
 
         # checking if there are any .md files only for description with a wrong name
         for path in md_paths:
--- a/demisto_sdk/commands/common/hook_validations/description.py
+++ b/demisto_sdk/commands/common/hook_validations/description.py
@@ -37,7 +37,7 @@ class DescriptionValidator(BaseValidator):
         self.verify_demisto_in_description_content()
 
         # make sure the description is a seperate file
-        if not self.data_dictionary.get('detaileddescription'):
+        if not self.data_dictionary.get('detaileddescription') and '.md' in self.file_path:
             self.is_valid_description_name()
             self.contains_contrib_details()
 
--- a/demisto_sdk/commands/common/hook_validations/description.py
+++ b/demisto_sdk/commands/common/hook_validations/description.py
@@ -133,15 +133,23 @@ class DescriptionValidator(BaseValidator):
     @error_codes('DS106')
     def is_valid_description_name(self):
         """Check if the description name is valid"""
+        integrations_folder = ''
+        description_file_base_name = ''
         description_path = glob.glob(os.path.join(os.path.dirname(self.file_path), '*_description.md'))
         md_paths = glob.glob(os.path.join(os.path.dirname(self.file_path), '*.md'))
+        for path in description_path:
+            description_file = os.path.basename(path)
+            integrations_folder = os.path.basename(os.path.dirname(path))
+
+            # drop file extension
+            description_file_base_name, _ = description_file.split('_')
 
         # checking if there are any .md files only for description with a wrong name
         for path in md_paths:
             if path.endswith("README.md") or path.endswith("CHANGELOG.md"):
                 md_paths.remove(path)
 
-        if not description_path and md_paths:
+        if not description_path and md_paths or integrations_folder != description_file_base_name:
             error_message, error_code = Errors.invalid_description_name()
 
             if self.handle_error(error_message, error_code, file_path=self.file_path):
--- a/demisto_sdk/commands/common/hook_validations/description.py
+++ b/demisto_sdk/commands/common/hook_validations/description.py
@@ -133,16 +133,15 @@ class DescriptionValidator(BaseValidator):
     @error_codes('DS106')
     def is_valid_description_name(self):
         """Check if the description name is valid"""
-        integrations_folder = ''
-        description_file_base_name = ''
         description_path = glob.glob(os.path.join(os.path.dirname(self.file_path), '*_description.md'))
         md_paths = glob.glob(os.path.join(os.path.dirname(self.file_path), '*.md'))
-        for path in description_path:
-            description_file = os.path.basename(path)
-            integrations_folder = os.path.basename(os.path.dirname(path))
 
-            # drop file extension
-            description_file_base_name = description_file.rsplit('_', 1)[0]
+        description_file_path = self.file_path
+        integrations_folder = os.path.basename(os.path.dirname(description_file_path))
+        description_file = os.path.basename(description_file_path)
+
+        # drop file extension
+        description_file_base_name = description_file.rsplit('_', 1)[0]
 
         # checking if there are any .md files only for description with a wrong name
         for path in md_paths:
--- a/demisto_sdk/commands/common/hook_validations/image.py
+++ b/demisto_sdk/commands/common/hook_validations/image.py
@@ -199,8 +199,13 @@ class ImageValidator(BaseValidator):
     def is_valid_image_name(self):
         """Check if the image name is valid"""
         image_path = self.file_path
+        integrations_folder = os.path.basename(os.path.dirname(image_path))
+        image_file = os.path.basename(image_path)
 
-        if not image_path.endswith("_image.png"):
+        # drop file extension
+        image_file_base_name, _ = image_file.split('_')
+
+        if not image_path.endswith("_image.png") or integrations_folder != image_file_base_name:
             error_message, error_code = Errors.invalid_image_name()
 
             if self.handle_error(error_message, error_code, file_path=image_path):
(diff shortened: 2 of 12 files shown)


demisto_sdk/commands/common/hook_validations/description.py
```diff
@@ -133,15 +133,23 @@ def is_duplicate_description(self):
...
+        integrations_folder = ''
+        description_file_base_name = ''
         description_path = glob.glob(os.path.join(os.path.dirname(self.file_path), '*_description.md'))
         md_paths = glob.glob(os.path.join(os.path.dirname(self.file_path), '*.md'))
+        for path in description_path:
+            description_file = os.path.basename(path)
+            integrations_folder = os.path.basename(os.path.dirname(path))
+
+            # drop file extension
+            description_file_base_name = description_file.rsplit('_', 1)[0]
```
dorschw, line 145: Please prefer pathlib.Path when possible

demisto_sdk/commands/common/hook_validations/image.py
```diff
@@ -199,8 +199,13 @@ def is_not_default_image(self):
     def is_valid_image_name(self):
```
ShahafBenYakir, line 199: @Shellyber How are you certain this only runs on integration and images and not also on other images?

demisto_sdk/commands/common/hook_validations/description.py
ShahafBenYakir, line 145: Don't really see a reason for the for loop here. Also, let's put this logic before line 138.
```suggestion

        description_file = os.path.basename(path)
        integrations_folder = os.path.basename(os.path.dirname(path))

        # drop file extension
        description_file_base_name = description_file.rsplit('_', 1)[0]
```

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -1263,6 +1264,36 @@ def is_valid_integration_file_path(self) -> bool:
 
         return True
 
+    @error_codes('IN137')
+    def is_valid_py_file_names(self):
+        # Gets the all integration .py files that may have the integration name as base name
```
ShahafBenYakir, line 1269: I don't think I understand the docstring here

demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -1263,6 +1264,36 @@ def is_valid_integration_file_path(self) -> bool:
...
+
+        for file_path in files_to_check:
+            file_name = os.path.basename(file_path)
+            if file_name not in excluded_files:
+                if file_name.endswith('_test.py'):
+                    base_name = file_name.rsplit('_', 1)[0]
+
+                else:
+                    # integration.py
+                    base_name = file_name.rsplit('.', 1)[0]
```
ShahafBenYakir, line 1284:
```suggestion
                # We unit tests has '_test.py' suffix whereas integration.py files only has file suffix
                splitter = '_' if file_name.endswith('_test.py') else '.'
                base_name = file_name.rsplit(splitter, 1)[0]
```

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1015,6 +1015,9 @@ def validate_beta_integration(self, structure_validator, pack_error_ignore_list)
         return integration_validator.is_valid_beta_integration()
 
     def validate_image(self, file_path, pack_error_ignore_list):
+        pack_name = get_pack_name(file_path)
+        if pack_name == 'NonSupported':
+            return True
```
ShahafBenYakir, line 1020: Why is thus suddenly needed?

demisto_sdk/commands/common/tests/description_test.py
```diff
@@ -106,10 +106,12 @@ def test_is_valid_description_name(repo):
     assert description_validator.is_valid_description_name()
 
 
-def test_is_invalid_description_name(repo):
+@pytest.mark.parametrize('file_name', ['IntName_desc.md', 'IntNameTest_description.md'])
+def test_is_invalid_description_name(repo, file_name):
```
ShahafBenYakir, line 110: nice

demisto_sdk/commands/common/tests/image_test.py
```diff
@@ -173,7 +175,8 @@ def test_is_valid_image_name_with_valid_name(repo):
     assert image_validator.is_valid_image_name()
 
 
-def test_is_valid_image_name_with_invalid_name(repo):
+@pytest.mark.parametrize('file_name', ['IntName_img.png', 'IntNameTest_image.png'])
```
ShahafBenYakir, line 178: let's fix docstrings accordingly

demisto_sdk/commands/common/tests/integration_test.py
```diff
@@ -969,6 +969,31 @@ def test_invalid_integration_path(self, integration, mocker):
 
         assert not validator.is_valid_integration_file_path()
 
+    @pytest.mark.parametrize('file_name', ['IntNameTest.py', 'IntNameTest_test.py'])
+    def test_invalid_py_file_names(self, repo, file_name):
+        """
+        Given
+            - An integration with invalid python file path.
```
ShahafBenYakir, line 976: There are two different cases here, let's address them separately as they belong to two different flows.

ShahafBenYakir (commented):
- Shouldn't we have this validation also for YMLs?
- Also I didn't see here any positive test (check that validation passes on a perfectly normal file). Do we already have such?

demisto_sdk/commands/validate/validate_manager.py
Shellyber: Cause it validates the NonSupported for Qradar_5_4_9 and it causes a lot of problem in the regular build. 
We don't need to validate this pack at all in my opinion

demisto_sdk/tests/workflow_test.py
```diff
@@ -286,34 +286,6 @@ def modify_entity(content_repo: ContentGitRepo, monkeypatch: MonkeyPatch):
     content_repo.run_validations()
 
 
-def all_files_renamed(content_repo: ContentGitRepo, monkeypatch: MonkeyPatch):
```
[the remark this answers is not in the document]
Shellyber: Cause this use case is unnecessary anymore.
This is exactly what's this validation tests. 
The check here changes files in the `HelloWorld' integration folder to `helloworld` (file names) - therefore it fails

Shellyber:
> * Shouldn't we have this validation also for YMLs?
> * Also I didn't see here any positive test (check that validation passes on a perfectly normal file). Do we already have such?

@ShahafBenYakir The validation for yml files already existed. I added the validation for the rest of the files
And yes, we have already some positive tests that in valid files this validation won't fail (I'll have another look just in case I missed something)

demisto_sdk/commands/common/hook_validations/image.py
Shellyber: Other images? like what? 
This runs as part of the validate_manager which runs on our files (packs).

idovandijk (commented):
@Shellyber I see that I was unassigned as reviewer, so I assume that playbook YMLs and playbook PNGs are not affected?
Also if I understand correctly this can cause us to have to rename lots of files now like [Shahaf said here](https://github.com/demisto/etc/issues/19766#issuecomment-950351453).
Can you please describe what is affected and why we need this verification (at such a large scale if I'm not wrong)?

Shellyber:
@idovandijk Indeed it does not affect playbooks at all. 
The `init` command of the sdk already taking care of most of integrations, but we want to prevent name changing and keep a unified convention to all - this is our standard, so just enforcing it. 
We already merged 2 PRs to fix the files already in the content (no more btw, so thankfully not a large scale)
- https://github.com/demisto/content/pull/19707
- https://github.com/demisto/content/pull/19708

demisto_sdk/commands/common/hook_validations/description.py
Shellyber: The code here is written like this:
`description_path = glob.glob(os.path.join(os.path.dirname(self.file_path), '*_description.md'))`
The description path returns a list, that's why the for loop is here.
I didn't want to refactor this at the moment

demisto_sdk/commands/common/tests/integration_test.py
Shellyber: Do you mean split the `integration.py` and the `_test.py`?
If so why?

demisto_sdk/tests/workflow_test.py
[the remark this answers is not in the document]
ShahafBenYakir: Why not just change the folder to `helloworld` too?

demisto_sdk/commands/common/hook_validations/image.py
ShahafBenYakir: I mean that you also have other types of images (not only integrations - for docs/author etc...), other types of images don't require the naming constrains we are adding now

demisto_sdk/commands/validate/validate_manager.py
ShahafBenYakir: The problem exists without your changes?

demisto_sdk/commands/common/hook_validations/image.py
Shellyber: It won't run on those. 
The author image has it's own validation and the others we don't validate at all. its just for our packages - won't affect others

demisto_sdk/tests/workflow_test.py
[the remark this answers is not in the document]
Shellyber: It makes this test irrelevant (as it is). It won't check anything - the whole purpose here is the renaming, if there isn't one it won't test anything. We don't need this anymore

demisto_sdk/commands/common/tests/integration_test.py
ShahafBenYakir: If I understand this test correctly - you are validating onc an integration py file and once a unit test py file, to make the test clearer it is better to address each of them separately in the docstring such that the next dev that looks at the logic understands exactly the two different use cases being checked here.

demisto_sdk/commands/validate/validate_manager.py
Shellyber: It is still happening for some reason at the build - because it runs `validate all` probably
If I do update the image, and I need to update also the size to be compatible, it kinda deletes the old file and add the image as new - which fails the validation completely as the file it looking for doesn't exists anymore. 
WDYT we should do?
Open a PR to fix only the naming and force merge or Open a PR fix the entire image and force? 
Or ignore this pack at all from our entire validation

demisto_sdk/commands/common/tests/integration_test.py
Shellyber: As you wish

ShahafBenYakir (commented):
See my comments

PR #2196, opened 2022-06-23
Add manual integration.

daryakoval:
## Status
- [x] In Progress
- [x] Ready

## Description
In XSIAM we can't create a new incident/alert using API call.
We need a correlation rule in order to create an alert.
We want to create an alert manually, when we send an API call to XSIAM server to create a new alert.
Server check which integration sent a new alert, if the request was sent manually and not from integration it
sets "sourceBrand" header to be "Manual". XSIAM Server looks for a correlation rule for such sourceBrand,
and if there is no such correlation rule, no alert will be created.
If there is a correlation rule for "Manual" integration the allert will be created.

If this step fails please create an integration with id and name "Manual", set isFetch: true for such
integration and make sure that the corresponding correlation rule is created.

## Must have
- [ ] Tests
- [x] Documentation

### DIFF
--- a/demisto_sdk/commands/test_content/TestContentClasses.py
+++ b/demisto_sdk/commands/test_content/TestContentClasses.py
@@ -41,7 +41,6 @@ from demisto_sdk.commands.test_content.tools import (
 
 json = JSON_Handler()
 
-
 ENV_RESULTS_PATH = './artifacts/env_results.json'
 FAILED_MATCH_INSTANCE_MSG = "{} Failed to run.\n There are {} instances of {}, please select one of them by using " \
                             "the instance_name argument in conf.json. The options are:\n{}"
@@ -710,7 +709,8 @@ class BuildContext:
             return {self.server: None}
         if self.is_xsiam:
             return {env.get('base_url'): None for env in self.env_json}
-        instances_ips = {env.get('InstanceDNS'): env.get('TunnelPort') for env in self.env_json if env.get('Role') == self.server_version}
+        instances_ips = {env.get('InstanceDNS'): env.get('TunnelPort') for env in self.env_json if
+                         env.get('Role') == self.server_version}
         return instances_ips
 
     def get_public_ip_from_server_url(self, server_url: str) -> str:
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -2068,6 +2068,8 @@ def __init__(self, build_context: BuildContext, server_private_ip: str, tunnel_p
         self.executed_in_current_round: Set[str] = set()
         self.prev_system_conf: dict = {}
         self.use_retries_mechanism: bool = use_retries_mechanism
+        if IS_XSIAM:
+            self.check_if_can_create_manual_alerts(self.client)
```
dantavori, line 2072:
```suggestion
            self.check_if_can_create_manual_alerts()
```
use self.client in the function

dantavori (approved):

demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -2200,6 +2203,70 @@ def execute_tests(self):
...
+            self.build_context.logging_module.exception('Failed to get integrations configuration.')
+        if int(res_raw[1]) != 200:
+            self.build_context.logging_module.error(
+                f'Failed to get integrations configuration with status code: {res_raw[1]}')
+            return
+
+        all_configurations = res['configurations']
+        for instance in all_configurations:
+            if instance.get('id') == "Manual":
+                self.build_context.logging_module.info('"Manual" integration found in XSIAM instance.')
```
dantavori, line 2237:
```suggestion
                self.build_context.logging_module.info('Server is able to create manual alerts ("Manual" integration exists).')
```
less vauge,

demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -2200,6 +2203,70 @@ def execute_tests(self):
...
+        try:
+            res_raw_integration = demisto_client.generic_request_func(self=self.client, method='PUT',
+                                                                      path='/settings/integration-conf',
+                                                                      body=manual_integration)
+        except ApiException:
+            self.build_context.logging_module.exception('No "Manual" integration found in XSIAM instance. '
+                                                        'Please add it in order to create Manual Correlation Rule.')
+        if int(res_raw_integration[1]) != 200:
+            self.build_context.logging_module.error(
+                f'Failed to get integrations configuration with status code: {res_raw_integration[1]}')
```
dantavori, line 2268: let's split to a separate method

dantavori (approved):

PR #2240, opened 2022-07-12
Small fix to validate names in yml

Shellyber:
## Status
- [x] Ready

## Description
Small fix to validating yml names. Removed the `lower()` part - caused an upload issue and it didn't catch the following pack: https://github.com/demisto/content/pull/19962/files
(notice that the yml file has lower `i` and the rest `I`)


demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -1257,7 +1257,7 @@ def is_valid_integration_file_path(self) -> bool:
         elif integration_file != integrations_folder:
             valid_integration_file = integration_file.replace('-', '').replace('_', '')
 
-            if valid_integration_file.lower() != integrations_folder.lower():
+            if valid_integration_file != integrations_folder:
```
dorschw, line 1260: please rename the variable to clarify that it marks the file name

dorschw (approved):

PR #2272, opened 2022-08-01
add file type to error message when it is not supported

dorschw:
[CIAC-3671](https://jira-hq.paloaltonetworks.local/browse/CIAC-3671)


ilaner (approved):
amazing job!

PR #2276, opened 2022-08-03
fix bug in MP-106 validation

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
fixes an issue where an incident field that actually exist in build in fields is not recognized correctly by the validation due to case-sensitive

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -2909,10 +2909,9 @@ def get_invalid_incident_fields_from_mapper(
         if mapping_type == "mapping-outgoing":
             # for inc timer type: "field.StartDate, and for using filters: "simple": "".
             if simple := inc_info.get('simple'):
-                simple = simple.lower()
                 if '.' in simple:
                     simple = simple.split('.')[0]
-                if simple not in content_fields:
+                if simple not in content_fields and simple.lower() not in content_fields:
                     non_existent_fields.append(inc_name)
 
     return non_existent_fields
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -2909,6 +2909,7 @@ def get_invalid_incident_fields_from_mapper(
         if mapping_type == "mapping-outgoing":
             # for inc timer type: "field.StartDate, and for using filters: "simple": "".
             if simple := inc_info.get('simple'):
+                simple = simple.lower()
                 if '.' in simple:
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -2911,7 +2911,7 @@ def get_invalid_incident_fields_from_mapper(
             if simple := inc_info.get('simple'):
                 if '.' in simple:
                     simple = simple.split('.')[0]
-                if simple not in content_fields:
+                if simple not in content_fields and simple.lower() not in content_fields:
```
JudahSchwartz, line 2914: content_fields are already lower arent they? Whats the point of the first check?

JudahSchwartz: or if it isnt lowercase,
 `if any(simple.lower() == field.lower() for field in content_fields)`

GuyAfik: no, only some of them are lower (build in fields), those in the ID set can be upper case as well. 

see this line:
```
content_incident_fields = get_all_incident_and_indicator_fields_from_id_set(id_set_file, 'mapper') + [
            field.lower() for field in BUILT_IN_FIELDS
        ] + LAYOUT_AND_MAPPER_BUILT_IN_FIELDS
```

GuyAfik: so in case a built-in field was being checked and it wasn't lower, we would miss that

GuyAfik:
```
if any(simple.lower() == field.lower() for field in content_fields)
```
is also a good solution

JudahSchwartz: when is the case that we want to be case sensitive? using LAYOUT_AND_MAPPER_BUILT_IN_FIELDS only??

JudahSchwartz (approved):

demisto_sdk/commands/common/tools.py
GuyAfik: no, also when getting the available incident fields from the id_set

GuyAfik: i am not sure why we use field.lower() when getting the `build_in_fields`, it was being used in the old implementation, so i kept it that way

GuyAfik: image
from old implementation

PR #2290, opened 2022-08-22
Api module contribution fix

dansterenson:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-3799)

PR in contribution Management - should be merged after the SDK release:
https://github.com/demisto/contribution-management/pull/225

## Description
Now when contributing from the UI, and changing some ApiModule, the PR will contain the changes in both files.
PR for example: 
https://github.com/demisto/contribution-management/pull/225/files

As we spoke @yaakovi. The validation will fail since we do not update the release notes of all packs that depend on the updated ApiModule.


demisto_sdk/commands/split/ymlsplitter.py
```diff
@@ -367,7 +369,18 @@ def print_logs(self, log_msg: str, log_color: str) -> None:
         if self.logging:
             print_color(log_msg, log_color)
 
-    def replace_imported_code(self, script):
+    def split_api_module_for_contribution(self, lines: list, imported_line: str):
+        imported_line_arr = imported_line.split(' ')  # example: imported_line = from CorIRApiModule import *
+        updated_lines = lines[4: len(lines) - 3]  # ignore first 3 lines and last line.
```
yaakovi, line 374: The comment doesn't match the code. Which is the correct one?
```suggestion
        updated_lines = lines[4:-3]  # ignore first 3 lines and last line.
```

demisto_sdk/commands/split/ymlsplitter.py
```diff
@@ -74,6 +75,7 @@ def __init__(self, input: str, output: str = '', file_type: str = '', no_demisto
         self.autocreate_dir = not no_auto_create_dir
         with open(self.input, 'rb') as yml_file:
             self.yml_data = yaml.load(yml_file)
+        self.api_module_path: Optional[str] = None
```
yaakovi, line 78: Why do you need it?

demisto_sdk/commands/split/ymlsplitter.py
```diff
@@ -367,7 +369,18 @@ def print_logs(self, log_msg: str, log_color: str) -> None:
         if self.logging:
             print_color(log_msg, log_color)
 
-    def replace_imported_code(self, script):
+    def split_api_module_for_contribution(self, lines: list, imported_line: str):
```
yaakovi, line 372: Add docstring

yaakovi (approved):

demisto_sdk/commands/split/ymlsplitter.py
dansterenson: We use it in the contrib-mgmt repo, we want to git add the updated api module path and we get it from the sdk.

PR #2345, opened 2022-09-11
single `get_id`

dorschw:
remove redundancy


DeanArbel (approved):

demisto_sdk/commands/common/tools.py
```diff
@@ -1814,21 +1814,24 @@ def get_not_registered_tests(conf_json_tests: list, content_item_id: str, file_t
     return not_registered_tests
 
 
-def _get_file_id(file_type: str, file_content: Dict):
+def get_id(file_content: dict) -> str:
```
yaakovi, line 1817: Do not delete the old function to avoid dead locks (see recent import change fiesta).
Deprecate the old version for now (remove all usage in SDK internal code). after the next SDK release, replace usage in other repos and delete the old function.

PR #2348, opened 2022-09-12
use cache for installing in `pre-install-steps` CI step

dorschw:
reducing installation time from 2:30 minutes to 7 seconds (when cache is available)


ilaner (approved):
Good job, 95% faster

PR #2420, opened 2022-10-19
Make sure all integration tests are configured in test config

DinaMeylakh:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-4006)

## Description
fix a bug where integration paybook tests were mentioned in yml but not configured properly in conf.json

## Screenshots 
Example run
Screen Shot 2022-10-23 at 15 46 13

## Must have
- [x] Tests
- [x] Documentation


demisto_sdk/commands/common/hook_validations/content_entity_validator.py
```diff
@@ -277,13 +277,33 @@ def are_tests_registered_in_conf_json_file_or_yml_file(self, test_playbooks: lis
...
+            unconfigured_test_playbook_ids = []
+
+            if test_playbooks:
+                configured_tests = []
+                for test_playbook in test_playbooks:
+                    test_config_matches = []
+                    for test_config in conf_json_tests:
+                        test_config_matches.append(is_test_config_match(test_config,
+                                                                        test_playbook_id=test_playbook,
+                                                                        integration_id=content_item_id))
```
DinaMeylakh, line 290: Here we have a philosophical question,
if we have `integration1` with 
```
tests: 
 - some-test
 ```
 in the yml, but not in the conf.json.
 But, some other integration, `integration2` uses the same test `some-test` and `some-test` is configured in conf.json with it
 ```
 {
      "integrations": "integration2",
      "playbookID": "some-test"
 }
 ```
 Should the validation fail? After all, the test is present in the conf.json and will run on nightly. 
 I believe (and implemented) that it should fail, since collect_tests on integration change, only collects tests from config.json rather then the yml
 
 From collect_tests.py:
 ```
                 else:
                    # integration to test mapping available, and support level == xsoar (so - we run the tests)
                    tests = tuple(self.conf.integrations_to_tests[yml.id_])
                reason = CollectionReason.INTEGRATION_CHANGED
 ```

DinaMeylakh:
fix for failing validate files: https://github.com/demisto/content/pull/21835

demisto_sdk/commands/common/hook_validations/content_entity_validator.py
dorschw: yes

dorschw (approved):

PR #2437, opened 2022-10-26
Enforce pre-defined categories on Marketplace Packs validation

YuvHayun:
## Related Issues
fixes: [CIAC-4297](https://jira-hq.paloaltonetworks.local/browse/CIAC-4297)

## Description
Added a validator that validates that the yml category field and pack metadata categories field are according to the standard.


JudahSchwartz (approved):
Amazing

PR #2449, opened 2022-11-02
Content graph hotfixes

ilaner:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: link to the issue

## Description
* Fix multiprocessing of dumping packs
* Add marketplace properties to specific content items
* Fix dependency calculation for test playbooks

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -44,22 +42,35 @@ def __new__(
...
     class Config:
         arbitrary_types_allowed = True  # allows having custom classes for properties in model
         orm_mode = True  # allows using from_orm() method
         allow_population_by_field_name = True  # when loading from orm, ignores the aliases and uses the property name
 
+    def __getstate__(self):
+        """Needed to for the object to be pickled correctly (to use multiprocessing)"""
+        state = self.__dict__.copy()
+
+        # This object is cannot be pickled
```
DeanArbel, line 62:
```suggestion
        # This object cannot be pickled
```

demisto_sdk/commands/content_graph/objects/integration_script.py
```diff
@@ -28,7 +28,7 @@ def dump(self, dir: Path, marketplace: MarketplaceVersions) -> None:
         dir.mkdir(exist_ok=True, parents=True)
         try:
             IntegrationScriptUnifier(
-                input=str(self.path.parent), output=str(dir), marketplace=marketplace
+                input=str(self.path.parent), output=str(dir), marketplace=marketplace, force=True
```
DeanArbel, line 31: Why do we need to use force?

demisto_sdk/commands/content_graph/objects/repository.py
```diff
@@ -23,12 +23,15 @@ class ContentDTO(BaseModel):
     def dump(
         self, dir: DirectoryPath, marketplace: MarketplaceVersions, zip: bool = True
     ):
+        dir.mkdir(parents=True, exist_ok=True)
+        Pack.__fields__.pop("relationships_data")
```
DeanArbel, line 27: Please add a comment on why this is popped.

demisto_sdk/commands/content_graph/objects/integration.py
```diff
@@ -49,6 +61,7 @@ def set_commands(self):
             Command(
                 # the related to has to be a command
                 name=r.content_item.name,  # type: ignore[union-attr]
+                marketplaces=r.content_item.marketplaces,
```
DeanArbel, line 64: Why does a command need the `marketplaces`?
How does this work with common commands such as `reputation` commands that might have different versions (i.e. one with [XSOAR, v2], another with just [XSOAR] and a 3rd with just [v2]).

Shouldn't this be on the `edge`?

DeanArbel (approved):

demisto_sdk/commands/content_graph/objects/integration_script.py
ilaner: That's the way we currently use it in the original `create-content-artifacts`, I'll take a look what difference this makes

demisto_sdk/commands/content_graph/objects/repository.py
ilaner: Not needed, outdated

demisto_sdk/commands/content_graph/objects/integration.py
ilaner: The marketplace attribute of a command node is inherited of the implementations. For example, if `ip` command is used only by XSOAR integrations, the `marketplace` field will be `xsoar`. If it is used by both, we will receive ['xsoar', 'v2']. We can search for a command and the command model will provide us all the implementations of it.

In addition, we use the same command model for a specific command implementation, and that's why we have additional optional fields of `description` and `deprecated`. In this case, we should provide the `marketplaces` from the specific integration and not from the field, and I fixed it right now.

demisto_sdk/commands/content_graph/interface/neo4j/queries/dependencies.py
```diff
@@ -119,6 +119,8 @@ def create_depends_on_relationships(tx: Transaction) -> None:
         AND NOT pack_a.name IN {IGNORED_PACKS_IN_DEPENDENCY_CALC}
         AND NOT pack_b.name IN {IGNORED_PACKS_IN_DEPENDENCY_CALC}
         AND NOT b.node_id IN {IGNORED_CONTENT_ITEMS_IN_DEPENDENCY_CALC}
+        AND a.content_type <> "{ContentType.TEST_PLAYBOOK}"
+        AND b.content_type <> "{ContentType.TEST_PLAYBOOK}"
```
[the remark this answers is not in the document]
ilaner: This is currently not working, because `is_test` field doesn't exist when this is not test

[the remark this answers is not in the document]
dantavori: then it returns NULL if not exists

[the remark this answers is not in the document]
ilaner: It fails my unit test in this case.
I added `is_test` to `ContentItem` that's defaults to `False`. Is it OK by you?

Issue #1896, opened 2022-02-16
Demisto-SDK Support for Multiple XSOAR Servers


Pinger77:
This request is not to report a bug or issue, but rather an idea for a new feature. Currently the Demisto-SDK will only connect to 1 XSOAR server at a time without utilizing other methods to change DEMISTO_BASE_URL and DEMISTO_API_KEY. When running several of the SDK commands that interface with an XSOAR server (upload/download/etc), it would be nice to also be able to pick the server you are uploading to/downloading from/etc. With the creation of the CI/CD content pack, and the large role the Demisto-SDK plays in the process, having the ability to change XSOAR servers quickly would make it easier to deploy on different environments (Test/Prod) with the flip of a switch.

ChanochShayner:
@dorschw @bakatzir

tbailey25:
Would it be possible to define multiple environment variables and have an argument to the `demisto-sdk` commands to specify which one to use?

Eg.
```
$ cat .env
DEMISTO_BASE_URL_DEV=<url1>
DEMISTO_API_KEY_DEV=<key1>
DEMISTO_BASE_URL_PROD=<url2>
DEMISTO_API_KEY_PROD=<key2>
```
```
$ demisto-sdk download -i <input> -o <output> --server dev
```

And possibly prompt the user for which one to use if not specified via the cli?

dorschw:
Thank you for the suggestions @Pinger77 @ping@tbailey25, we'll consider adding this feature in future releases.

tbailey25:
If that's the direction that should be taken, then I would suggest a function that will clean the environment afterwards. If the .env files are sourced prior to the `demisto-sdk` command then you risk leaving the variables in the environment and then forgetting to source a different .env file on the next execution of `demisto-sdk`. This could lead to taking action against a XSOAR server that you did not intend.

```bash
demisto-sdk-dev () {
  source .env-dev
  demisto-sdk "$@"
  unset DEMISTO_BASE_URL
  unset DEMISTO_API_KEY
}
```

PaulBartruff:
> 

I use separate shell scripts that I source before using the sdk, and I often forget which one I left off at!  Also, when in vscode, it will pop open a new terminal and run before I have sourced my API and BASE, so I must then source and re-run the command.

PaulBartruff:
[...]

PR #2472, opened 2022-11-14
Add-multitenant-support

mattbibbydw:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: #1896

## Description
The goal of this PR is to add the base support for running subcommands on multi-tenant demisto servers. There have been several requests for this feature. Proposed solutions often revolve around wrapper scripts that can execute the `demisto-sdk` commands multiple times across a series of tenants. While this solution may work in some cases, this solution has several limitations:

1. Increased complexity, as both the wrapper and the SDK will need to be maintained.
2. Limited support for subcommands that can be executed on either tenants or the main server. 
3. Limited support multi-tenant specific subcommands.

This PR offers an alternative. Instead of using an external wrapper to manage execution, define a single set of options for managing multi-tenancy execution control that can be passed to all other subcommands. Each subcommand will still be responsible for determining how it should execute on a multi-tenant environment, but using the context and methods of the passable config object will significantly reduce the complexity of these operations.

### Managing demisto_client Environment Variables

Most interactions with the Demisto server is performed using the `demisto_client` library. This uses the `DEMISTO_BASE_URL` and `DEMISTO_API_KEY` variables to determine the location of the server and a means to authenticate. In the case of multi-tenant executions, these variables must be changed for each tenant so the `demisto_client` will target the tenant server instead of the main server.

This PR solves this problem via the `DemistoSDK.load_tenant_env()` method. This method will temporarily override the existing envvars with whatever envvars are stored within a `.env-{tenant_name}` file located within the `.envs/` directory. Since this method is a context manager, this override only applies to the code underneath the scope of the context manager. Once the context manager is exited, the variables stored within the base `.env` file are restored. Here's an example of how this would be used for the `run` subcommand:
```python
# Assume at this point the `DEMISTO_BASE_URL` and `DEMISTO_API_KEY` variables are loaded from the base `.env` file
results = {}
for tenant in config.tenants:  # Where `config.tenants == ["foo", "bar", "baz"]
    with config.load_tenant_env(tenant):
        # Once entered, the `.envs/.env-{tenant}` file will override existing envvars
        runner = Runner(**kwargs)
        results[tenant] = runner.run()
    # Once exited, the envvars within the base `.env` file be restored
return results
```

This approach is similar to the solution described in #1896, without the risk of stale environment variables.

### Determining the demisto server type and tenants

This PR uses the `get_demisto_tenants()` function in order to determine both the server type and the tenants that belong to the server. It does this by querying the `/accounts` endpoint of the host configured for the `demisto_client`. However, while it uses the `demisto_client` to determine the host of the request, it does **NOT** use the `demisto_client` to perform the REST call. This is due to the fact that `demisto_client` does not handle responses from this endpoint well unless they are sent to the main server of a multi-tenant server. Requests sent to a single-server deployment return a generic HTTP error and requests sent to a tenant of a multi-tenant server result in a timeout. Since no other endpoint exists for determining the server type of a demisto server, I opted to perform this API call using the `requests` library instead.

## Screenshots
N/A

## Must have
- [x] Tests
- [ ] Documentation


dorschw:
Hi, thanks for this contribution! We truly appreciate it :)  
We'll assign a developer to review it, and keep you updated here.

gal-forer:
@mattbibbydw Hi, were so sorry we missed your contribution, if you're still willing we'd love to finish the review process, can you please fix the conflicts so we'll be able to continue ?

gal-forer:
Hi @mattbibbydw , 
Thank you for your contribution. 
I have not received a response from you in the last two weeks, so I will close your PR now.
Please feel free to re-open it when you are available to continue. 
Thanks again.

PR #2497, opened 2022-11-22
pb nexttasks validations

tkatzir:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-3383


demisto_sdk/commands/common/hook_validations/playbook.py
```diff
@@ -314,19 +315,79 @@ def is_script_condition_branches_handled(self, task: Dict) -> bool:
...
+        Args:
+            task (dict): task json loaded from a yaml
+
+        Return:
+            bool. if the task's next tasks have more than just the default option.
+        """
+        is_default_not_only_condition_res: bool = True
+        next_tasks: Dict = task.get('nexttasks', {})
+
+        if len(next_tasks) == 1:
```
idovandijk, line 360: Actually, it doesn't matter if the length of next_tasks is 1. The conditional task may have 3 next_tasks with "Else" label, which is also a problem. The check should verify whether the next_tasks have labels other than "Else". If not - error.

idovandijk (commented):
Great validations, please see my comment

demisto_sdk/commands/common/hook_validations/playbook.py
tkatzir: https://github.com/demisto/demisto-sdk/pull/2497/commits/5e676671ffcd634df2b18811c0f217738525ed04

idovandijk: I phrased myself completely wrong. I meant:
Doesn't matter the amount of next_tasks. If ALL of the next_tasks (whether 1 out of 1, or 3 out of 3) lead to "Else" - it should be an error, because we can't have paths only for "else"

idovandijk (approved):
Looks good!

PR #2508, opened 2022-11-27
adding default path to xpanse id set

daryakoval:


demisto_sdk/commands/create_id_set/create_id_set.py
```diff
@@ -85,8 +86,12 @@ def create_command_to_implemented_integration_map(self):
 
     def save_id_set(self):
         if not self.output:
-            self.output = MP_V2_ID_SET_PATH if self.marketplace == MarketplaceVersions.MarketplaceV2.value \
-                else DEFAULT_ID_SET_PATH
+            if self.marketplace == MarketplaceVersions.MarketplaceV2:
+                self.output = MP_V2_ID_SET_PATH
+                if self.marketplace == MarketplaceVersions.XPANSE:
```
dorschw, line 91: wrong indentation?

demisto_sdk/commands/create_id_set/create_id_set.py
```diff
@@ -85,8 +86,12 @@ def create_command_to_implemented_integration_map(self):
...
+                    self.output = XPANSE_ID_SET_PATH
+                else:
+                    self.output = DEFAULT_ID_SET_PATH
```
dorschw, line 94: I'd replace it with self.output = {MarketplaceVersions.MarketplaceV2: MP_V2_ID_SET_PATH, MarketplaceVersions.XPANSE:XPANSE_ID_SET_PATH}.get(self.marketplace, DEFAULT_ID_SET_PATH)

demisto_sdk/commands/create_id_set/create_id_set.py
daryakoval: thanks, your reviews are the best

demisto_sdk/commands/create_id_set/create_id_set.py
daryakoval: It will break all the places we use self.output... mypy will not accept it.

Shellyber (approved):

Issue #2318, opened 2022-09-05
exists on disk, but not in 'origin/master'  error with absolute paths


ennozdd:
```
$ demisto-sdk validate -i Packs/mysweetpack/
You are using demisto-sdk 1.7.4.

Running in a private repository

================= Validating pack /home/ennozdd/ennozdd/Packs/mysweetpack =================

Validating /home/ennozdd/ennozdd/Packs/mysweetpack unique pack files
Got an error while trying to connect to git - Cmd('git') failed due to: exit code(128)
  cmdline: git show origin/master:/home/ennozdd/ennozdd/Packs/mysweetpack/pack_metadata.json
    stderr: 'fatal: Path '/home/ennozdd/ennozdd/Packs/mysweetpack/pack_metadata.json' exists on disk, but not in 'origin/master'.'
```

Even though the pack's location is given as a relative path, demisto-sdk chooses to go with its absolute path. Absolute paths seem to not work with git show it should use the relative form from the repo root. 

Example:
$ git show origin/master:/home/ennozdd/ennozdd/Packs/mysweetpack/pack_metadata.json
fatal: Path '/home/ennozdd/ennozdd/Packs/mysweetpack/pack_metadata.json' exists on disk, but not in 'origin/master'.
$ git show origin/master:Packs/mysweetpack/pack_metadata.json

https://stackoverflow.com/a/65783920

dorschw:
Hi @ennozdd, thanks for reporting this issue, this is tracked internally as [CIAC-4000](https://jira-hq.paloaltonetworks.local/browse/CIAC-4000)

upstart-swiss:
Also encountering this here and in CI. Any recommendation of how to better run the command to not have this error?

DinaMeylakh:
@ennozdd @upstart-swiss This is actually supposed to be handled by the line
```
relative_file_path = os.path.relpath(full_file_path, self.git_path())
``` 
where the function is defined by

```    
def git_path(self) -> str:
        git_path = self.repo.git.rev_parse('--show-toplevel')
        return git_path.replace('\n', '')
```

@ennozdd Could you run `git rev-parse --show-toplevel` and see if it returns '/home/ennozdd/ennozdd/' as expected?

Also, are you getting a message like the following?
```
Could not get local remote file because of: ...
Searching the remote file content with the API.
```
If so, it will be helpful if you could paste the message here.

[...]

PR #2513, opened 2022-11-28
nodisk error fixed

ennozdd:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://github.com/demisto/demisto-sdk/issues/2318

## Description
get_master_private_repo_meta_file function should either take relative path as argument or it should extract the relative form from the absolute path and work with it.

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


demisto_sdk/commands/common/hook_validations/pack_unique_files.py
```diff
@@ -736,7 +737,7 @@ def get_master_private_repo_meta_file(self, metadata_file_path: str):
                 click.secho("Running on master branch - skipping price change validation", fg="yellow")
             return None
         try:
-            old_meta_file_content = current_repo.git.show(f'{self.prev_ver}:{metadata_file_path}')
+            old_meta_file_content = current_repo.git.show(f'{self.prev_ver}:{relative_file_path}')
```
DinaMeylakh, line 740: Let's be consistent here and get the file the same way we do everywhere else
```suggestion
            old_meta_file_content = get_remote_file(metadata_file_path, tag=self.prev_ver)
```

DinaMeylakh (changes requested):
Great Catch & Thanks for the contribution!
1. See my inline comments
2. Add a unit-test to `demisto_sdk/commands/common/tests/pack_unique_files_test.py`
3. Add a brief description of what you changed to the `Unreleased` section in the  `CHANGELOG.md` file

PR #2519, opened 2022-11-29
added cpu_count function

esharf:
## Status
- [x] Ready

## Related Issues
fixes: build resources issues

## Description
since we are running in a single machine we can’t use "all" the available cpus so we need to be able to decrease it.

## Must have
- [x] Tests

### DIFF
--- a/demisto_sdk/commands/common/cpu_count.py
+++ b/demisto_sdk/commands/common/cpu_count.py
@@ -1,12 +1,12 @@
 import os
 
-MAX_DEMISTO_SDK_THREADS = 'MAX_DEMISTO_SDK_THREADS'
+DEMISTO_SDK_MAX_THREADS = 'DEMISTO_SDK_MAX_THREADS'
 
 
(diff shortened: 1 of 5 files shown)


demisto_sdk/commands/common/cpu_count.py
```diff
@@ -0,0 +1,14 @@
...
+MAX_DEMISTO_SDK_THREADS = 'MAX_DEMISTO_SDK_THREADS'
+
+
+def cpu_count():
+    max_available_cpus = os.cpu_count() or 1
+    try:
+        max_allowed_cpus = int(os.environ[MAX_DEMISTO_SDK_THREADS])
+        requsted_cpus = min(max_allowed_cpus, max_available_cpus)
+        return max(1, requsted_cpus)
+    except Exception:
```
dorschw, line 12: Please specify the exceptions

dorschw (approved):

PR #2543, opened 2022-12-06
Support from/to version in XSIAM triggers

yaakovpraisler:
## Status
- [x] Ready

### DIFF
--- a/demisto_sdk/commands/common/content/objects/pack_objects/trigger/trigger.py
+++ b/demisto_sdk/commands/common/content/objects/pack_objects/trigger/trigger.py
@@ -41,7 +41,7 @@ class Trigger(JSONContentObject):
         created_files.extend(super().dump(dest_dir=dest_dir))
         new_file_path = created_files[0]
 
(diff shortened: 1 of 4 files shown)


anara123 (approved):

demisto_sdk/commands/common/content/objects/pack_objects/trigger/trigger.py
```diff
@@ -33,12 +34,36 @@ def type(self):
...
+
         created_files: List[Path] = []
         created_files.extend(super().dump(dest_dir=dest_dir))
         new_file_path = created_files[0]
-        if new_file_path.name.startswith('external-'):
-            copy_to_path = str(new_file_path).replace('external-', '')
+
+        if Version(self.get('fromversion', '0.0.0')) >= Version('6.10.0'):
+            # export XSIAM 1.3 items only with the external prefix
+            if not new_file_path.name.startswith('external-'):
```
dorschw, line 46: what if this condition does not hold?

PR #2544, opened 2022-12-07
Update README.md

Shellyber:
## Status
- [x] Ready

## Description
Updated find-dependencies command readme with missing flags.


Shellyber:
@tkatzir No. This is a command we usually use internally so it won't be on the main readme.

PR #2545, opened 2022-12-07
Fix Invocation of 'init-test-data' from 'test' Command

avidan-H:
## Status
- [x] Ready

## Description
When running the `demisto-sdk modeling-rules test` command on a modeling rule directory, if the associated testdata file does not exist, the invocation of `init-test-data` from within `test` did not work properly (this assumes that the original command is running in `--interactive` mode and that the user answers with the affirmative to the prompts asking the user if they'd like to generate a testdata file). The invocation of `init-test-data` would raise an exception _and_ when applying the changes to make it work, the log messages from the invocation were duplicated.
- Adds a unit test to check the above-described scenario.
- Adds the changes to invoke `init-test-data` properly from `demisto-sdk modeling-rules test` command.
- Adds logic to `init-test-data` to not call `setup_rich_logging` when invoked from the `test` command since rich logging has already been setup.

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/test_content/test_modeling_rule/tests/test_modeling_rule_test.py
+++ b/demisto_sdk/commands/test_content/test_modeling_rule/tests/test_modeling_rule_test.py
@@ -804,3 +804,66 @@ class TestTheTestModelingRuleCommandMultipleRules:
                     assert 'Mappings validated successfully' in result.stdout
         except typer.Exit:
             assert False, "No exception should be raised in this scenario."
+
+
+class TestTheTestModelingRuleCommandInteractive:
+
+    def test_no_testdata_file_exists(self, repo, monkeypatch, mocker, enable_logging):
+        """
+        Given:
+            - A modeling rule with no test data file.
+
+        When:
+            - The command is run in interactive mode.
+
+        Then:
+            - Verify we get a message saying the test data file does not exist.
+            - Ensure we are prompted  to create a test data file.
+            - The command returns with a non-zero exit code.
+            - Ensure the test data file was created.
+            - Ensure that the log output from creating the testdata file is not duplicated.
+        """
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/test_content/test_modeling_rule/init_test_data.py
```diff
@@ -67,7 +68,11 @@ def init_test_data(
     """
     Initialize or update a test data file for a modeling rule
     """
-    setup_rich_logging(verbosity, quiet, log_path, log_file_name)
+    # don't invoke rich logging if this is invoked from the "demisto-sdk modeling-rules test" command
+    # since it's already setup there
+    if not (ctx.parent and ctx.parent.info_name == 'test'):
```
tkatzir, line 73:
```suggestion
    if not (ctx.parent and ctx.parent.info_name and ctx.parent.info_name.lower() == 'test'):
```

demisto_sdk/commands/test_content/test_modeling_rule/test_modeling_rule.py
```diff
@@ -332,11 +333,25 @@ def validate_modeling_rule(
...
-                init_td = app.command()(init_test_data.init_test_data)
                 events_count = typer.prompt(
                     'For how many events would you like to generate templates?', type=int, default=1, show_default=True
                 )
-                init_td([mrule_dir], events_count)
+
+                from demisto_sdk.commands.test_content.test_modeling_rule.init_test_data import app as init_td_app
+
+                # the init-test-data typer application only has one command, so we can just get the first one
+                init_td_cmd_info = init_td_app.registered_commands[0]
```
tkatzir, line 343: No need to check that `len(init_td_app.registered_commands) > 0` ?

tkatzir (commented):
Good!
Some comments and questions.

tkatzir (approved):
Great!

PR #2557, opened 2022-12-11
add codeql

dorschw:


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,41 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+  schedule:
+    - cron: "1 14 * * 4"
```
tkatzir: Every Thursday at 14:01? Why right before the weekend?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,41 @@
...
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+  schedule:
+    - cron: "1 14 * * 4"
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
```
tkatzir, line 14: Do we **need** ubuntu?

tkatzir (commented):
Great!

.github/workflows/codeql.yml
dorschw: That's the suggested default. I don't mind using another.

tkatzir: Try alpine or debian.

PR #2612, opened 2022-12-25
Add linting api modules dependencies

DinaMeylakh:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: link to the issue

## Description

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


demisto_sdk/commands/lint/lint_manager.py
```diff
@@ -74,24 +76,36 @@ def __init__(self, input: str, git: bool, all_packs: bool, quiet: bool, verbose:
...
         self.linters_error_list: list = []
 
+    def _get_api_module_dependent_items(self) -> list:
+        changed_api_modules = {pkg.name for pkg in self._pkgs if API_MODULES_PACK in pkg.parts}
+        if changed_api_modules:
+            dependent_items = []
+            for changed_api_module in changed_api_modules:
+                print(f'Checking for packages dependent on the modified API module {changed_api_module}...')
+                with Neo4jContentGraphInterface() as graph:
+                    api_module_node = graph.search(object_id=changed_api_module)[0]
```
dantavori, line 96: I think we can't assume there is a built graph (locally / on circle)
@ilaner let's discuss this tomorrow - maybe we need an inner check for this

dantavori (commented):
looks good!
need to discuss also if we want to temporarily keep the support in the ID set (so we can merge but don't use it on circle until we migrate to the graph there)

demisto_sdk/commands/content_graph/objects/script.py
```diff
@@ -25,3 +25,11 @@ def prepare_for_upload(self, marketplace: MarketplaceVersions = MarketplaceVersi
             data['nativeImage'] = supported_native_images
 
         return data
+
+    @property
+    def imported_by(self) -> list:
```
ilaner, line 30:
```suggestion
    def imported_by(self) -> List[Integration]:
```

demisto_sdk/commands/lint/lint_manager.py
ilaner: anyway, please wrap it in a try except clause, and handle a case which the object_id is not found the graph

DinaMeylakh:
@dantavori @ilaner we don't use the id set right now in circle anyway for lint (we don't create it) & I've tested the feature locally with graph and it worked :)  so I don't see a point in keeping support.

demisto_sdk/commands/content_graph/objects/script.py
DinaMeylakh: Changed to `List[BaseContent]` instead because that is what RelationshipData keeps:

``` 
class RelationshipData(BaseModel):
    relationship_type: RelationshipType
    source: BaseContent
    target: BaseContent

    # this is the attribute we're interested in when querying
    content_item: BaseContent
```

demisto_sdk/commands/lint/lint_manager.py
DinaMeylakh: @dantavori @ilaner Why can't we assume the graph is built?

I think this step should fail if the graph is not updated/created. That way we won't pass failing modules that import the api modules.

I raised an exception for an out of date graph file & I don't see sense in try excepting the graph errors just to raise them again, unless you say they are unreadable in which case I can add a print indicating the errors came from trying to interact with the content graph

DinaMeylakh:
@dantavori @ilaner I added `update_content_graph`, please review.

demisto_sdk/commands/lint/tests/linter_manager_test.py
```diff
@@ -763,3 +766,193 @@ def test_create_json_output_xsoar_linter(repo, mocker):
...
+    def __init__(self, path):
+        self.path = path
+
+
+class NodeMock:
+    def __init__(self, imported_by):
+        self.imported_by = imported_by
+
+
+@pytest.mark.parametrize(
```
dantavori, line 781: great tests 🥇

demisto_sdk/commands/lint/lint_manager.py
```diff
@@ -109,30 +114,56 @@ def __init__(
...
+        if changed_api_modules:
+            dependent_items = []
+            for changed_api_module in changed_api_modules:
+                print(
+                    f"Checking for packages dependent on the modified API module {changed_api_module}..."
+                )
+
+                with Neo4jContentGraphInterface() as graph:
+                    print("Updating graph...")
+                    update_content_graph(graph, use_git=True, dependencies=True)
```
dantavori, line 140: I'm okay to wait with merging until we implement automatic detection of the need to run update_content_graph() internally in the interface
@ilaner fyi

dantavori (approved):
awesome 👍🏽

PR #2619, opened 2022-12-26
Validate deprecation flag

dorschw:
- If an integration display name ends with `(Deprecated)`, make sure it has `deprecated: true` in its yaml (IN157)
- If an integration description matches the deprecation regex, make sure it has `deprecated: true` in its yaml (IN158)
- clean & improve unit test code

fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-4650


demisto_sdk/commands/common/errors.py
```diff
@@ -988,6 +990,16 @@ def invalid_hidden_attribute_for_param(param_name: str, invalid_value: str):
     def invalid_deprecated_integration_display_name():
         return 'The display_name (display) of all deprecated integrations should end with (Deprecated)".'
 
+    @staticmethod
+    @error_code_decorator
+    def invalid_integration_deprecation__only_display_name_suffix():
+        return 'All integrations whose display_names end with `(Deprecated)` must have `deprecated:true`.'
```
yaakovpraisler, line 996: Add a suggestion to run `format` with the flag `-d, --deprecate` to fully deprecate an content item

dorschw: done, thanks

yuvalbenshalom (approved):

Issue #2622, opened 2022-12-26
cannot download layouts with `detailsV2=None`


ennozdd:
```
$ demisto-sdk download  -i 'My indicator layout'-o Packs/mypack

You are using demisto-sdk 1.8.1.
Exception raised when fetching custom content:
'NoneType' object has no attribute 'get'
```

I am getting the error above when trying to download a custom indicator layout.

ennozdd:
```
{
	"cacheVersn": 0,
	"close": null,
	"definitionId": "",
	"description": "",
	"detached": false,
	"details": null,
	"detailsV2": null,
	"edit": null,
	"fromServerVersion": "",
	"group": "indicator",
	"id": "ffba70f8-1649-45da-8cd9-fc796a3aeff8",
	"indicatorsDetails": {
		"TypeName": "",
		"tabs": null
	},
	"indicatorsQuickView": {
		"TypeName": "",
		"tabs": [
			{
				"id": "indicator-quick-view-info",
				"name": "Info",
				"sections": [
					{
						"h": 1,
						"hideName": true,
						"i": "indicator-quick-view-info-expirationStatus",
						"maxW": 3,
						"moved": false,
						"name": "Expiration Status",
						"static": false,
						"type": "expirationStatus",
						"w": 1,
						"x": 0,
						"y": 0
					},
					{
						"h": 2,
						"i": "indicator-quick-view-info-relatedIncidents",
						"maxW": 3,
						"moved": false,
						"name": "Related Incidents",
						"static": false,
						"type": "relatedIncidents",
						"w": 1,
						"x": 0,
						"y": 8
					},
					{
						"h": 3,
						"i": "indicator-quick-view-info-comments",
						"maxW": 3,
						"moved": false,
						"name": "Comments",
						"static": false,
						"type": "comments",
						"w": 1,
						"x": 0,
						"y": 10
					},
					{
						"displayType": "ROW",
						"h": 2,
						"hideName": false,
						"i": "indicator-quick-view-info-34c55dc0-ba19-11ea-96e1-0b84fdedf11e",
						"items": [
							{
								"endCol": 2,
								"fieldId": "timestamp",
								"height": 22,
								"id": "38200c30-84ee-11ec-a589-e73aadc98c9a",
								"index": 0,
								"sectionItemType": "field",
								"startCol": 0
							},
							{
								"endCol": 2,
								"fieldId": "techname",
								"height": 22,
[...]

PR #2623, opened 2022-12-26
fix download layout bug

dorschw:
fixes: https://github.com/demisto/demisto-sdk/issues/2622

### DIFF
--- a/demisto_sdk/commands/download/downloader.py
+++ b/demisto_sdk/commands/download/downloader.py
@@ -254,7 +254,7 @@ class Downloader:
         file_json_object = json.loads(layout_string)
         layout_name = file_json_object.get('name')
         if layout_name and (layout_name in self.input_files or self.all_custom_content):
-            for tab in file_json_object.get('detailsV2', {}).get('tabs', ()):
+            for tab in (file_json_object.get('detailsV2') or {}).get('tabs', ()):
                 for section in tab.get('sections', ()):
                     for item in section.get('items', ()):
                         script_id = item.get('scriptId')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 ## Unreleased
 * Fixed an issue where modeling rules with arbitrary whitespace characters were not parsed correctly.
 * Added support for the **nativeimage** key for an integration/script in the **prepare-content** command.
-
+* Fixed an issue where downloading content layouts with `detailsV2=None` resulted in an error.
 ## 1.8.1
 * Fixed an issue where **format** created duplicate configuration parameters.
 * Added hidden properties to integration command argument and script argument.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Fixed an issue where modeling rules with arbitrary whitespace characters were not parsed correctly.
 * Added support for the **nativeimage** key for an integration/script in the **prepare-content** command.
 * Fixed an issue where downloading content layouts with `detailsV2=None` resulted in an error.
+
 ## 1.8.1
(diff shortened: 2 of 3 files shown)


demisto_sdk/commands/download/downloader.py
```diff
@@ -254,7 +254,7 @@ def handle_layout(self, layout_string, scripts_mapper):
         file_json_object = json.loads(layout_string)
         layout_name = file_json_object.get('name')
         if layout_name and (layout_name in self.input_files or self.all_custom_content):
-            for tab in file_json_object.get('detailsV2', {}).get('tabs', ()):
+            for tab in (file_json_object.get('detailsV2') or {}).get('tabs', ()):
```
ennozdd, line 257: I replaced detailsV2 with indicatorsQuickView and it downloaded the indicator layout.

ennozdd:
```
    def handle_layout(self, layout_string, scripts_mapper):
        # In case the layout uses a custom script, replace the id value of the script with its name
        file_json_object = json.loads(layout_string)
        layout_name = file_json_object.get('name')
        if layout_name and (layout_name in self.input_files or self.all_custom_content):
            json_objects_with_tabs = []
            for obj in file_json_object:
                if isinstance(obj, dict) and obj.get('tabs', ''):
                    json_objects_with_tabs.append(obj)
                    
            for obj in json_objects_with_tabs:
                for tab in file_json_object.get(obj, {}).get('tabs', ()):
                    for section in tab.get('sections', ()):
                        for item in section.get('items', ()):
                            script_id = item.get('scriptId')
                            if script_id and script_id in scripts_mapper:
                                layout_string = layout_string.replace(script_id, scripts_mapper[script_id])

        return layout_string
```
Does it make sense to go through all the objects that contains tabs?

ennozdd:
```
    def handle_layout(self, layout_string, scripts_mapper):
        # In case the layout uses a custom script, replace the id value of the script with its name
        file_json_object = json.loads(layout_string)
        layout_name = file_json_object.get('name')
        if layout_name and (layout_name in self.input_files or self.all_custom_content):

            json_objects_with_tabs = []
            for obj in file_json_object:
                if isinstance(file_json_object[obj], dict):
                    if file_json_object[obj].get('tabs', None):
                        json_objects_with_tabs.append(obj)
                    
            for obj in json_objects_with_tabs:
                for tab in file_json_object.get(obj, {}).get('tabs', ()):
                    for section in tab.get('sections', ()):
                        for item in section.get('items', ()):
                            script_id = item.get('scriptId')
                            if script_id and script_id in scripts_mapper:
                                layout_string = layout_string.replace(script_id, scripts_mapper[script_id])

        return layout_string

```

Does it make sense to go through all the json objects that contain a tab?
EDIT: modified the code a bit to get it working.

dorschw: What do you mean? Why change `detailsV2`to `indicatorsQuickView`?

ennozdd: if there is a script associated to a layout this function seems to rename the scriptId of it. Not only detailsV2 but also indicatorsQuickView and indicatorsDetails hold the same type of content in the yml but for indicator layouts. That's just my observation. so if they are practically the same, the function should also look into these objects for scriptIds.

for the first comment, I was trying to download an indicator layout and it failed on detailsV2 but it looked like this function was actually looking for indicatorsQuickView.

dorschw: I see, thanks for the explanation. This looks valid but requires proper unit tests to merge, we'll do it.
I will merge this fix for now anyway, to allow downloading cases with a `None` value until `indicatorsQuickView` is handled.

PR #2625, opened 2022-12-26
doc_review fixes

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes:[CIAC-4961](https://jira-hq.paloaltonetworks.local/browse/CIAC-4961)

## Description
Fixed an issue where doc-review failed when reviewing command name in some cases.


demisto_sdk/commands/doc_reviewer/doc_reviewer.py
```diff
@@ -319,6 +319,8 @@ def check_word(self, word):
         word = self.remove_punctuation(word)
         sub_words = []
         if '-' in word:
+            if not self.spellchecker.unknown([word]):
```
JudahSchwartz, line 322: shouldnt we split by - and check each?

YuvHayun: that's exactly the issue..
because we split the command name by '-' we get the error..
if we first check to ensure it's not a command name then there's a point to split it.

JudahSchwartz: not sure I understand. Lets discuss

JudahSchwartz (approved):

PR #2626, opened 2022-12-26
Update the IN145 error

mmhw:
## Status
- [ ] In Progress
- [x] Ready

## Description
Changed the **validate** command to fail on the IN145 error code only when the parameter with type 4 is not hidden.

## Must have
- [x] Tests


bziser (approved):
Great!

PR #2651, opened 2023-01-03
Update the IN124 error

mmhw:
## Status
- [ ] In Progress
- [x] Ready

## Description
As a result of [this PR](https://github.com/demisto/demisto-sdk/pull/2637), I later realized that it should support other types of hidden parameters (not just type 4).

## Must have
- [x] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/integration.py
+++ b/demisto_sdk/commands/common/hook_validations/integration.py
@@ -1463,9 +1463,9 @@ class IntegrationValidator(ContentEntityValidator):
 
         return True
 
-    def _is_type4_replaced_by_type9(self, display_name: str) -> bool:
+    def _is_replaced_by_type9(self, display_name: str) -> bool:
         """
-        This function is used to check the case where a parameter is hidden but because is type 4 and is replaced by a type 9 parameter.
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/hook_validations/integration.py
```diff
@@ -1522,7 +1521,7 @@ def is_str_bool(input_: str):
             ) == set(MarketplaceVersions)
 
             if invalid_bool or hidden_in_all_marketplaces:
-                if type_ == 4 and self._is_type4_replaced_by_type9(display_name):
```
dorschw, line 1525: Please allow only types (0, 4, 14)

demisto_sdk/commands/common/tests/integration_test.py
```diff
@@ -2181,7 +2181,7 @@ def test_additional_info_contained(self):
     }
     HIDDEN_TRUE_BUT_REPLACED_USER = {
         "configuration": [
-            {"type": 4, "display": "Username", "hidden": True},
+            {"type": 0, "display": "Username", "hidden": True},
```
dorschw, line 2184: please check all the allowed, as well as one that isn't allowed (which should be considered erroneous)

dorschw (approved):

PR #2659, opened 2023-01-05
handle docker_native_image_config.json file not found

GuyAfik:
## Description
Handles issue where docker_native_image_config was not found in 1.8.2 when running the upload command causing it to fail if not updated from content master

edit: managed to test that this PR fix works on a content branch where i deleted the `docker_native_image_config.json`file.

### DIFF
--- a/demisto_sdk/commands/content_graph/objects/integration_script.py
+++ b/demisto_sdk/commands/content_graph/objects/integration_script.py
@@ -39,13 +39,13 @@ class IntegrationScript(ContentItem):
     def get_supported_native_images(
         self, marketplace: MarketplaceVersions, ignore_native_image: bool = False
     ) -> List[str]:
-        try:
-            if marketplace == MarketplaceVersions.XSOAR and not ignore_native_image:
+        if marketplace == MarketplaceVersions.XSOAR and not ignore_native_image:
+            try:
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/content_graph/objects/integration_script.py
```diff
@@ -40,9 +43,13 @@ def get_supported_native_images(
...
-                native_image_config=file_to_native_image_config(),
-            ).get_supported_native_image_versions(get_raw_version=True)
+            try:
+                return ScriptIntegrationSupportedNativeImages(
+                    _id=self.object_id,
+                    docker_image=self.docker_image,
+                    native_image_config=file_to_native_image_config(),
+                ).get_supported_native_image_versions(get_raw_version=True)
+            except FileNotFoundError:
+                logger.debug(f"The {NATIVE_IMAGE_FILE_NAME} file could not be found.")
```
dorschw, line 53: this _could_ be raised with another file, hiding the real cause for the failure. Please replace the `try-except` with `if not Path(...).exists(): return []`

GuyAfik: done

dorschw (approved):

PR #2661, opened 2023-01-05
Update release_guide.md

dorschw:


docs/release_guide.md
```diff
@@ -39,13 +40,24 @@ The scope of the upgrade is decided by the SDK owner on each release, based on t
 ## Triggering nightlies manually
 The following should _only_ be done when new PRs were mergerd between the nightly trigger and the release start. This is seldom required.
 
+**SDK nightly**
+0) Run the following from content `./Utils/trigger_nightly_sdk_build.sh -b master -sr <sdk branch name> -ch dummy-channel -g -ct <gitlab_token>`
+
 1) Under the content repo, change the `demisto-sdk` dependency under [pyproject.toml]([url](https://github.com/demisto/content/blob/master/pyproject.toml)) to `demisto_sdk = {git = "https://github.com/demisto/demisto-sdk.git", rev =<commit hash here>}`
```
thefrieddan1, line 48: After this step. step 2 failed because pipfile.lock file was inconsistent.
I think we should add here pipfile lock step. what do you think?

thefrieddan1 (approved):

docs/release_guide.md
dorschw: I added [this](https://github.com/demisto/demisto-sdk/pull/2661/files#diff-f58e165ff22d5fe8ece2cb586ff9daab82549caf67b984ae79fef463f2aa6923R12)

PR #2672, opened 2023-01-09
fix sdk get identifier

esharf:
## Status
- [x] Ready


demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -1540,7 +1540,7 @@ def create_integration_instance(
                     credentials = params[key]
                     param_value = {
                         "credential": "",
-                        "identifier": credentials["identifier"],
+                        "identifier": credentials.get("identifier", ""),
```
dorschw, line 1543:
```suggestion
                        "identifier": credentials.get("identifier", ""),  # when there's only an API key
```

dorschw (approved):

PR #2692, opened 2023-01-16
Add Timeframe to Queries

avidan-H:
## Status
- [x] Ready

## Related Issues
fixes: link to the issue

## Description
- Adds an error message notifying of potential parsing rule-related errors.
- When testdata is pushed and there are parsing rules for its dataset that modify the _time field, it can mean that the event will be old in XSIAM. So adding a timeframe of 10 years to our xql queries will account for when the testdata event time is old.
- When interactively running test, sleep one second after doing an upload of the pack to allow time for the pack installation.

## Must have
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/demisto_sdk/commands/test_content/test_modeling_rule/test_modeling_rule.py
+++ b/demisto_sdk/commands/test_content/test_modeling_rule/test_modeling_rule.py
@@ -154,8 +154,8 @@ def generate_xql_query(rule: SingleModelingRule, test_data_event_ids: List[str])
         [f'"{td_event_id}"' for td_event_id in test_data_event_ids]
     )
     query = (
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/test_content/test_modeling_rule/test_modeling_rule.py
```diff
@@ -335,6 +344,8 @@ def verify_pack_exists_on_tenant(
                         f"[red]Failed to upload pack {containing_pack_id} to tenant[/red]",
                         extra={"markup": True},
                     )
+                # wait for pack to finish installing
+                sleep(1)
```
tkatzir, line 348: Upload can take (a lot) more than 1 sec

tkatzir (approved):
Good!
A few comments.

PR #2698, opened 2023-01-17
Validate: make sure that all validations have the ignore-list from .pack-ignore

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-5143)

## Description
This PR fixes an issue where some validations are not ignored even when they can be ignored in the `.pack-ignore`

## Must have
- [x] Tests


GuyAfik:
failed integration test: https://app.circleci.com/pipelines/github/demisto/demisto-sdk/20943/workflows/6d1b1152-efbd-4dfc-ad3d-8792a3b9b5e0/jobs/156659

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Fixed an issue where the **upload** and **prepare-content** commands didn't add `fromServerVersion` and `toServerVersion` to layouts.
 * Updated **lint** to use graph instead of id_set when running with `--check-dependent-api-module` flag.
 * Added the flag `--xsoar-only` to the **doc-review** command which enables reviewing documents that belong to XSOAR-supported Packs.
+* Fixed an issue where the **validate** command does not ignore several errors even when those errors are in the `.pack-ignore` file.
```
dorschw, line 10:
```suggestion
* Fixed an issue where **validate** would not ignore always ignore errors listed under `.pack-ignore`.
```

demisto_sdk/commands/common/hook_validations/base_validator.py
```diff
@@ -154,6 +154,7 @@ def handle_error(
         suggested_fix=None,
         warning=False,
         drop_line=False,
+        ignored_errors=None,
```
dorschw, line 157: can we remove `ignored_errors`'s default `None`, to prevent such bugs from happening again?

GuyAfik: but then every validation that uses it would require to pass in the `ignored_files` even though its already initialized in the constructor of `BaseValidator`. 

This argument has been added in cases where we have `BaseValidator` that was not initialized with `ignored_errors`

dorschw: Do we have any way to check for cases where it's not passed, and not in the parent? (except for adding an `if` everywhere)

GuyAfik: working on it

GuyAfik: Well, after carefully viewing it, it seems like we have a few issues to force developers to pass in the `ignored errors`:

1) There are some cases where we use the `BaseValidator` not only to handle validation errors, but to other purposes which does not involve the `ignored_errors`

2) implementing such option causes a lot of unit-tests to fail since they do not pass into the validators constructors the `ignored_errors`.

3) I think because of (1) - we cannot force a developer to pass in the `ignored_errors` into the base-validator, devs should know what they are doing. 

**general note**: i think we should highly consider refactoring the code involves validations because its very hard to read.

dorschw: Thank you, agreed.

dorschw (approved):

JudahSchwartz (approved):

PR #2730, opened 2023-02-06
fix lint `-cadm` duplicate files filtering

YuvHayun:
## Status
- [x] In Progress
- [ ] Ready

## Related Issues
fixes: [CIAC-5554](https://jira-hq.paloaltonetworks.local/browse/CIAC-5554)

## Description
Fixed an issue where **lint** sometimes failed when using the `-cdam` flag due to wrong file duplications filtering.


demisto_sdk/commands/lint/lint_manager.py
```diff
@@ -116,7 +116,13 @@ def __init__(
 
         if check_dependent_api_module:
             dependent_on_api_module = self._get_api_module_dependent_items()
-            self._pkgs = list(set(self._pkgs + dependent_on_api_module))
+            self._pkgs.extend(
+                [
+                    dependent_item
+                    for dependent_item in dependent_on_api_module
+                    if Path(os.path.dirname(dependent_item)) not in self._pkgs
```
dorschw, line 123:
```suggestion
                    if Path(dependent_item).parent.name not in self._pkgs
```

let's not mix `os.path` and `pathlib.Path`, but use the latter. I think what I did works, please check it.

dorschw (approved):

PR #2731, opened 2023-02-07
Move workflow python path

dorschw:


AradCarmi (approved):
I think it's better to leave the `link_pr_to_jira_issue` script name as is because it matches the same one in the `content` repo.

PR #2747, opened 2023-02-12
Retry on session timeouts when querying for playbook states

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5587

## Description
Fixed an issue where in some occasions when running test playbooks that take very long time (> 60 minutes), session timeouts could occur.

## Must have
- [x] Tests


demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -1872,18 +1872,24 @@ def _poll_for_playbook_state(self) -> str:
         """
         timeout = time.time() + self.playbook.configuration.timeout
         number_of_attempts = 1
+
+        should_re_authenticate = True
```
esharf, line 1876:
```suggestion
```

demisto_sdk/commands/test_content/TestContentClasses.py
```diff
@@ -1872,18 +1872,24 @@ def _poll_for_playbook_state(self) -> str:
...
-                )
+            except demisto_client.demisto_api.rest.ApiException as e:
+                if e.status == 401 and should_re_authenticate:
+                    self.server_context._configure_new_client()
+                    should_re_authenticate = False
+                else:
+                    playbook_state = "Pending"
+                    self.build_context.logging_module.exception(
+                        "Error when trying to get investigation playbook state"
+                    )
```
esharf, line 1892:
```suggestion
            except demisto_client.demisto_api.rest.ApiException as e:
                if e.status == 401 and should_re_authenticate:
                    self.server_context._configure_new_client()
                    self.build_context.logging_module.debug('new demisto_client created')
                    continue
                playbook_state = "Pending"
                self.build_context.logging_module.exception(
                    "Error when trying to get investigation playbook state"
                )
```

esharf (changes requested):

esharf (approved):

demisto_sdk/commands/test_content/TestContentClasses.py
GuyAfik: @esharf do you want to each time we receive 401, we will reset the client? not only the first time it occurs?  

this should be:

```
if e.status == 401:
```

cause `should_re_authenticate` is removed

esharf (approved):

PR #2749, opened 2023-02-12
fix e2e docker tests failing on 409

GuyAfik:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-5598)

## Description
fixed an issue where e2e docker tests were failing on 409 conflict error

## Must have
- [x] Tests

### DIFF
--- a/demisto_sdk/commands/common/MDXServer.py
+++ b/demisto_sdk/commands/common/MDXServer.py
@@ -52,11 +52,14 @@ def start_docker_MDX_server(
     """
     logging.info("Starting docker mdx server")
     get_docker().pull_image(DEPENDENCIES_DOCKER)
-    if running_container := init_global_docker_client().containers.list(
-        filters={"name": DEMISTO_DEPS_DOCKER_NAME}
-    ):
-        running_container[0].stop()
+
+    docker_client = init_global_docker_client()
+
     location_in_docker = f"/content/{_SERVER_SCRIPT_NAME}"
+    while running_container := docker_client.containers.list(
+        filters={"name": DEMISTO_DEPS_DOCKER_NAME}, all=True
+    ):
+        running_container[0].remove(force=True)
     container: docker.models.containers.Container = get_docker().create_container(
         name=DEMISTO_DEPS_DOCKER_NAME,
         image=DEPENDENCIES_DOCKER,
--- a/demisto_sdk/commands/common/MDXServer.py
+++ b/demisto_sdk/commands/common/MDXServer.py
@@ -40,7 +40,7 @@ def start_docker_MDX_server(
     """
         This function will start a docker container running a node server listening on port 6161.
         The container will erase itself after exit.
-        If there's a running server already with the same name it will be stopped before starting a new one.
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/common/MDXServer.py
```diff
@@ -83,13 +96,14 @@ def start_docker_MDX_server(
...
+        remove_container(container)
 
 
-def stop_docker_container(container):
+def remove_container(container):
     if container:
-        logging.info("Stopping mdx docker server")
-        container.stop()  # type: ignore
+        print(f"Removing container {container.name}")
+        container.remove(force=True)
```
tkatzir, line 105: If the container isn't there, will `remove` fail, or be quiet?

tkatzir (approved):
Looks good. One question.
Also - You'll need to create another PR to replace the `print`s with `logger.info`.

demisto_sdk/commands/common/MDXServer.py
GuyAfik: as long as you have container object, it exists, hence it wouldn't fail, but if docker api returned and error then an exception will be raised. In our case it can raise only if container doesn't exist.

https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.remove

GuyAfik:
> Looks good. One question. Also - You'll need to create another PR to replace the `print`s with `logger.info`.

no problem, when will we support logging in e2e tests? i can change it now

GuyAfik:
@tkatzir @JudahSchwartz do we need a changelog here? in my opinion we dont

demisto_sdk/commands/common/MDXServer.py
```diff
@@ -83,13 +96,14 @@ def start_docker_MDX_server(
...
-        stop_docker_container(container)
+        remove_container(container)
 
 
-def stop_docker_container(container):
+def remove_container(container):
     if container:
-        logging.info("Stopping mdx docker server")
-        container.stop()  # type: ignore
+        print(f"Removing container {container.name}")
```
JudahSchwartz, line 104: also print that stopping mdx server

demisto_sdk/commands/common/MDXServer.py
```diff
@@ -52,25 +52,38 @@ def start_docker_MDX_server(
...
-        command=["node", location_in_docker],
-        user=f"{os.getuid()}:4000",
-        files_to_push=[(server_script_path(), location_in_docker)],
-        auto_remove=True,
-        ports={"6161/tcp": 6161},
-    )
+    while mdx_container := docker_client.containers.list(
+        filters={"name": DEMISTO_DEPS_DOCKER_NAME}, all=True
+    ):
+        remove_container(mdx_container[0])
```
JudahSchwartz, line 62: add a print line that a container was found here

JudahSchwartz, line 62: print iteration number

JudahSchwartz (approved):

PR #2753, opened 2023-02-13
Add docker image target in lint for testing native image supported content

DinaMeylakh:
## Status
- [ ] In Progress
- [x] Ready

## Related Issues
fixes: [link to the issue](https://jira-hq.paloaltonetworks.local/browse/CIAC-5611)

## Description

## Screenshots

## Must have
- [ ] Tests
- [ ] Documentation


JudahSchwartz (approved):
looks good to me! nice

demisto_sdk/commands/lint/tests/test_linter/gather_facts_test.py
```diff
@@ -335,12 +352,72 @@ def test_invalid_docker_image_as_docker_image_flag(self, mocker, pack):
...
+                pack.repo_path,
+                test_integration.path,
+                True,
+                docker_image_flag=linter.DockerImageFlagOption.NATIVE_TARGET.value,
+                docker_image_target=invalid_docker_image,
+            )
+            runner._gather_facts(modules={})
+
+        # Verify docker images:
+        print(f"log error {log_error.call_args_list}")
```
GuyAfik, line 407: why do we use print here and not log?

GuyAfik (approved):
awesome!

demisto_sdk/commands/lint/linter.py
```diff
@@ -1389,6 +1393,7 @@ def _check_native_image_flag(self, docker_image_flag):
             DockerImageFlagOption.NATIVE_DEV.value,
             DockerImageFlagOption.NATIVE_GA.value,
             DockerImageFlagOption.NATIVE_MAINTENANCE.value,
+            DockerImageFlagOption.NATIVE_TARGET.value,
```
ShacharKidor, line 1396: I you added this value here, I think we need to add it to the error message and to the function's docstring as a possible option.

demisto_sdk/commands/lint/linter.py
```diff
@@ -1530,7 +1535,11 @@ def _get_all_docker_images(
         return imgs
 
     def _get_docker_images_for_lint(
-        self, script_obj: Dict, script_id: str, docker_image_flag: str
+        self,
+        script_obj: Dict,
+        script_id: str,
+        docker_image_flag: str,
+        docker_image_target: Union[str, None],
```
ShacharKidor, line 1542:
```suggestion
        docker_image_target: Optional[str],
```

demisto_sdk/commands/lint/linter.py
```diff
@@ -1601,15 +1612,28 @@ def _get_docker_images_for_lint(
...
-                docker_image_flag
-            ):
+            supported_section = docker_image_flag
+            if docker_image_flag == DockerImageFlagOption.NATIVE_TARGET.value:
+                supported_section = DockerImageFlagOption.NATIVE_DEV.value
+
+            native_image = self._get_native_image_name_from_config_file(
+                supported_section
+            )
+            if native_image:
```
ShacharKidor, line 1622: you can use the walrus operator as I did before: 
```suggestion
            if native_image := self._get_native_image_name_from_config_file(
                supported_section
            )
```

demisto_sdk/commands/lint/linter.py
```diff
@@ -1601,15 +1612,28 @@ def _get_docker_images_for_lint(
 
             self._check_native_image_flag(docker_image_flag)
 
-            if native_image := self._get_native_image_name_from_config_file(
-                docker_image_flag
-            ):
+            supported_section = docker_image_flag
```
ShacharKidor, line 1615: I'm not sure about the naming of this var 'supported_section' - what is the meaning here?

demisto_sdk/commands/lint/linter.py
```diff
@@ -1601,15 +1612,28 @@ def _get_docker_images_for_lint(
...
+            if docker_image_flag == DockerImageFlagOption.NATIVE_TARGET.value:
+                supported_section = DockerImageFlagOption.NATIVE_DEV.value
```
ShacharKidor, line 1617: Why did you choose here the native dev? just in order to get into the 'if native_image' cond? 
Cause in the end the native image will be the docker_image_target image right? 

What happens if the script\integration isn't supported by the native:dev (supported by the native:ga only) but you still want to run it on the docker image target? is it a possible scenario?

demisto_sdk/commands/lint/linter.py
```diff
@@ -1601,15 +1612,28 @@ def _get_docker_images_for_lint(
...
+
+            native_image = self._get_native_image_name_from_config_file(
+                supported_section
+            )
+            if native_image:
 
                 if self._is_native_image_support_script(
                     native_image, supported_native_images, script_id
                 ):  # Integration/Script is supported by the requested native image
+                    native_image_ref: Union[str, None] = ""
```
ShacharKidor, line 1627:
```suggestion
                    native_image_ref: Optional[str] = ""
```

demisto_sdk/commands/lint/linter.py
```diff
@@ -1560,6 +1569,8 @@ def _get_docker_images_for_lint(
             script_obj (dict): A yml dict of the integration/script that lint runs on.
             script_id (str): The ID of the integration/script that lint runs on.
             docker_image_flag (str): A flag indicates on which docker images lint should run.
+            docker_image_target (Optional[str]): The docker image to test native supported content on.
```
ShacharKidor, line 1572: I think it is worth adding an exception in case a user selects the 'native:target' flag without setting a docker_image_target or setting a docker_image_target with another flag.

ShacharKidor (changes requested):
@DinaMeylakh - Nice work!

I added a few comments and suggestions - please have a look. 

Please also add documentation of the new arg to the demisto_sdk/commands/lint/README.md.

Just to make sure I fully understand - the purpose of this new flag is to run docker tests on a specific docker image ref, only for content items that are supported by native images in general. Right?

demisto_sdk/__main__.py
```diff
@@ -923,7 +923,15 @@ def secrets(config, **kwargs):
...
-    " 'from-yml'.",
+    " 'from-yml', 'native:target'. To run lint only on native supported content with a specific image,"
+    " use 'native:target' with --docker-image-target <specific-image>.",
+)
+@click.option(
+    "-dit",
+    "--docker-image-target",
+    default="",
+    help="The docker image to lint native supported content with. Should only be used with "
+    "--docker-image native:target. Will be ignored otherwise.",
```
dorschw, line 934: let's raise an error when used without, instead of ignoring

demisto_sdk/commands/lint/linter.py
DinaMeylakh: so, this is mainly used for providing `_get_native_image_name_from_config_file` an argument. 
This function "Gets a native docker image flag and returns its mapped native image name from the 'docker_native_image_config.json' file."

when using native target, we want all the native:dev *supported* images to lint on, which is different than the "docker_image_flag" (in our case native:target)

So instead of calling it docker_image_flag, I want the supported section from config file
I agree that the naming is a little confusing, I'm changing to `image_support`, let me know if you think of a better name for this variable.

demisto_sdk/commands/lint/linter.py
DinaMeylakh: that's not a possible scenario, we only want to target native-dev supported images

CHANGELOG.md
```diff
@@ -1,9 +1,37 @@
...
+## 1.10.4
+* Added support for running **lint** in multiple native-docker images.
+
+## 1.10.3
+* Fixed an issue where running **format** would fail after running npm install.
+* Improved the graph validations in the **validate** command:
+  - GR100 will now run on all content items of changed packs.
+  - GR101 and GR102 will now catch invalid fromversion/toversion of files **using** the changed items.
+  - GR103 errors will raise a warning when using the *-a* flag, but an error if using the *-i* or *g* flags.
+* Fixed an issue where test-playbooks timed out.
```
JudahSchwartz, line 20: why are all these part of your changes here?

DinaMeylakh:
@ShacharKidor 
`Just to make sure I fully understand - the purpose of this new flag is to run docker tests on a specific docker image ref, only for content items that are supported by native images in general. Right?`

That's right, specifically content items that are supported by native:dev

demisto_sdk/commands/lint/lint_manager.py
```diff
@@ -657,6 +662,16 @@ def run(
...
+        # Check docker image flags are in order
+        if (
+            docker_image_target
+            and docker_image_flag != DockerImageFlagOption.NATIVE_TARGET.value
+        ):
+            raise ValueError(
+                f"Recieved docker image target {docker_image_target} without docker "
+                f"image flag {DockerImageFlagOption.NATIVE_TARGET.value}. Aborting."
+            )
+
```
dorschw, line 674: What would be the implications of just assuming that `dit` implies `docker_image_flag`?

DinaMeylakh: not sure I understand, it would raise a value error like you can see from code section that is highlighted

CHANGELOG.md
DinaMeylakh: pre commit hook made merging difficult, I belive it is fixed now? where did you see this?

demisto_sdk/commands/lint/lint_manager.py
dorschw: yeah, that's the existing behavior according to the code. We could decide that instead of raising an error, we'd just assume the other flag was used as well. What do you think about that?

DinaMeylakh: if one would specify --docker_image_flag "native:target" --dit "native:dev", no such image would be found in the registry

dorschw: in this case, we could raise an error, rather than block usage without the flag

DinaMeylakh: I see that it makes the first flag useless, but I don't think I want that, this can cause confusion between `--docker_image` and `--docker_image_target`
If I want to test a certain file (that is not native:dev supported) with a specified docker image, I could use --docker_image_target by mistake, causing it to run on an empty set of content items, and pass without me knowing this was never checked. 
if we use both, we guarantee the user knows a little bit more about what they are doing

that said, this is not such a strong argument & I can change this if you feel this is worth it

CHANGELOG.md
JudahSchwartz: looks fixed

ShacharKidor (approved):
Very nice :)

PR #2802, opened 2023-02-26
Fixed validation in feeds integration for collect/connect sections

Shellyber:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5481

## Description
Fixed the Feed Integration validation for the Collect/Connect section

### DIFF
--- a/demisto_sdk/commands/common/tests/integration_test.py
+++ b/demisto_sdk/commands/common/tests/integration_test.py
@@ -2131,14 +2131,14 @@ class TestIsFeedParamsExist:
     def test_section_field_feed(self):
         """
         Given:
-        - Parameters of feed integration.
-        - One param with the section field
+        - Parameters of feed integration, where one parameter as a 'section: collect' property.
 
         When:
-        - Integration has all feed required params, and validation won't fail on the section field.
+        - Integration has all feed required params and running all_feed_params_exist on it.
 
         Then:
-        - Ensure that all_feed_params_exists() returns true.
+        - Ensure that all_feed_params_exists() returns true,
+            which means validation did not fail on the additional section field.
         """
         configuration = self.validator.current_file["configuration"]
         for item in configuration:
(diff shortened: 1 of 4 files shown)


Shellyber:
@ShahafBenYakir the `sectionorder` and the `section` field are already in the integration schema.
The Feed validation is a bit different, it is working with constants that check all the mandatory fields.
So this is the best solution for this issue.
It is that or to change the entire flow and create a new and different schema for feed integrations and change the entire validation flow accordingly

Shellyber:
@ShahafBenYakir Test was added, please review :)

demisto_sdk/commands/common/tests/integration_test.py
```diff
@@ -2128,6 +2128,26 @@ def test_hidden_feed_reputation_field(self):
             self.validator.all_feed_params_exist() is True
         ), "all_feed_params_exist() returns False instead True for feedReputation param"
 
+    def test_section_field_feed(self):
+        """
+        Given:
+        - Parameters of feed integration.
+        - One param with the section field
```
ShahafBenYakir, line 2135:
```suggestion
        - Parameters of feed integration, where one parameter as a 'section: collect' property.
```

demisto_sdk/commands/common/tests/integration_test.py
```diff
@@ -2128,6 +2128,26 @@ def test_hidden_feed_reputation_field(self):
...
+
+        When:
+        - Integration has all feed required params, and validation won't fail on the section field.
```
ShahafBenYakir, line 2138:
```suggestion
        - Integration has all feed required params and running all_feed_params_exist on it.
```

demisto_sdk/commands/common/tests/integration_test.py
```diff
@@ -2128,6 +2128,26 @@ def test_hidden_feed_reputation_field(self):
...
+
+        Then:
+        - Ensure that all_feed_params_exists() returns true.
```
ShahafBenYakir, line 2141:
```suggestion
        - Ensure that all_feed_params_exists() returns true, which means validation did not fail on the additional section field.
```

ShahafBenYakir (approved):
Great

PR #2805, opened 2023-02-26
Error code params parsing bug fix

dansterenson:
Until now error code arguments used to have parameters with type 'list'. In the last SDK update, new error codes were added and were using List[str] in their arguments which caused the following exception:
`TypeError: Type List cannot be instantiated; use list() instead`

Fixed it by adding this type to the TYPE_FILLER_MAPPING dict.

FYI @MLainer1 @dantavori


CHANGELOG.md
```diff
@@ -5,9 +5,8 @@
 * Added a newline at the end of README files generated in **generate-docs**.
 * Added the value `3` (out of bounds) to the `onChangeRepAlg` and `reputationCalc` fields under the `IncidentType` and `GenericType` schemas. **validate** will allow using it now.
 * Fixed an issue where **doc-review** required dot suffixes in release notes describing new content.
-
-## 1.10.4
 * Added support for running **lint** in multiple native-docker images.
```
dorschw, line 8: bad merge, fix please

dorschw (approved):

PR #2807, opened 2023-02-27
GR103 fixes

dantavori:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5785

## Description
1. GR103 will now be split into two:
   - Cases where a warning should be raised - if the using content item is a test playbook/test script,
      or if the dependency in the used content item is optional.
   - Cases where an error should be raised - the complementary case.
2. Fixed an issue in the layout, incident field/type, and indicator field/type parsers, where USES relationships to items with empty string as a name, were created.
3. Updated the `IGNORED_FIELDS` list in playbook parser to avoid some false positives in GR103.
4. Added the `extract_field_from_mapping()` method which is used in the mapper parser, to improve the collection of USES relationships to incident/indicator fields from outgoing mappers.
5. Extended the server content items list to avoid some false positives in GR103.


demisto_sdk/commands/common/hook_validations/graph_validator.py
```diff
@@ -206,14 +206,30 @@ def handle_invalid_toversion(
...
+            or if the dependency is optional.
+        2. Cases where an error should be raised - the complementary case.
+        """
+        return all(
+            [
+                self._find_unknown_content_uses(raises_error=False),
+                self._find_unknown_content_uses(raises_error=True),
+            ]
+        )
+
```
MLainer1, line 221:
```suggestion
        is_valid = [
                self._find_unknown_content_uses(raises_error=False),
                self._find_unknown_content_uses(raises_error=True),
            ]
        return all(is_valid)

```

MLainer1: For debugging reasons, I find that it is better to do it this way (easier to see which one of the validation returned the faulty error)

demisto_sdk/commands/common/hook_validations/graph_validator.py
```diff
@@ -206,14 +206,30 @@ def handle_invalid_toversion(
...
+    def _find_unknown_content_uses(self, raises_error: bool) -> bool:
         """Validates that there is no usage of unknown content items.
         Note: if self.file_paths is empty, the validation runs on all files - in this case, returns a warning.
-        otherwise, returns an error.
+        otherwise, returns an error iff raises_error is True.
```
MLainer1, line 225: iff ==  if and only if ?

demisto_sdk/commands/common/hook_validations/graph_validator.py
```diff
@@ -206,14 +206,30 @@ def handle_invalid_toversion(
 
     @error_codes("GR103")
     def is_file_using_unknown_content(self):
+        """Validates that there is no usage of unknown content items.
+        The validation runs twice:
+        1. Cases where a warning should be raised - if the using content item is a test playbook/test script,
```
MLainer1, line 211: maybe we can implement in the future the same thing also for the from_version and the to_version validations

dantavori: possible, but as long as it does not make noise it's less necessary.

demisto_sdk/commands/common/hook_validations/graph_validator.py
dantavori: yes.

PR #2813, opened 2023-02-28
Fix server min version field in pack metadata

yaakovpraisler:
Fixed the serverMinVersion when dumping the pack metadata to calculating by the minimum from version of the pack's content items.

@YuvHayun FYI


dantavori (approved):
please add a link to the issue

PR #2817, opened 2023-03-02
upgrade neo4j version to 5

ilaner:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5725

## Description
* Upgrade neo4j to version 5
* Fix typing issues

## Content nightly
https://code.pan.run/xsoar/content/-/pipelines/4727822


demisto_sdk/commands/content_graph/neo4j_service.py
```diff
@@ -86,6 +87,34 @@ def _download_apoc():
     with open(plugins_folder / "apoc.jar", "wb") as f:
         f.write(response.content)
 
+def _docker_start():
+        docker_client = init_global_docker_client()
+        _stop_neo4j_service_docker(docker_client)
```
dantavori, line 92: you have these two lines twice now

demisto_sdk/commands/content_graph/neo4j_service.py
```diff
@@ -101,31 +130,12 @@ def start():
...
-            "test": f"curl --fail {NEO4J_DATABASE_HTTP} || exit 1",
-            "interval": 5 * 1000000000,
-            "timeout": 15 * 1000000000,
-            "retries": 10,
-        },
-    )
+    try:
+        _docker_start()
+    except Exception as e:
+        logger.debug(f"Could not start neo4j container, deleting data folder and trying again. {e}")
```
dantavori, line 136:
```suggestion
        logger.debug(f"Could not start neo4j container, delete data folder and trying again. {e}")
```
are you sure this is the only possible error?

dantavori (approved):

demisto_sdk/commands/content_graph/neo4j_service.py
ilaner: This is not the only possible error, but it's possible one that I had when run locally in the new version.
In case of another error, `_docker_start` will fail again.

Issue #2138, opened 2022-05-31
flag --force behaves improperly when downloading layout


K-Yo:
when using the command `demisto-sdk download` **without** the flag `--force` on a layout, if the layout is already on the local pack it still gets downloaded and overwrites local changes.

steps to reproduce:

- `demisto-sdk download -fmt -r 'mylayout' -o Packs/MyPack`
- modify ` Packs/MyPack/Layouts/layoutscontainer-mylayout.json`
- `demisto-sdk download -fmt -r 'mylayout' -o Packs/MyPack` (a second time) <- it will get 
- modifications are lost

dorschw:
Hey @K-Yo, thanks for opening an issue,
this will be handled internally as [CIAC-3019](https://jira-hq.paloaltonetworks.local/browse/CIAC-3019).

PR #2861, opened 2023-03-29
Fix layouts gets overriden in download without -f

GuyAfik:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-22748
fixes: https://github.com/demisto/demisto-sdk/issues/2138

## Description
Fixes an issue that when trying to download a layout without the `-f` arg, it would override the layout inside a pack.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 * Fixed an issue in the **download** command where layouts were overriden even without the `-f` option.
+* Fixed an issue where Demisto-SDK did not detect layout ID when using the **download** command.
 * Fixed an issue where the **lint** command ran on `native:dev` supported content when passing the `--docker-image all` flag, instead it will run on `native:candidate`.
 * Added support for `native:candidate` as a docker image flag for **lint** command.
 * Fixed an issue where logs and messages would not show when using the **download** command.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+* Fixed an issue in the **download** command where layouts were overriden even without the `-f` option.
```
dorschw, line 4: add another bullet for the actual bug, where the SDK didn't appropriately detect the `id` (so that if anything happens down the road, we'll notice it in the changelog)

demisto_sdk/commands/common/tools.py
```diff
@@ -849,7 +849,8 @@ def get_entity_id_by_entity_type(data: dict, content_entity: str):
         if content_entity in (INTEGRATIONS_DIR, SCRIPTS_DIR):
             return data.get("commonfields", {}).get("id", "")
         elif content_entity == LAYOUTS_DIR:
```
dorschw, line 851: is the content entity the _directory_, or do we just reuse the constant with the same value?
If it's the latter, please find/create another constant

dorschw (commented):
Nice. Please make sure that the TestSuite object creates the id in the right place before merging

demisto_sdk/commands/common/tools.py
GuyAfik: its the directory

GuyAfik:
> Nice. Please make sure that the TestSuite object creates the id in the right place before merging

its in the right place :)

dorschw (approved):

PR #2881, opened 2023-04-04
Fix pb inputs validation

adi88d:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6206

## Description
- Updates the `collect_all_inputs_in_use` method to remove `}` from the PB input name.
- Added to **playbook_valid_inputs.yml** the input `ResourceName` and use it in the task like this:
`simple: ${inputs.ResourceName}`
 for testing the validation after the change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 * Added the pack version to the code files when calling **unify**. The same value is removed when calling **split**.
 * Contribution PRs that update outdated packs now display a warning message.
+* Fixed a bug where **validate** returned error on playbook inputs with special characters.
 
 ## 1.12.0
 * Added the **pre-commit** command, to improve code quality of XSOAR content.
--- a/demisto_sdk/commands/common/hook_validations/playbook.py
+++ b/demisto_sdk/commands/common/hook_validations/playbook.py
@@ -123,7 +123,10 @@ class PlaybookValidator(ContentEntityValidator):
             input = input.strip()
             splitted = input.split(".")
             if len(splitted) > 1 and splitted[1] and not splitted[1].startswith(" "):
-                result.add(splitted[1].replace('}', ''))
+                input_in_use = splitted[1]
+                if input_in_use.endswith('}'):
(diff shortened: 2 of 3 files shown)


demisto_sdk/commands/common/hook_validations/playbook.py
```diff
@@ -123,7 +123,7 @@ def collect_all_inputs_in_use(self) -> Set[str]:
             input = input.strip()
             splitted = input.split(".")
             if len(splitted) > 1 and splitted[1] and not splitted[1].startswith(" "):
-                result.add(splitted[1])
+                result.add(splitted[1].replace('}', ''))
```
dorschw, line 126: do we replace _all_ `}`s in the value, or just when it's a suffix?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 * Added the pack version to the code files when calling **unify**. The same value is removed when calling **split**.
 * Contribution PRs that update outdated packs now display a warning message.
+* Fixed a bug where **validate** returned error on playbook inputs with special characters.
```
dorschw, line 6:
```suggestion
* Fixed an issue where **validate** would fail running when playbook inputs contained special characters.
```

demisto_sdk/commands/common/hook_validations/playbook.py
```diff
@@ -123,7 +123,10 @@ def collect_all_inputs_in_use(self) -> Set[str]:
             input = input.strip()
             splitted = input.split(".")
             if len(splitted) > 1 and splitted[1] and not splitted[1].startswith(" "):
-                result.add(splitted[1])
+                input_in_use = splitted[1]
+                if input_in_use.endswith("}"):
+                    input_in_use = input_in_use[:-1]
+                result.add(input_in_use)
```
dorschw, line 129:
```suggestion
                input_in_use = splitted[1]
                result.add(input_in_use[:-1] if input_in_use.endswith("}") else if input_in_use)
```

dorschw (approved):

PR #2896, opened 2023-04-10
get file fix

dantavori:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-4661
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-4701

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -767,7 +767,9 @@ def get_file(file_path: Union[str, Path], type_of_file: str, clear_cache: bool =
     if clear_cache:
         get_file.cache_clear()
 
-    file_path = Path(get_content_path()) / file_path  # type: ignore[arg-type]
+    file_path = Path(file_path)  # type: ignore[arg-type]
+    if not file_path.exists():
+        file_path = Path(get_content_path()) / file_path  # type: ignore[arg-type]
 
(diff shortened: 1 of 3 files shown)


ilaner:
Please import CONTENT_PATH from content_path_path and use CONTENT_PATH / {name}

demisto_sdk/commands/common/tools.py
```diff
@@ -767,7 +767,7 @@ def get_file(file_path: Union[str, Path], type_of_file: str, clear_cache: bool =
     if clear_cache:
         get_file.cache_clear()
 
-    file_path = Path(file_path).absolute()
+    file_path = Path(get_content_path()).joinpath(file_path)  # type: ignore
```
dorschw, line 770: i guess mypy screams because of the optional `None` output of `get_content_path()`, please either fix it there (perhaps use `or ""`?), or use a specific ignore (rather than a catch-all `type: ignore`)

dantavori:
> Please import CONTENT_PATH from content_path_path and use CONTENT_PATH / {name}

@ilaner I did that at first - see https://github.com/demisto/demisto-sdk/pull/2896/commits/3dd398b7f590cca23788e0aa679cba6ceb34c21e
It causes a circular import, so I changed to the current

ilaner (approved):

Issue #2906, opened 2023-04-14
Wrong unified yml file name printed to the screen


ennozdd:
![image](https://user-images.githubusercontent.com/49711791/232034886-7b6bb198-5e3f-47ed-8665-9225dfd503c0.png)

unified file name should be integration-Microsoft365Defender.yml

ennozdd:
```[2023-04-14T14:54:02Z] - [MainThread] - [DEBUG] - cmd.py:976 - Popen(['git', 'remote', 'get-url', '--all', 'origin'], cwd=/home/ennozdd/community/content, universal_newlines=False, shell=None, istream=None)
[2023-04-14T14:54:02Z] - [MainThread] - [DEBUG] - connectionpool.py:1003 - Starting new HTTPS connection (1): pypi.org:443
[2023-04-14T14:54:02Z] - [MainThread] - [DEBUG] - connectionpool.py:456 - https://pypi.org:443 "GET /pypi/demisto-sdk/json HTTP/1.1" 200 32231
[2023-04-14T14:54:02Z] - [MainThread] - [INFO] - __main__.py:216 - You are using demisto-sdk 1.12.0.
[2023-04-14T14:54:03Z] - [MainThread] - [DEBUG] - base_content.py:115 - Loading content item from path: Packs/Microsoft365Defender/Integrations/Microsoft365Defender/Microsoft365Defender.yml
[2023-04-14T14:54:03Z] - [MainThread] - [DEBUG] - content_item.py:96 - Parsing content item Packs/Microsoft365Defender/Integrations/Microsoft365Defender/Microsoft365Defender.yml
[2023-04-14T14:54:03Z] - [MainThread] - [DEBUG] - content_item.py:129 - Parsed Integration:Microsoft 365 Defender
[2023-04-14T14:54:03Z] - [MainThread] - [DEBUG] - base_content.py:136 - Loading content item from path: Packs/Microsoft365Defender/Integrations/Microsoft365Defender/Microsoft365Defender.yml as <class 'demisto_sdk.commands.content_graph.objects.integration.Integration'>
[2023-04-14T14:54:03Z] - [MainThread] - [DEBUG] - content_item.py:174 - Normalized file name from Microsoft365Defender.yml to integration-Microsoft365Defender.yml
[...]

PR #2907, opened 2023-04-16
log the output file in prepare_content

ilaner:
## Related Issues
fixes: https://github.com/demisto/demisto-sdk/issues/2906

## Description
Currently we log only the unify step and print that we were succeeded to unify the YML, but we don't log the output file path so the user does not know where the file is saved.


dorschw (approved):

dorschw (commented):
changelog

PR #2911, opened 2023-04-16
Removing authentication when sending HTTP request to CircleCi (SDK)

rshunim:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6219

## Description
Authentication isn't needed since it's an open source. We'll be required to return authentication and generate a token if there's a rate limit.
Authentication isn't working right now, so we put '# auth=self.auth' under comment.


GuyAfik (approved):
@rshunim nice, please change the PR title to be more meaningful.

i don't think we need a change-log here, but better verify with @dorschw

please also mention the reason that it fixes the issue hen we put self.auth in comment.

PR #2918, opened 2023-04-17
add xif to secrets detection

dantavori:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-4845


anara123 (approved):
Add changelog and make sure the fix is actually works

PR #2923, opened 2023-04-19
Run Linter on Docker

tkatzir:
Added logs.

## Related Issues
fixes:

### DIFF
--- a/demisto_sdk/commands/lint/linter.py
+++ b/demisto_sdk/commands/lint/linter.py
@@ -1002,7 +1002,7 @@ class Linter:
             # see: https://github.com/docker/docker-py/issues/2696#issuecomment-721322548
             system_is_not_darwin = platform.system() != "Darwin"
             not_connetion_broken = "Connection broken" not in str(err)
-            logger.info(
+            logger.warning(
                 f"Didn't remove container {container_name} as it wasn't found, {system_is_not_darwin=}, {not_connetion_broken=}"
             )
             if system_is_not_darwin or not_connetion_broken:
--- a/demisto_sdk/commands/lint/linter.py
+++ b/demisto_sdk/commands/lint/linter.py
@@ -1007,6 +1007,11 @@ class Linter:
             )
             if system_is_not_darwin or not_connetion_broken:
                 raise
+        except Exception as exc:
+            logger.exception(
+                f"Didn't remove container {container_name}. Got an Exception {exc}"
+            )
+            raise exc
 
     def _docker_run_linter(
         self, linter: str, test_image: str, keep_container: bool
--- a/demisto_sdk/commands/lint/linter.py
+++ b/demisto_sdk/commands/lint/linter.py
@@ -990,13 +990,22 @@ class Linter:
 
     def _docker_remove_container(self, container_name: str):
         try:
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/lint/linter.py
```diff
@@ -990,13 +990,22 @@ def _docker_image_create(self, docker_base_image: List[Any]) -> Tuple[str, str]:
 
     def _docker_remove_container(self, container_name: str):
         try:
+            logger.info(f"Trying to remove container {container_name}")
```
dorschw, line 993:
```suggestion
            logger.debug(f"Trying to remove container {container_name}")
```

demisto_sdk/commands/lint/linter.py
```diff
@@ -990,13 +990,22 @@ def _docker_image_create(self, docker_base_image: List[Any]) -> Tuple[str, str]:
...
             container = self._docker_client.containers.get(container_name)
+            logger.info(f"Found container {container_name=}, {container.id=}")
```
dorschw, line 995:
```suggestion
            logger.debug(f"Found container {container_name=}, {container.id=}")
```

demisto_sdk/commands/lint/linter.py
```diff
@@ -990,13 +990,22 @@ def _docker_image_create(self, docker_base_image: List[Any]) -> Tuple[str, str]:
...
             container.remove(force=True)
+            logger.info(f"Successfully removed container {container_name}")
```
dorschw, line 997:
```suggestion
            logger.debug(f"Successfully removed container {container_name}")
```

demisto_sdk/commands/lint/linter.py
```diff
@@ -990,13 +990,22 @@ def _docker_image_create(self, docker_base_image: List[Any]) -> Tuple[str, str]:
...
         except docker.errors.NotFound:
+            logger.info(f"Didn't remove container {container_name} as it wasn't found")
             pass
         except requests.exceptions.ChunkedEncodingError as err:
             # see: https://github.com/docker/docker-py/issues/2696#issuecomment-721322548
-            if platform.system() != "Darwin" or "Connection broken" not in str(err):
+            system_is_not_darwin = platform.system() != "Darwin"
+            not_connetion_broken = "Connection broken" not in str(err)
+            logger.info(
```
dorschw, line 1005: either error or warning imo

dorschw (approved):

demisto_sdk/commands/lint/linter.py
tkatzir: These are usually info level. See line 1015

demisto_sdk/commands/lint/linter.py
tkatzir: These are usually info level. See line 1015

demisto_sdk/commands/lint/linter.py
tkatzir: It's not necessarrily an error.
We try to remove the container anyway.
If it's not found - We're good.
But we want to know if it didn't find it, and then we'll get:
`Conflict. The container name "foo" is already in use by container "bar". You have to remove (or rename) that container to be able to reuse that name.`

PR #2930, opened 2023-04-20
RN checker: Do not require dot for new items

michal-dagan:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5762


demisto_sdk/commands/doc_reviewer/tests/rn_checker_test.py
```diff
@@ -82,6 +82,13 @@
             True,
         ),
         (["##### New: script", "Blah.", "##### script", "Blah."], False),
+        (
+            [
+                "#### Indicator Fields",
+                "- New: **Indicator Field**",
```
dorschw, line 88: Is the prefix dash needed here? I thought the fix was removing it from `rn_checker.py`

dorschw (approved):

PR #2939, opened 2023-04-23
fix MYPYPATH in demisto-sdk precommit

ilaner:
## Related Issues
fixes:

## Description
* Mypy does not support "site-packages" imports in the import

### DIFF
--- a/demisto_sdk/commands/pre_commit/.pre-commit-config_template.yaml
+++ b/demisto_sdk/commands/pre_commit/.pre-commit-config_template.yaml
@@ -32,7 +32,7 @@ repos:
     hooks:
       - id: mypy
         name: mypy
-        entry: mypy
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/pre_commit/.pre-commit-config_template.yaml
```diff
@@ -32,8 +32,8 @@ repos:
     hooks:
       - id: mypy
         name: mypy
-        entry: mypy
-        language: system
+        entry: poetry run mypy
```
dorschw, line 35: do we always have poetry installed?

ilaner: you right we dont need it

dorschw (approved):

PR #2941, opened 2023-04-23
Lint: move init into except block

JudahSchwartz:
looks like an issue where if docker isnt installed, will fail before the try block


demisto_sdk/commands/lint/lint_manager.py
```diff
@@ -258,13 +259,16 @@ def _gather_facts() -> Dict[str, Any]:
...
-        )
+
         try:
+            docker_client: docker.DockerClient = init_global_docker_client(
+                log_prompt="LintManager"
+            )
             logger.debug("pinging docker daemon")
             docker_client.ping()
         except (
+            docker.errors.DockerException,
```
JudahSchwartz, line 270: going to remove this 1, but why was this translated at all? @ilaner ?

ShahafBenYakir (approved):

PR #2944, opened 2023-04-24
add from/to version to xsiam items schema

anara123:
## Related Issues
fixes: @nkanon please add the relevant ticket

## Description
Added from/to version to xsiam content items schema


dantavori (commented):
is it going to be required? if so - need to specify this in the schemas and of course update the relevant content items accordingly.
in addition, need to add a default fromversion value in `FILETYPE_TO_DEFAULT_FROMVERSION` constant (the first supported versions of the types).

anara123:
@dantavori I don't see a reason for this to be required

nkanon (commented):
Hi @anara123,
I've began testing the fromVersion on content items.
For now, it looks like this filter mechanism is working as expected on dashboards.

Will update soon on the rest of the content items.

dantavori (approved):

dorschw (changes requested):
We use `fromversion` for `yml` files, and `fromVersion` for `json`s. Please change accordingly.

anara123:
> We use `fromversion` for `yml` files, and `fromVersion` for `json`s. Please change accordingly.

@dorschw I checked. it should be correct

PR #2970, opened 2023-05-03
Update graph git enhancement

ilaner:
## Related Issues
fixes:

## Description
* Add staged files to the git diff.


demisto_sdk/commands/content_graph/content_graph_commands.py
```diff
@@ -72,7 +72,8 @@ def update_content_graph(
             content_graph_interface, marketplace, dependencies, output_path
         )
         return
-
+    if not imported_path and not use_current:
+        use_git = True
```
dantavori, line 76: I'm not in favor of this, if someone wants to use git they need to specify it. if they don't specify the flag, they expect to not use git.
also f we do this, we will stop supporting updating only the specified packs.

PR #2974, opened 2023-05-03
make sure graph validations cause error

DinaMeylakh:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6337

## Description
Enable graph ids validations for non mandatory dependencies


YuvHayun (approved):
Nice!

PR #2978, opened 2023-05-04
[generate-outputs] json-to-outputs - generate context paths for empty lists or dictionaries

michal-dagan:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-5925

## Description
* Fixed an issue where **generate-outputs** did not generate context paths for empty lists or dictionaries in the response.

### DIFF
--- a/demisto_sdk/commands/generate_outputs/json_to_outputs/json_to_outputs.py
+++ b/demisto_sdk/commands/generate_outputs/json_to_outputs/json_to_outputs.py
@@ -99,10 +99,10 @@ def flatten_json(nested_json, camelize=False):
         except IndexError:
             name = name.title() if camelize else name
 
-        if isinstance(x, dict):
+        if isinstance(x, dict) and x:
             for a in x:
                 flatten(x[a], name + a + ".")
-        elif isinstance(x, list):
+        elif isinstance(x, list) and x:
             for a in x:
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/generate_outputs/json_to_outputs/json_to_outputs.py
```diff
@@ -99,10 +99,10 @@ def flatten(x, name=""):
         except IndexError:
             name = name.title() if camelize else name
 
-        if isinstance(x, dict):
+        if isinstance(x, dict) and x:
             for a in x:
                 flatten(x[a], name + a + ".")
```
dorschw, line 104: instead of adding this condition, can we edit `flatten` to accept an empty input?

michal-dagan: `flatten` accepts an empty input

dorschw: I think the `and x` here can be pulled upwards, OR we can change the following line to `for a in x or ()` :)

dorschw (approved):

bziser (approved):

PR #2980, opened 2023-05-04
update demisto-py version to support python auth proxy url decoding

anara123:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-16632

## Description
related to https://github.com/demisto/demisto-py/pull/111


poetry.lock
```diff
@@ -1,101 +1,10 @@
-# This file is automatically @generated by Poetry 1.4.2 and should not be changed by hand.
```
dorschw, line 1: this line being removed is a sign for using an outdated version of poetry. Please see [here](https://panw-global.slack.com/archives/G011E63JXPB/p1681918675380719).

dorschw (approved):

poetry.lock
dorschw: I locked locally and committed to this branch

PR #2986, opened 2023-05-07
marshal graph before update the graph

ilaner:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6521

## Description
Marshalling the graph validates the graph in bucket is in the same schema of the SDK models.

### DIFF
--- a/demisto_sdk/commands/content_graph/content_graph_commands.py
+++ b/demisto_sdk/commands/content_graph/content_graph_commands.py
@@ -74,7 +74,7 @@ def update_content_graph(
             content_graph_interface, marketplace, dependencies, output_path
         )
         return
-    
+
     builder = ContentGraphBuilder(content_graph_interface)
     if not use_current:
         content_graph_interface.clean_import_dir()
--- a/demisto_sdk/commands/content_graph/content_graph_commands.py
+++ b/demisto_sdk/commands/content_graph/content_graph_commands.py
@@ -91,17 +91,8 @@ def update_content_graph(
                     content_graph_interface, marketplace, dependencies, output_path
                 )
                 return
-    content_graph_interface.import_graph(imported_path)
-    try:
-        # Test that the imported graph is valid by marshaling it
-        content_graph_interface.marshal_graph(MarketplaceVersions.XSOAR)
-        # we need to clear cache after validating the graph
-        content_graph_interface.clear_cache()
-    except ValidationError as e:
-        logger.warning(
-            "Failed to marshal the graph, probably the schema has changed. Will create a new graph"
-        )
-        logger.debug(f"Validation Error: {e}")
+    if not content_graph_interface.import_graph(imported_path):
+        # if the import failed, we need to create a new graph
         create_content_graph(
             content_graph_interface, marketplace, dependencies, output_path
         )
--- a/demisto_sdk/commands/content_graph/content_graph_commands.py
+++ b/demisto_sdk/commands/content_graph/content_graph_commands.py
@@ -75,15 +75,24 @@ def update_content_graph(
         )
         return
 
-    builder = ContentGraphBuilder(content_graph_interface)
     if not use_current:
         content_graph_interface.clean_import_dir()
         if not imported_path:
             # getting the graph from remote, so we need to clean the import dir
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/content_graph/content_graph_commands.py
```diff
@@ -81,7 +83,17 @@ def update_content_graph(
             extract_remote_import_files(content_graph_interface, builder)
 
     content_graph_interface.import_graph(imported_path)
-
+    try:
+        content_graph_interface.marshal_graph(MarketplaceVersions.XSOAR)
```
dantavori, line 87:
```suggestion
    try:
        # Test that the imported graph is valid by marshaling it
        content_graph_interface.marshal_graph(MarketplaceVersions.XSOAR)
```

dantavori (approved):

demisto_sdk/commands/content_graph/content_graph_commands.py
```diff
@@ -78,9 +80,32 @@ def update_content_graph(
...
+        content_graph_interface.clear_cache()
+    except ValidationError as e:
+        logger.warning(
+            "Failed to marshal the graph, probably the schema has changed. Will create a new graph"
+        )
+        logger.debug(f"Validation Error: {e}")
+        create_content_graph(
+            content_graph_interface, marketplace, dependencies, output_path
+        )
+        return
```
dantavori, line 108: let's move the try-except ValidationError to `import_graph` and return the success status:
```suggestion
    success = content_graph_interface.import_graph(imported_path)
    if not success:
            create_content_graph(
                content_graph_interface, marketplace, dependencies, output_path
            )
            return
```
the try-except should be called in `import_graph()` from a `is_successful_import() -> bool` method:
```
try:
    # Test that the imported graph is valid by marshaling it
    content_graph_interface.marshal_graph(MarketplaceVersions.XSOAR)
    content_graph_interface.clear_cache()
    return True
except ValidationError as e:
    logger.warning(
        "Failed to marshal the graph, probably the schema has changed. Will create a new graph"
    )
    logger.debug(f"Validation Error: {e}")
    return False
```

demisto_sdk/commands/content_graph/interface/graph.py
```diff
@@ -131,6 +131,10 @@ def validate_duplicate_ids(self, file_paths: List[str]) -> None:
     def clean_graph(self):
         ...
 
+    @abstractmethod
+    def clear_cache(self):
+        ...
+
```
dantavori, line 137: no need IMO, the cache is specific to our neo4j interface implementation

demisto_sdk/commands/content_graph/content_graph_commands.py
```diff
@@ -78,9 +80,32 @@ def update_content_graph(
...
             # getting the graph from remote, so we need to clean the import dir
-            extract_remote_import_files(content_graph_interface, builder)
-
+            try:
+                extract_remote_import_files(content_graph_interface)
+            except RuntimeError as e:
+                logger.warning(
+                    "Failed to download the content graph, will create a new graph"
+                )
+                logger.debug(f"Runtime Error: {e}")
```
dantavori, line 89: won't it show only `Failed to download the content graph` and mask the original error message?
maybe we can go with something similar to the suggestion above with the import (success = ...)

dantavori (changes requested):

demisto_sdk/commands/content_graph/interface/graph.py
ilaner: But we cannot call methods which only exist in our implementation.
How do you suggest clearing the id_to_obj cache?

demisto_sdk/commands/content_graph/content_graph_commands.py
ilaner: We will see the original error message in DEBUG, we don't care about the original message in INFO

demisto_sdk/commands/content_graph/interface/graph.py
dantavori: if it's called from within a method in the neo4j graph interface impl (as suggested above), it will work

dantavori (approved):

PR #3001, opened 2023-05-09
change precommit flags behavior

ilaner:
## Related Issues
fixes:

## Description
Change the flags to <do> / --no-<do> for the optional hooks.
In the pre-commit hook itself we will disable it

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -3251,7 +3251,7 @@ def update_content_graph(
         )
 
 
-@main.command(short_help="""Runs pre-commit hooks on the files in the repository""")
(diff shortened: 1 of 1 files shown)


demisto_sdk/__main__.py
```diff
@@ -3251,7 +3251,7 @@ def update_content_graph(
         )
 
 
-@main.command()
+@main.command(short_help="""Runs pre-commit hooks on the files in the repository""")
```
dorschw, line 3254:
```suggestion
@main.command(short_help="Runs pre-commit hooks on the files in the repository")
```

dorschw (approved):

PR #3003, opened 2023-05-09
Update in122

AradCarmi:
## Description
Fix an issue when validate fails on integration for the XSIAM marketplace that has a new advanced YML configuration.
Updated the TLP format.


dantavori (approved):
great job - approving but please resolve the one note above

PR #3022, opened 2023-05-15
run-unit-tests python2 support

ilaner:
## Related Issues
fixes:

### DIFF
--- a/demisto_sdk/commands/run_unit_tests/pytest_runner.sh
+++ b/demisto_sdk/commands/run_unit_tests/pytest_runner.sh
@@ -2,6 +2,13 @@
 set -e
 
 echo "Running tests..."
+output=$(python --version 2>&1)
+echo "Python version: $output"
 
-python -m pip install --no-cache-dir -q pytest pytest-mock requests-mock pytest-asyncio pytest-xdist pytest-datadir-ng freezegun pytest-cov hypothesis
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/run_unit_tests/pytest_runner.sh
```diff
@@ -2,6 +2,13 @@
...
 echo "Running tests..."
+output=$(python --version 2>&1)
+echo "Python version: $output"
 
-python -m pip install --no-cache-dir -q pytest pytest-mock requests-mock pytest-asyncio pytest-xdist pytest-datadir-ng freezegun pytest-cov hypothesis
+if echo "$output" | grep -q "Python 3"; then
+    additional_dependencies="pytest pytest-mock requests-mock pytest-xdist pytest-datadir-ng freezegun pytest-cov hypothesis pytest-asyncio"
+else
+    additional_dependencies="pytest pytest-mock requests-mock pytest-xdist pytest-datadir-ng freezegun pytest-cov hypothesis"
+fi
```
dorschw, line 12: perhaps it'd be cleaner if we have one list as baseline (without pytest-asyncio), and append to it if the condition holds?

ilaner: This is a temp fix, I will think about a better way to get those additional dependencies (similar as we plan to do with mypy)

PR #3033, opened 2023-05-18
modeling-Rules testing - support multi rules

ilappe:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6584

## Description
currently, the `modeling-rules` commands support modeling rules with one dataset
if there is in the *.xif file more then one dataset - all the `expected_values` are collected to one union set
and this cause the tests to fail with false positive errors

fix:
added support for multi dataset in modeling-rules


daryakoval:
Looks Good, Please ask for a review from your TL as well. :)

demisto_sdk/commands/test_content/test_modeling_rule/init_test_data.py
```diff
@@ -81,68 +81,94 @@ def init_test_data(
...
+                    new_mapping = dataset_to_fields_map.get(
+                        event_log.dataset, {}
+                    ).copy()
+                    if not new_mapping:
+                        logger.error(
+                            f"[red]Ignoring update the event log {event_log.test_data_event_id} as no dataset is provided for it[/red]",
+                            extra={"markup": True},
+                        )
+                        continue
+                    # update existing values ad remove fields from expected_values that are no longer in the rule
```
DeanArbel, line 120:
```suggestion
                    # update existing values and remove fields from expected_values that are no longer in the rule
```

DeanArbel (approved):
Good job.
@daryakoval Please take a look

PR #3065, opened 2023-05-24
Support uploading packs as custom

israelpoli:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-24134

## Description
Allows to upload packs via --input-config-file so that they are custom and are not system


ilaner (approved):
Please change the upload_result to be `int` instead of `list`, and change its value based on the `upload` return value.

ilaner (approved):

PR #3075, opened 2023-05-28
Fix upload on packs without -z

ilaner:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6781


darkushin (approved):
Nice catch!

PR #3082, opened 2023-05-30
fix multipacks upload

ilaner:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-24735

## Description

* Fixed an issue in **upload** with `--input-config-file` where the content items weren't uploaded in the correct pack.

The problem was that the mutiple packs zip (uploadable_packs.zip) was a zip with pack directories, however the platform accepts a zip of pack zips.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+
+## 1.15.3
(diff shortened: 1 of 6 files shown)


bziser:
@ilaner  please add in the description of the pr what was the problem and solution

PR #3090, opened 2023-05-31
demisto-sdk precommit fix + basic CI

ilaner:
## Related Issues
fixes:

## Description
* disable markdownlint for now
* Add require serieal to no-implicit-optional hook
* Add CI


kobymeir (approved):

demisto_sdk/commands/run_unit_tests/unit_tests_runner.py
```diff
@@ -148,6 +160,11 @@ def unit_test_runner(file_paths: List[Path], verbose: bool = False) -> int:
                 )
                 # wait for container to finish
                 if status_code := container.wait()["StatusCode"]:
+                    if status_code == 5:
```
kobymeir, line 163: @ilaner what's `5` ? why magic numbers

kobymeir (changes requested):

kobymeir (approved):

demisto_sdk/commands/pre_commit/hooks/format.py
```diff
@@ -0,0 +1,11 @@
+from pathlib import Path
+from typing import Iterable, Optional
+
+from demisto_sdk.commands.pre_commit.hooks.hook import Hook
```
kobymeir, line 4: @ilaner please document the behaviour of the `[-1]` why the hack? and i would consider creating a common class as both the FormatHook and ValidateHook share the exact same logic.

kobymeir (changes requested):

demisto_sdk/commands/pre_commit/hooks/format.py
ilaner: I combined both of the hooks to the same class as they are the same for now.
Basically, all demisto-sdk hooks should be the same as they follow the same API (-i or -g).

In addition I added a comment:
We do it because the default parameter is `-g` (run on changed files). If we want to run on input files, we should modify the `-g` parameter and append the input parameter.
We decided to have all the default values of the pre-commit in the config file instead of setting them in the code. In the code we just modify for non-default scenarios.

demisto_sdk/commands/run_unit_tests/unit_tests_runner.py
```diff
@@ -56,25 +52,46 @@ def fix_coverage_report_path(code_directory: Path):
...
         return
-    cov.combine(files, keep=True)
+    fixed_files = []
+    for file in files:
+        try:
+            fix_coverage_report_path(file)
+            fixed_files.append(file)
+        except Exception:
+            logger.warning(f"Broken .coverage file found: {file}, deleting it")
+            Path(file).unlink(missing_ok=True)
```
kobymeir, line 93: isn't it better to move the try...except logic into fix_coverage_report_path? and return None if not coverage file and leave the code here as list-comp?

kobymeir (approved):

demisto_sdk/commands/run_unit_tests/unit_tests_runner.py
ilaner: Greed, done

kobymeir (approved):

PR #3096, opened 2023-06-04
Parsers: validate that the YAML and JSON is dictionary

ilaner:
## Related Issues
fixes:

## Description
* Validate that the JSON/YAML is a dictionary
* If a parser is not found, it is not a content item and should be skipped


demisto_sdk/commands/content_graph/parsers/json_content_item.py
```diff
@@ -19,6 +20,8 @@ def __init__(
     ) -> None:
         super().__init__(path, pack_marketplaces)
         self.json_data: Dict[str, Any] = self.get_json()
+        if not isinstance(self.json_data, dict):
+            raise InvalidContentItemException("The JSON should be a dictionary")
```
dantavori, line 24:
```suggestion
            raise InvalidContentItemException(f"The content of {self.path} must be in a JSON dictionary format")
```
also in the YAML

dantavori (approved):
see note

PR #3126, opened 2023-06-07
updated versions properties for xsiam content items in graph

yucohen:
## Related Issues
fixes:

## Description
Edited the parser of XSIAM Dashboard and reports to parse the fromVersion, toVersion fields from the top level of those json files (it caused an issue in the parsing of the pack server min version that led us to install packs that should not be installed)


ilaner (approved):
Great job!

PR #3135, opened 2023-06-11
fix create_lock_files

adi88d:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6729


demisto_sdk/commands/test_content/IntegrationsLock.py
```diff
@@ -151,9 +152,13 @@ def lock_integrations(test_playbook, storage_client: storage.Client) -> bool:
...
         # If it has not expired - it means the integration is currently locked by another test.
-        workflow_id, build_number, lock_timeout = (
-            lock_file.download_as_string().decode().split(":")
-        )
+        try:
+            workflow_id, build_number, lock_timeout = (
+                lock_file.download_as_string().decode().split(":")
+            )
+        except NotFound:
+            continue
```
dorschw, line 160: let's log this, so it's easier to find when debugging

demisto_sdk/commands/test_content/IntegrationsLock.py
```diff
@@ -209,24 +214,22 @@ def create_lock_files(
...
-            test_playbook.build_context.logging_module.warning(
-                f"Could not lock integration {integration}, Create file with precondition failed."
-                f"delaying test execution."
-            )
-            unlock_integrations(locked_integrations, test_playbook, storage_client)
-            return False
-        except InvalidResponse as exc:
-            if exc.__cause__ and type(exc.__cause__) == PreconditionFailed:
+
+            if type(exception) == PreconditionFailed \
```
dorschw, line 222:
```suggestion
            if inistance(exception, PreconditionFailed) \
```

demisto_sdk/commands/test_content/IntegrationsLock.py
```diff
@@ -209,24 +214,22 @@ def create_lock_files(
...
+                    or exception.__cause__ and type(exception.__cause__) == PreconditionFailed:
```
dorschw, line 223:
```suggestion
                    or exception.__cause__ and inistance(exception.__cause__, PreconditionFailed) :
```

demisto_sdk/commands/test_content/IntegrationsLock.py
```diff
@@ -209,24 +214,22 @@ def create_lock_files(
...
                 test_playbook.build_context.logging_module.warning(
-                    f"The lock file of the integration {integration}, already created, delaying test execution."
+                    f"Could not lock integration {integration}, Create file with precondition failed."
+                    f"delaying test execution."
                 )
                 unlock_integrations(locked_integrations, test_playbook, storage_client)
                 return False
 
+            raise exception
```
dorschw, line 231:
```suggestion
            raise
```
:)

demisto_sdk/commands/test_content/IntegrationsLock.py
```diff
@@ -240,7 +243,7 @@ def unlock_integrations(
         test_playbook (TestPlaybook): The test playbook instance we want to test under the lock's context
         storage_client: The GCP storage client
     """
-    locked_integrations = [integration.name for integration in integrations_to_unlock]
+    locked_integrations = [getattr(integration, 'name', integration) for integration in integrations_to_unlock]
```
dorschw, line 246: why?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "demisto-sdk"
-version = "1.15.5"
+version = "1.15.15"
```
dorschw, line 3: ?

dorschw (changes requested):

demisto_sdk/commands/test_content/IntegrationsLock.py
adi88d: because in some cases the **integrations_to_unlock** contains strings that appends in **create_lock_files** method:
image

pyproject.toml
adi88d: ignore it was for testing

demisto_sdk/commands/test_content/IntegrationsLock.py
```diff
@@ -209,24 +216,24 @@ def create_lock_files(
...
-            if exc.__cause__ and type(exc.__cause__) == PreconditionFailed:
+
+            if isinstance(exception, PreconditionFailed) or (
+                exception.__cause__
+                and isinstance(exception.__cause__, PreconditionFailed)
+            ):
                 test_playbook.build_context.logging_module.warning(
-                    f"The lock file of the integration {integration}, already created, delaying test execution."
+                    f"Could not lock integration {integration}, Create file with precondition failed."
+                    f"delaying test execution."
```
dorschw, line 230: where does the delay happen? I only see a call to `unlock_integration` (or is this what causes the delay?)

dorschw (approved):

demisto_sdk/commands/test_content/IntegrationsLock.py
adi88d: this logs is inside `lock_integrations` method that returns True if all the test's integrations were successfully locked, else False, the line after the log is `return False` means we not able to lock the integration file this iteration.

PR #3140, opened 2023-06-13
Graph: safe get file in graph

ilaner:
## Related Issues
fixes:

## Description
This happens when content items deleted from content but the official graph in bucket (which is uploaded twice a day) contain those.

### DIFF
--- a/demisto_sdk/commands/content_graph/objects/content_item.py
+++ b/demisto_sdk/commands/content_graph/objects/content_item.py
@@ -238,6 +238,7 @@ class ContentItem(BaseContent):
     ) -> None:
         if not self.path.exists():
             logger.warning(f"Could not find file {self.path}, skipping dump")
+            return
         dir.mkdir(exist_ok=True, parents=True)
         try:
             with (dir / self.normalize_name).open("w") as f:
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/content_graph/objects/content_item.py
```diff
@@ -233,6 +236,8 @@ def dump(
         dir: DirectoryPath,
         marketplace: MarketplaceVersions,
     ) -> None:
+        if not self.path.exists():
+            logger.warning(f"Could not find file {self.path}, skipping dump")
```
dantavori, line 240: wouldn't it be raised in the exception below anyway?

demisto_sdk/commands/content_graph/objects/content_item.py
```diff
@@ -150,6 +150,9 @@ def handler(self) -> XSOAR_Handler:
 
     @property
     def data(self) -> dict:
+        if not self.path.exists():
+            logger.warning(f"Could not find file {self.path}")
+            return {}
```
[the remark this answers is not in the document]
ilaner: In a second thought returning an empty dict will just make it fail later.

This is relevant only when dumping content items, so I added a condition in the beginning of the dump.

demisto_sdk/commands/content_graph/objects/content_item.py
ilaner: If this is the case we will skip dumping and not call `self.data`.

dorschw (approved):

PR #3147, opened 2023-06-14
use symlink in lint

ilaner:
## Related Issues
fixes:

### DIFF
--- a/demisto_sdk/commands/lint/linter.py
+++ b/demisto_sdk/commands/lint/linter.py
@@ -26,7 +26,9 @@ from demisto_sdk.commands.common.constants import (
     TYPE_PWSH,
     TYPE_PYTHON,
 )
+from demisto_sdk.commands.common.content_constant_paths import CONTENT_PATH, PYTHONPATH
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/lint/linter.py
```diff
@@ -508,6 +518,12 @@ def _linter_to_commands(self):
             ),
         }
 
+    def _get_files_to_push(self):
+        files_to_push = [(self._pack_abs_dir, "/devwork")]
+        if CAN_MOUNT_FILES:
+            files_to_push.append((CONTENT_PATH, "/content"))  # type: ignore[arg-type]
```
dorschw, line 524: we can avoid this `type: ignore` by adding type hints to files_to_push

ilaner: The problems is that `self._pack_abs_dir` is `wcmpath.pathlib`` and `CONTENT_PATH` is a normal `pathlib`.
We can use Union but we need to import the regular `pathlib` for it.

PR #3153, opened 2023-06-15
Fix BA106 Validation

MichaelYochpaz:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6997

## Description
Fix BA106 validation to update the value of `self.is_valid`.
It didn't before, which caused returning `self.is_valid` to return True even if the validation failed (as long as it wasn't changed to False by a different validation beforehand).

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/content_entity_validator.py
+++ b/demisto_sdk/commands/common/hook_validations/content_entity_validator.py
@@ -170,7 +170,8 @@ class ContentEntityValidator(BaseValidator):
             True if version is valid, else False
         """
         if self.current_file.get("version") != self.DEFAULT_VERSION:
-            error_message, error_code = Errors.wrong_version(self.DEFAULT_VERSION)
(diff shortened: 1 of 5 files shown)


bziser:
From my side looks good
@YuvHayun  please review too

demisto_sdk/commands/common/hook_validations/content_entity_validator.py
```diff
@@ -575,7 +584,8 @@ def is_valid_fromversion_for_generic_objects(self):
             return True
 
         if LooseVersion(
-            self.current_file.get("fromVersion", DEFAULT_CONTENT_ITEM_FROM_VERSION)
+            self.current_file.get(
```
dorschw, line 587: this seems to happen due to different `black` configs in your local env and pre-commit. Please run `pre-commit run -a` to be able to merge

MichaelYochpaz:
Has already been fixed by https://github.com/demisto/demisto-sdk/pull/3154

PR #3154, opened 2023-06-15
fix validate xsiam entities bug 

merit-maita:
## Related Issues
fixes:

## Description
fixed an issue where validate always returned files of type xsiam dashboards and correlation rules as valid files


DeanArbel (approved):
Nice

PR #3156, opened 2023-06-15
handle custom docker for none xsoar supported packs

YuvHayun:
## Related Issues
[XSUP-25287](https://jira-hq.paloaltonetworks.local/browse/XSUP-25287)

## Description
**validate** will no longer throw errors for custom dockers in case of none xsoar supported pack.


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Removed the `timezone_offset` argument in the *modeling-rules test* command.
 * Fixed an issue where **lint** failed when importing functions from CommonServerUserPython.
 * The **format** command now will sync hidden parameters with master branch.
+* **validate** will no longer throw errors for custom dockers in case of none xsoar supported pack.
```
JudahSchwartz, line 8:
```suggestion
* **validate** will only fail on docker related errors if the pack is supported by xsoar.
```

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -58,6 +58,9 @@ def __init__(
         self.code_type = self.get_code_type()
         self.yml_docker_image = self.get_docker_image_from_yml()
         self.from_version = self.yml_file.get("fromversion", "0")
+        self.is_pack_xsoar_supported = (
+            get_pack_metadata(yml_file_path).get("support", "xsoar").lower() == "xsoar"
```
JudahSchwartz, line 62: should the default be xsoar? I imagine we would be more careful with this than customers will be. IMO should change default to cimmunity

demisto_sdk/commands/common/hook_validations/docker.py
```diff
@@ -96,7 +99,12 @@ def is_docker_image_valid(self):
             error_message, error_code = Errors.non_existing_docker(
                 self.yml_docker_image
             )
-            if self.handle_error(error_message, error_code, file_path=self.file_path):
+            if self.handle_error(
+                error_message,
+                error_code,
+                file_path=self.file_path,
+                warning=not self.is_pack_xsoar_supported,
```
JudahSchwartz, line 106: why add to warning? Whats this do?

JudahSchwartz (commented):
nice fix!

demisto_sdk/commands/common/hook_validations/docker.py
YuvHayun: It'll print this error as a warning.
I think that if we decide not to fail the validation for none-xsoar supported then we should at least warn about it.

demisto_sdk/commands/common/hook_validations/docker.py
YuvHayun: If no info is found than validate as it was xsoar supported - sounds the safest.

JudahSchwartz: are we able to assume that customers will have this field set on their own cicd pipelines? @dorschw maybe you know if we can rely on that?

JudahSchwartz (approved):
Approved pending the discussion about the default pack level. Great work

demisto_sdk/commands/common/hook_validations/docker.py
YuvHayun: Yes, there's a validate to ensure this field exists in the pack_metadata

PR #3159, opened 2023-06-18
use precommit template from content

ilaner:
move the pre-commit template to content (see https://github.com/demisto/demisto-sdk/pull/3159)


demisto_sdk/commands/common/tools.py
```diff
@@ -830,6 +830,20 @@ def get_file(
...
+    if file_path.is_absolute():
+        absolute_file_path = file_path
+        relative_file_path = file_path.relative_to(content_path)
+    else:
+        absolute_file_path = content_path / file_path
+        relative_file_path = file_path
+    try:
+        return get_file(absolute_file_path, cache_clear=cache_clear)
+    except FileNotFoundError:
+        return get_remote_file(relative_file_path)
```
dorschw, line 844:
```suggestion
        logger.debug(f"could not read/find {absolute_file_path} locally, fetching from remote"
        return get_remote_file(relative_file_path)
```

demisto_sdk/commands/common/tools.py
```diff
@@ -2006,13 +2020,13 @@ def get_content_path() -> Union[str, PathLike, None]:
 
         if not is_fork_repo and not is_external_repo:
             raise git.InvalidGitRepositoryError
-        return git_repo.working_dir
+        return Path(git_repo.working_dir).absolute()  # type: ignore[arg-type]
```
dorschw, line 2023: why ignoring `arg-type`?

demisto_sdk/commands/common/tools.py
```diff
@@ -2006,13 +2020,13 @@ def get_content_path() -> Union[str, PathLike, None]:
...
     except (git.InvalidGitRepositoryError, git.NoSuchPathError):
         if not os.getenv("DEMISTO_SDK_IGNORE_CONTENT_WARNING"):
             logger.info(
                 "[yellow]Please run demisto-sdk in content repository![/yellow]"
             )
-    return ""
+    return Path.cwd()
```
dorschw, line 2029: thank you

demisto_sdk/commands/common/tools.py
ilaner: `Path` expects `str` in `__init__`, however `git_repo.working_dir` could be `None`, `str` or `Path`.

dorschw: Can we handle the `None` case here?

ilaner: handled this case. Will return cwd in such case

demisto_sdk/commands/common/tools.py
ilaner: This make issues in tests (where we use `ChangeCWD`). Changed :(

demisto_sdk/commands/common/tools.py
```diff
@@ -830,6 +830,29 @@ def get_file(
...
 
+def get_file_or_remote(file_path: Path, clear_cache=False):
+    content_path = get_content_path()
+    relative_file_path = None
+    if file_path.is_absolute():
+        absolute_file_path = file_path
+        try:
+            relative_file_path = file_path.relative_to(content_path)
+        except ValueError:
+            logger.warning(f"Could not get relative path for {file_path}")
```
dorschw, line 841: what do we do on this case? Let's log it so someone reading the logs can understand the implication

demisto_sdk/commands/common/tools.py
```diff
@@ -830,6 +830,29 @@ def get_file(
...
+        absolute_file_path = content_path / file_path
+        relative_file_path = file_path
+    try:
+        return get_file(absolute_file_path, clear_cache=clear_cache)
+    except FileNotFoundError:
+        logger.warning(
+            f"Could not read/find {absolute_file_path} locally, fetching from remote"
+        )
+        if not relative_file_path:
+            raise
```
dorschw, line 852: let's move this before the log, **or** log a different message

dorschw, line 852: let's move this before the log, **or** log a different message

demisto_sdk/commands/common/tools.py
dorschw, line 841: what do we do on this case? Let's log it so someone reading the logs can understand the implication

ilaner: We need the relative path if we want to get the remote file from GitHub.
This function supports both absolute and relative paths

ilaner: changed to debug log and added an error message before raising

demisto_sdk/commands/common/tools.py
ilaner: added an error log

demisto_sdk/commands/content_graph/neo4j_service.py
```diff
@@ -16,7 +16,8 @@
 )
 
 REPO_PATH = CONTENT_PATH
-
+if not CONTENT_PATH.is_absolute():
+    REPO_PATH = CONTENT_PATH.absolute()
```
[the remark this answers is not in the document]
ilaner: You are correct

demisto_sdk/commands/common/tools.py
dorschw: can we remove arg-type then? :)

demisto_sdk/commands/common/tools.py
dorschw: oof

demisto_sdk/commands/common/tools.py
```diff
@@ -830,6 +830,32 @@ def get_file(
...
 
+def get_file_or_remote(file_path: Path, clear_cache=False):
+    content_path = get_content_path()
+    relative_file_path = None
+    if file_path.is_absolute():
+        absolute_file_path = file_path
+        try:
+            relative_file_path = file_path.relative_to(content_path)
+        except ValueError:
+            logger.debug(f"{file_path} is not a subpath of {content_path}")
```
dorschw, line 841: then what happens?

demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -84,20 +87,21 @@ def _get_hooks(pre_commit_config: dict) -> dict:
         for repo in pre_commit_config["repos"]:
             for hook in repo["hooks"]:
                 hooks[hook["id"]] = hook
+                # if the hook has a skip key, we add it to the SKIPPED_HOOKS set
+                if hook.pop("skip", None):
```
dorschw, line 91: why `pop` over `get`?

ilaner: Because this is not a value that `pre-commit` framework reads. we want to remove it from the template after reading it.

demisto_sdk/commands/common/tools.py
ilaner: If the file exists locally, nothing.
If it doesn't it will raise an exception later.

demisto_sdk/commands/common/tools.py
ilaner: already removed

demisto_sdk/commands/common/tools.py
dorschw: then let's add this info to the log, to make it easy for users and debuggers

ilaner: done

dorschw (approved):

PR #3162, opened 2023-06-18
Download playbook with different name and id

israelpoli:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-24134

## Description
Fixed an issue in download command when the playbook has different name and id


ilaner (approved):

demisto_sdk/commands/download/downloader.py
```diff
@@ -712,6 +723,30 @@ def build_pack_content_object(
...
+            Optional[str]: The ID of a playbook
+        """
+        endpoint = "/playbook/search"
+        playbooks = demisto_client.generic_request_func(
+            self.client,
+            endpoint,
+            "POST",
+            response_type="object",
+            body={"query": f"name:{playbook_name}"},
+        )[0]["playbooks"]
```
ilaner, line 745: let's make it safer.

```suggestion
        response = demisto_client.generic_request_func(
            self.client,
            endpoint,
            "POST",
            response_type="object",
            body={"query": f"name:{playbook_name}"},
        )
        if not response:
            return None
        if not (playbooks := response[0].get("playbooks"):
            return None
```

PR #3172, opened 2023-06-20
update default fromversion

dantavori:


demisto_sdk/commands/common/hook_validations/graph_validator.py
```diff
@@ -221,6 +221,7 @@ def handle_invalid_toversion(
         self, content_item: ContentItem, warning: bool = False
     ):
         """Handles a single invalid toversion query result"""
+        is_valid = True
```
dorschw, line 224: related to this PR?

dorschw (approved):

demisto_sdk/commands/common/hook_validations/graph_validator.py
dantavori: it's fixing a failure in a unit test that was found after changing the value of the constant.

PR #3174, opened 2023-06-20
Filter linter results

dorschw:


ilaner:
Next steps:
* Make this a script, which entirely wraps the linter with file_paths and argument.
* Add this script to the `pyproject.toml`.
* Create a new hook in `.pre-commit-config.yaml`, which calls this wrapper
* Replace the pre-commit hook in the template to call the hook from demisto-sdk instead of the official one.

PR #3176, opened 2023-06-20
IntegrationScriptUnifier: use get_file to get the code file

ilaner:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-24892?filter=-1

## Description
* Added `return_content` argument to `get_file`, to return the content of the file
* Use the `get_file` in `IntegrationScriptUnifier` to avoid unicode errors.

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -812,8 +812,7 @@ def get_file(
         raise FileNotFoundError(file_path)
 
     try:
-        file_content = _read_file(file_path)
-        if return_content:
+        if return_content := _read_file(file_path):
             return file_content
     except IOError as e:
         logger.error(f"Could not read file {file_path}.\nError: {e}")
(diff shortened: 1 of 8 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -812,6 +813,8 @@ def get_file(
 
     try:
         file_content = _read_file(file_path)
+        if return_content:
+            return file_content
```
dorschw, line 817:
```suggestion
        if return_content := _read_file(file_path):
            return file_content
```

dorschw (approved):
Please add a changelog entry before merging

demisto_sdk/commands/common/tools.py
ilaner: This is not the correct behavior.
We want to read the file anyway, and return its content only of return_content is True

dorschw: You're right, it's getting late..

PR #3182, opened 2023-06-22
Fix update additional dependencies script

ilaner:
## Related Issues
fixes:

### DIFF
--- a/demisto_sdk/scripts/update_additional_dependencies.py
+++ b/demisto_sdk/scripts/update_additional_dependencies.py
@@ -24,7 +24,7 @@ def update_additional_dependencies(
         int: 1 if failed, 0 if succeeded
     """
     try:
-        if "poetry.lock" not in GitUtil()._get_all_changed_files():
+        if Path("poetry.lock") not in GitUtil()._get_all_changed_files():
             logger.info(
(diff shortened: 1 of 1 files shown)


demisto_sdk/scripts/update_additional_dependencies.py
```diff
@@ -24,7 +24,7 @@ def update_additional_dependencies(
         int: 1 if failed, 0 if succeeded
     """
     try:
-        if "poetry.lock" not in GitUtil()._get_all_changed_files():
+        if Path("poetry.lock") not in GitUtil()._get_all_changed_files():
```
dorschw, line 27: Is this safe, considering relative paths?

ilaner: This function  _get_all_changed_files returns the relative paths that was changed

ilaner: relative to the git path, but it's the same

dorschw: Perhaps `CONTENT_PATH / "poetry.lock"` would be safer

ilaner: but it will not work

dorschw (approved):

PR #3195, opened 2023-06-26
remove default fields from validate

ilappe:
## Related Issues
related:https://jira-hq.paloaltonetworks.local/browse/CIAC-4906
content PR: https://github.com/demisto/content/pull/27616

## Description
in the content repo we removed the default fields from the YML files
so update the related fields that is used to validate files


DeanArbel (approved):
Very nice

PR #3200, opened 2023-06-27
fix error message when API key is invalid

sapirshuker:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6663

## Description
before the change:
Screenshot 2023-05-16 at 14 15 55
after the change:
image

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -2431,7 +2431,7 @@ def get_demisto_version(client: demisto_client) -> Version:
         return Version(Version(about_data.get("demistoVersion")).base_version)
     except Exception:
         logger.warning(
-            "Could not parse Xsoar version, please make sure the environment is properly configured."
+            "Could not parse server version, please make sure the environment is properly configured."
         )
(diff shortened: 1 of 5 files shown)


Shellyber (approved):
Nice job!

demisto_sdk/commands/common/tools.py
```diff
@@ -2431,7 +2431,7 @@ def get_demisto_version(client: demisto_client) -> Version:
         return Version(Version(about_data.get("demistoVersion")).base_version)
     except Exception:
         logger.warning(
-            "Could not parse Xsoar version, please make sure the environment is properly configured."
+            "Could not parse server version, please make sure the environment is properly configured."
```
dorschw, line 2434: @Shellyber perhaps we should halt the run, instead of just warning

Shellyber: We are halting the run. In the upload function below (see images in the PR body).
Here we didn't touch the code, and for this use-case I don't think it worth changing the logic here

PR #3206, opened 2023-06-29
Added tenant timezone for test modeling rules

MosheEichler:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-7232

## Description
The tenant timezone has been added in converting the epoch time to readable time.


demisto_sdk/commands/test_content/xsiam_tools/xsiam_client.py
```diff
@@ -131,6 +131,13 @@ def installed_packs(self) -> List[Dict[str, Any]]:
         response.raise_for_status()
         return response.json()
 
+    @property
+    def tenant_timezone(self) -> str:
+        endpoint = urljoin(self.base_url, "system/timezone")
```
MosheEichler, line 136: NOTE! This is not the correct endpoint

dorschw (changes requested):
Can we always use UTC, instead?

MosheEichler:
@dorschw
We can't since we want to be compatible with the tenant as much it's possible.
@eepstain FYI

eepstain:
@dorschw
Only using UTC +0000 means we don't actually testing anything for timestamp parsing with timezone calculation.

demisto_sdk/commands/test_content/test_modeling_rule/test_modeling_rule.py
```diff
@@ -145,9 +153,11 @@ def verify_results(
         # get expected_values for the given query result
         td_event_id = result.pop(f"{tested_dataset}.test_data_event_id")
         expected_values = None
+        tenant_timezone: str = ""
         for e in test_data.data:
             if str(e.test_data_event_id) == td_event_id:
                 expected_values = e.expected_values
+                tenant_timezone = e.tenant_timezone
                 break
```
dorschw, line 162: do we _always_ get the tenant_timezone, after this loop is done?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Changelog
 ## Unreleased
+* Added the tenant timezone to the **modeling-rules test** command.
```
dorschw, line 3: please add another half sentence about how it affects the runtime

demisto_sdk/commands/test_content/test_modeling_rule/test_modeling_rule.py
MosheEichler: Basically yes (just like `expected_values`) since the `tenant_timezone` is in the root such as `test_data_event_id`
Screenshot 2023-07-05 at 15 18 50

dorschw: so the empty-string default could cause trouble down the road.
Please add `if not tenant_timezone: raise ValueError("could not find timezone")`, or at least `logger.warning`

MosheEichler: @dorschw 
I can set the default to UTC

dorschw: this may go unnoticed as well

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 ## Unreleased
 * Improved memory efficiency in **update-content-graph** and **create-content-graph** commands.
+* Added the `tenant_timezone` key to the test_modeling_rule file so that you can now control the timezone in the **modeling-rules test** command.
```
dorschw, line 4: Users (and frankly, I too) don't know what the `test_modeling_rule file` is. Write a changelog to make it clear to users of what changed and why.

bziser (approved):

PR #3208, opened 2023-07-02
speed up prepare incident-to-alert scripts

israelpoli:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-7252


demisto_sdk/commands/prepare_content/preparers/incident_to_alert.py
```diff
@@ -245,6 +245,8 @@ def set_deprecated_for_scripts(data: dict, old_script: bool):
 
 
 def prepare_script_access_fields(data: dict, incident_to_alert: bool) -> dict:
+    if "incident" not in data["name"].lower():
```
dorschw, line 248: 1. Do we have a constant for this?  
2. Do we always have the `name` field?

PR #3227, opened 2023-07-05
remove dependencies, remove py484, remove get_dev_reqs

dorschw:


demisto_sdk/commands/pre_commit/tests/test_data/.pre-commit-config_template.yaml
```diff
@@ -72,6 +72,3 @@ repos:
 
       - id: secrets
         args: ["--ignore-entropy"]
-
-      - id: no-implicit-optional
-        skip: true
```
ilaner, line 77: Can you skip a differnt hook and make sure it is skipped in the test?

ilaner (approved):

PR #3296, opened 2023-07-13
Run on local docker if exist

MosheEichler:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-7400

## Description
Before pulling from remote, we check that the Docker Image does not exist locally.


dorschw (approved):
Nice

PR #3302, opened 2023-07-16
Require JSON/YAML_Handlers

dorschw:
* Added Ruff `TID251`, preventing import of `json` and `ruamel.yaml` directly (configured in `pyproject.toml`)
* Replaced all direct imports with handler imports (or, rarely, `noqa`s)

### DIFF
--- a/demisto_sdk/commands/common/handlers/json/ujson_handler.py
+++ b/demisto_sdk/commands/common/handlers/json/ujson_handler.py
@@ -24,7 +24,7 @@ class UJSON_Handler(XSOAR_Handler):
         try:
             return ujson.loads(s)
         except ValueError as e:
-            raise JSONDecodeError(f"input: {s}, error: {e}")  # type: ignore
+            raise JSONDecodeError(f"input: {s!r}, error: {e}") from e
 
     def load(self, fp: IO[str]):
         try:
--- a/demisto_sdk/commands/common/handlers/json/ujson_handler.py
+++ b/demisto_sdk/commands/common/handlers/json/ujson_handler.py
@@ -1,6 +1,6 @@
 from typing import IO, Any, AnyStr
 
-import ujson
+import ujson  # noqa: TID251 - this is the handler
 
 from demisto_sdk.commands.common.handlers.xsoar_handler import XSOAR_Handler
 
--- a/demisto_sdk/commands/test_content/timestamp_replacer.py
+++ b/demisto_sdk/commands/test_content/timestamp_replacer.py
@@ -1,11 +1,4 @@
-"""BEWARE:
-
-This file is used outside of the demisto-sdk.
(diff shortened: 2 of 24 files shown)


demisto_sdk/commands/test_content/timestamp_replacer.py
```diff
@@ -1,11 +1,4 @@
-"""BEWARE:
```
kobymeir, line 1: why the comment was removed?

demisto_sdk/commands/common/handlers/json/ujson_handler.py
```diff
@@ -24,7 +24,7 @@ def loads(self, s: AnyStr):
         try:
             return ujson.loads(s)
         except ValueError as e:
-            raise JSONDecodeError(f"input: {s}, error: {e}")  # type: ignore
+            raise JSONDecodeError(f"input: {s!r}, error: {e}") from e
```
kobymeir, line 27: why the change to repr() call?

pyproject.toml
```diff
@@ -175,4 +176,8 @@ max-complexity = 30
 
 [tool.ruff.flake8-tidy-imports]
 [tool.ruff.flake8-tidy-imports.banned-api]
-"logging".msg = "The logging framework is banned. Please use demisto_sdk.commands.common.logger."
+# "logging".msg = "The logging framework is banned. Please use demisto_sdk.commands.common.logger."
```
kobymeir, line 179: why logging is removed?

pyproject.toml
```diff
@@ -175,4 +176,8 @@ max-complexity = 30
...
+"json".msg = "use JSON_Handler instead"
+"ujson".msg = "use JSON_Handler instead"
+"orjson".msg = "use JSON_Handler instead"
+"ruamel.yaml".msg = "use YAML_Handler instead"
```
kobymeir, line 183: why not just `ruamel` will the existing line prevent something like `from ruamel import yaml` ?

kobymeir (changes requested):

demisto_sdk/commands/common/handlers/json/ujson_handler.py
dorschw: to remove the `#type: ignore`

pyproject.toml
dorschw: it was configured but not enabled (under `select`, above), so it never really ran.

Now, when I enable the rule, it fails in 20+ places on `logging`.
Opened [CIAC-7433](https://jira-hq.paloaltonetworks.local/browse/CIAC-7433) for that.

pyproject.toml
dorschw: ruff is smart enough to prevent those too.

kobymeir (approved):

PR #3319, opened 2023-07-19
fixed regex for the wrong image url detection

esharf:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-7325

## Description
Fixed the regex for the detection of wrong image URLs in README files (RM101 error).

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/readme.py
+++ b/demisto_sdk/commands/common/hook_validations/readme.py
@@ -298,7 +298,7 @@ class ReadMeValidator(BaseValidator):
             bool: True If all links are valid else False.
         """
         invalid_paths = re.findall(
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/hook_validations/readme.py
```diff
@@ -298,7 +298,7 @@ def is_image_path_valid(self) -> bool:
             bool: True If all links are valid else False.
         """
         invalid_paths = re.findall(
-            r"(\!\[.*?\]|src\=)(\(|\")(https://github.com/demisto/content/(?!raw).*?)(\)|\")",
+            r"(\!\[.*?\]|src\=)(\(|\")(https://github.com/demisto/content/blob/.*?)(\)|\")",
```
bziser, line 301: please add to the test of this validation a case that was failing before and now is ok

bziser (approved):
great!

PR #3332, opened 2023-07-19
[docker_pull] - return Image object instead of DockerClient

GuyAfik:
## Description
* Fixed an issue where in case a docker image could not be found locally, we would return DockerClient image instead of an Image object leading to possible `AttributeError` in functions using the `pull_image` function.

### DIFF
--- a/demisto_sdk/commands/common/docker_helper.py
+++ b/demisto_sdk/commands/common/docker_helper.py
@@ -131,7 +131,7 @@ class DockerBase:
     @staticmethod
     def pull_image(image: str) -> docker.models.images.Image:
         """
-        Get a docker image, in case not found, will pull it.
+        Get a local docker image, or pull it when unavailable.
         """
         docker_client = init_global_docker_client(log_prompt="pull_image")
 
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/common/docker_helper.py
```diff
@@ -131,16 +131,17 @@ def installation_files(self, container_type: str) -> FILES_SRC_TARGET:
...
         """
-        Pulling an image if it dosnt exist localy.
+        Get a docker image, in case not found, will pull it.
         """
         docker_client = init_global_docker_client(log_prompt="pull_image")
+
         try:
             return docker_client.images.get(image)
         except docker.errors.ImageNotFound:
             logger.debug(f"docker image {image} not found, pulling")
```
dorschw, line 141:
```suggestion
            logger.debug(f"docker {=image} not found locally, pulling")
```

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -131,16 +131,17 @@ def installation_files(self, container_type: str) -> FILES_SRC_TARGET:
...
             docker_client.images.pull(image)
             logger.debug(f"docker image {image} finished pulling")
```
dorschw, line 143:
```suggestion
            logger.debug(f"pulled docker {image=} successfully")
```

dorschw (approved):

PR #3339, opened 2023-07-19
Add python version for content graph

GuyAfik:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-7277

## Description
Add the **python_version** field to the graph for each integration/script as long as they use a python image.


demisto_sdk/__main__.py
```diff
@@ -3155,19 +3155,32 @@ def error_code(ctx, config, **kwargs):
     help="Whether or not to include dependencies.",
     default=False,
 )
+@click.option(
+    "-npv",
+    "--no-python-version",
```
dorschw, line 3160: please change it to be `--python-version/--no-python-version`, default=`False`, for readability.

demisto_sdk/__main__.py
```diff
@@ -3155,19 +3155,32 @@ def error_code(ctx, config, **kwargs):
...
     **kwargs,
 ):
     from demisto_sdk.commands.content_graph.content_graph_commands import (
         create_content_graph as create_content_graph_command,
     )
 
+    if not no_python_version:
+        # if no_python_version == False, need to retrieve the python version from each/script integration
+        # todo: need to think of a better way to pass in flags when creating the content graph
+        os.environ["GRAPH_GET_PYTHON_VERSION"] = "true"
```
dorschw, line 3182: Can we pass this along, instead of using an env var?

demisto_sdk/commands/common/constants.py
```diff
@@ -1845,5 +1845,7 @@ class ParameterType(Enum):
 FORMATTING_SCRIPT = "indicator-format"
 
 ENV_SDK_WORKING_OFFLINE = "DEMISTO_SDK_OFFLINE_ENV"
+DOCKERFILES_INFO_REPO = "demisto/dockerfiles-info"
+PYTHON_IMAGE_REGEX = re.compile(r"[\d\w]+/python3?:(?P<python_version>[23]\.\d+)")
```
kobymeir, line 1849: won't the python3 in the first part of the regex prevent this from working? why only in the second part you're doing python[23]

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -467,22 +478,45 @@ def _get_python_version_from_env(env: List[str]) -> Version:
...
+        return None
+
+    if python_version := _get_python_version_by_regex(image):
+        return python_version
+
+    if use_only_api:
+        return _get_python_version_from_dockerhub_api(image)
+
     try:
+        logger.debug(f"Getting python version from {image=} from client")
```
kobymeir, line 515: what's "from client"? which client are you referring to?

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -467,22 +478,45 @@ def _get_python_version_from_env(env: List[str]) -> Version:
...
         return _get_python_version_from_image_client(image)
     except Exception:
         logger.debug(
-            "Could not get the python version from client. Trying with API",
+            f"Could not get the python version for image {image=} from client. Trying with API",
```
kobymeir, line 519:
```suggestion
            f"Couldn't get the python version for image {image=} from client. Trying with API",
```

demisto_sdk/commands/common/docker_images_metadata.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class DockerImagesMetadata(Singleton, BaseModel):
+    docker_images: Dict[str, Dict[str, DockerImageTagMetadata]]
+
+    @classmethod
+    def from_github(
+        cls, file_name: str = DOCKER_IMAGES_METADATA_NAME, tag: str = "master"
+    ):
+        tag = "4c162e56174bec3ee7bb1b418ad2b20e4bdce3e0"
```
kobymeir, line 26: what's this hardcoded value?

demisto_sdk/commands/common/docker_images_metadata.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+
+    def get_docker_image_metadata_value(
+        self, docker_image: str, docker_metadata_key: str
+    ) -> Optional[str]:
+        """
+        Get the content of the requested key in the metadata
+        """
+        try:
+            docker_name, tag = docker_image.replace("demisto/", "").split(":")
```
kobymeir, line 43: why not use regex to extract the match?

demisto_sdk/commands/content_graph/objects/integration_script.py
```diff
@@ -53,3 +58,33 @@ def get_supported_native_images(
                 native_image_config=file_to_native_image_config(),
             ).get_supported_native_image_versions(get_raw_version=True)
         return []
+
+    def get_python_version(self) -> Optional[Version]:
+        """
+        Get the python version from the script/integration docker-image in case it's a python image
+        """
+        if "python" not in self.type or not self.docker_image:
```
kobymeir, line 66: are we sure all python integrations their docker image name contains python?

demisto_sdk/commands/lint/linter.py
```diff
@@ -382,8 +382,12 @@ def _gather_facts(self, modules: dict) -> bool:
             if self._facts["docker_engine"]:
                 # Getting python version from docker image - verifying if not valid docker image configured
                 for image in self._facts["images"]:
-                    py_num_version = get_python_version(image=image[0])
-                    py_num = f"{py_num_version.major}.{py_num_version.minor}"
+                    if py_num_version := get_python_version(image=image[0]):
+                        py_num = f"{py_num_version.major}.{py_num_version.minor}"
+                    else:
+                        raise RuntimeError(
```
kobymeir, line 388: is this the correct exception type to raise in this case?

demisto_sdk/commands/lint/linter.py
```diff
@@ -382,8 +382,12 @@ def _gather_facts(self, modules: dict) -> bool:
             if self._facts["docker_engine"]:
                 # Getting python version from docker image - verifying if not valid docker image configured
                 for image in self._facts["images"]:
-                    py_num_version = get_python_version(image=image[0])
-                    py_num = f"{py_num_version.major}.{py_num_version.minor}"
+                    if py_num_version := get_python_version(image=image[0]):
+                        py_num = f"{py_num_version.major}.{py_num_version.minor}"
```
kobymeir, line 386: call the variable `python_version_string` as you did in other places

demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -244,8 +243,12 @@ def group_by_python_version(files: Set[Path]) -> Dict[str, set]:
             continue
 
         code_file_path = integration_script.path.parent
-        python_version = get_python_version(integration_script.docker_image)
-        python_version_string = f"{python_version.major}.{python_version.minor}"
+        if python_version := integration_script.get_python_version():
+            python_version_string = f"{python_version.major}.{python_version.minor}"
+        else:
+            raise RuntimeError(
```
kobymeir, line 249: same Q, regarding exception type

kobymeir (changes requested):

demisto_sdk/commands/common/docker_images_metadata.py
GuyAfik: the commit of the dockerfiles-info PR, it will be the master once https://github.com/demisto/dockerfiles-info/pull/6 will be merged, at the meantime its used for testing

demisto_sdk/commands/content_graph/objects/integration_script.py
GuyAfik: yes

demisto_sdk/commands/lint/linter.py
GuyAfik: hmm, what do you think the exception should be then?

demisto_sdk/commands/pre_commit/pre_commit_command.py
GuyAfik: same Q as above

demisto_sdk/commands/common/docker_helper.py
GuyAfik: changed it to be more understandable

demisto_sdk/commands/common/constants.py
GuyAfik: its the same regex like in https://github.com/demisto/demisto-sdk/pull/3339/files#diff-6f1f6f0276b6ce9340907841f027175e315cf55853561f31c0e2920febba3674L39,

i just copied it from docker_helper to constants, however i will revert it completely

demisto_sdk/commands/lint/linter.py
dorschw: `ValueError`

demisto_sdk/commands/pre_commit/pre_commit_command.py
dorschw: same answer as above

demisto_sdk/commands/common/docker_images_metadata.py
GuyAfik: done

demisto_sdk/commands/pre_commit/pre_commit_command.py
GuyAfik: done

demisto_sdk/commands/content_graph/objects/integration_script.py
```diff
@@ -50,6 +54,18 @@ def get_supported_native_images(
             return ScriptIntegrationSupportedNativeImages(
                 _id=self.object_id,
                 docker_image=self.docker_image,
-                native_image_config=file_to_native_image_config(),
+                native_image_config=NativeImageConfig.get_instance(),
             ).get_supported_native_image_versions(get_raw_version=True)
         return []
+
+    def get_python_version(self) -> Optional[Version]:
```
JudahSchwartz, line 61: is there no way to make a lazyloadproperty decorator to make this more generic?

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -286,8 +285,12 @@ def pull_or_create_test_image(
             The test image name and errors to create it if any
         """
         errors = ""
-        if not python_version and container_type != TYPE_PWSH:
-            python_version = get_python_version(base_image).major
+        if (
+            not python_version
```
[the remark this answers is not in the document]
GuyAfik: i think its meant for lint, i just fixed mypy issues here

demisto_sdk/commands/common/constants.py
```diff
@@ -503,6 +503,7 @@ class FileType(str, Enum):
     "jliiBUXY1lWWs9Hs+PgZlCMAXNBIRoRFMH1JDZzgeyl5/MFthk+FY4FyyyEqEzxKbI2DQxMv7MMJnQZDvovVqz"
     "A1HUzK9kdzt0/2+exnQr4g2hrAAAAABJRU5ErkJggg=="
 )
+DOCKER_IMAGE_REGEX = r"^demisto/([^\s:]+):(\d+(\.\d+)*)$"
```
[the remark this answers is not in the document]
GuyAfik: in this case how can i build a regex for external repos which i do not have the option to know their names?
what are the criterias that the regex would be based upon?

demisto_sdk/commands/common/singleton.py
```diff
@@ -1,7 +1,20 @@
-class Singleton:
```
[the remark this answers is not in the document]
GuyAfik: `NativeImageConfig` class

demisto_sdk/commands/common/singleton.py
```diff
@@ -1,7 +1,20 @@
...
-            cls._instance = super().__new__(cls)
+    @classmethod
+    def get_instance(cls, *args, **kwargs):
+        if cls._instance is None:
+            cls._instance = cls.get_instance_from(*args, **kwargs)
         return cls._instance
+
+    @classmethod
+    @abstractmethod
+    def get_instance_from(cls, *args, **kwargs):
```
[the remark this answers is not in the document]
GuyAfik: in pydantic objects we most of the times want to initialize the class not from the constructor, but from a class method.

this is a general abstract method that each singelton pydantic object should inherit and implement it.  

i changed the way the singelton was implemented because it didn't work with pydnatic

demisto_sdk/commands/common/singleton.py
```diff
@@ -1,7 +1,20 @@
-class Singleton:
+from abc import abstractmethod
+
+
+class PydanticSingleton:
```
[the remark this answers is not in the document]
GuyAfik: i dont think there is builtin Singleton class as far as i know

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -467,22 +483,56 @@ def _get_python_version_from_env(env: List[str]) -> Version:
 
 
 @functools.lru_cache
-def get_python_version(image: Optional[str]) -> Version:
-    log_prompt = f"Get python version from image {image}"
-    logger.debug(f"{log_prompt} - Start")
-    if not image or "pwsh" in image or "powershell" in image:
+def get_python_version(
+    image: Optional[str], should_pull_image: bool = True
```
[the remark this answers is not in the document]
ilaner: I think @GuyAfik meant to give precedence to the dockerhub api instead of the docker client with this argument.
But I don't think we need this argument. Maybe we should use the dockerhub api first when the `DOCKERHUB_USER` and `DOCKERHUB_PASSWORD` is set (to make sure we don't rate limit.

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -467,22 +483,56 @@ def _get_python_version_from_env(env: List[str]) -> Version:
...
+            f"Getting python version from {image=} by pulling its image and query its env"
+        )
         return _get_python_version_from_image_client(image)
     except Exception:
         logger.debug(
-            "Could not get the python version from client. Trying with API",
+            f"Couldn't get the python version for image {image=} by pulling its image and query its env. Trying with API",
             exc_info=True,
         )
         return _get_python_version_from_dockerhub_api(image)
```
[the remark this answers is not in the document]
ilaner: if the DOCKERHUB_USER and DOCKERHUB_PASSWORD env variables are not set we could get rate limits.
However, if you login to the Docker with the PANW user, you will never get rate limited.

demisto_sdk/commands/common/docker_images_metadata.py
```diff
@@ -0,0 +1,87 @@
...
+
+DOCKER_IMAGES_METADATA_NAME = "docker_images_metadata.json"
+
+
+class DockerImageTagMetadata(BaseModel):
+    python_version: Optional[str]
+
+
+class DockerImagesMetadata(PydanticSingleton, BaseModel):
+    docker_images: Dict[str, Dict[str, DockerImageTagMetadata]]
```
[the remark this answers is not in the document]
ilaner: This should not be private as this object corresponds to the dockerfiles-info metadata file with validation.
If we make this private we will lose the automatic tranformation and validation from the file to the object.

demisto_sdk/commands/common/docker_images_metadata.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @classmethod
+    def get_instance_from(cls, *args, **kwargs):
+        return cls.from_github(*args, **kwargs)
+
+    @classmethod
+    def from_github(
```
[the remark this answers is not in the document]
GuyAfik: why?

demisto_sdk/commands/common/docker_images_metadata.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+        if not dockerfiles_metadata:
+            logger.error(
+                f"Could not retrieve the {DOCKER_IMAGES_METADATA_NAME} from {DOCKERFILES_INFO_REPO} repo"
+            )
+            return cls.parse_obj({"docker_images": {}})
+
+        return cls.parse_obj(dockerfiles_metadata)
+
+    def get_docker_image_metadata_value(
```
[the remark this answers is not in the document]
GuyAfik: changed to `__get_metadata_value`

demisto_sdk/commands/common/docker_helper.py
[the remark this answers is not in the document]
GuyAfik: @ilaner but i remember you said that its significant to pull the image before we try to query the version directly from the api, mainly used in the pre-commit which contains `DOCKERHUB_USER` and `DOCKERHUB_PASSWORD`.

demisto_sdk/commands/content_graph/objects/integration_script.py
```diff
@@ -50,6 +54,18 @@ def get_supported_native_images(
...
+        """
+        Get the python version from the script/integration docker-image in case it's a python image
+        """
+        if "python" not in self.type:
+            logger.debug(
+                f"The {self.content_type} = {self.object_id=} that uses {self.docker_image=} is not a python image"
+            )
+            return None
+
+        return get_python_version(self.docker_image, should_pull_image=False)
```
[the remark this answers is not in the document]
GuyAfik: to query the dockerhub api and not pull the docker image, but will be changed based on the conversation here:
https://github.com/demisto/demisto-sdk/pull/3339#discussion_r1297179519

demisto_sdk/commands/common/docker_helper.py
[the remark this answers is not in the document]
GuyAfik: @ilaner are you sure we won't get rate-limits even with `docker pull`? as far as i know you are limited in the number of images you can pull.

demisto_sdk/commands/content_graph/objects/integration_script.py
GuyAfik: pyndatic do not support it at all unfortunately https://github.com/pydantic/pydantic/issues/1035

demisto_sdk/commands/common/docker_helper.py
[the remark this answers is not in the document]
GuyAfik: btw it means that each time a person creates a content graph in local dev environment, one doesn't have  `DOCKERHUB_USER` and `DOCKERHUB_PASSWORD` configured, so in any case of misses it would just automatically pull the image.

demisto_sdk/commands/content_graph/objects/integration_script.py
ilaner: Anyway I think this should be a property

demisto_sdk/commands/common/docker_helper.py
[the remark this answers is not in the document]
ilaner: If we get the python version from the API we don't need to pull the image with the docker client, we pull it using the API. That's why we need this variables to exist to avoid pull rate limit when pulling the docker image from the API

demisto_sdk/commands/common/constants.py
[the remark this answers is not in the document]
JudahSchwartz: it can be anything. why need a specific match?
If they give me 
asdasdasdad
then we will try to run 
```
docker inspect asdasdasdad
docker pull asdasdasdad
```
etc

demisto_sdk/commands/common/docker_helper.py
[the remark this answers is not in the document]
JudahSchwartz: either way this is logic which should be done from inside the function and it should be consistent. No need to make this an option from outside

demisto_sdk/commands/common/docker_helper.py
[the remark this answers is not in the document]
JudahSchwartz: i think we should try catch for rate limit but still always try it first. Is there an issue with that @ilaner ? The potential loss is small (an api call) and the gain is saving a pull we dont need

demisto_sdk/commands/content_graph/objects/integration_script.py
dantavori: it might look more elegant as a property, but every time we create an integration script object it will be calculated and sometimes we don't need it (e.g. prepare content). I understood it adds 30 seconds to the graph calculation, so I'm in favor of avoiding these calculations where it's not needed.

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -120,6 +120,25 @@ def normalize_name(self) -> str:
...
+    def __add_metadata(self, json_dct: Dict):
+        """
+        Add extra metadata to the model.
+
+        Args:
+            json_dct: dict representation of the object
+
+        """
+        if "path" in json_dct and Path(json_dct["path"]).is_absolute():
+            json_dct["path"] = (Path(json_dct["path"]).relative_to(CONTENT_PATH)).as_posix()  # type: ignore
```
dantavori, line 132: why is it needed?

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -120,6 +120,25 @@ def normalize_name(self) -> str:
...
+
+        json_dct["content_type"] = self.content_type
+
+        if self.content_type in (ContentType.SCRIPT, ContentType.INTEGRATION):
+            python_version = self.get_python_version()  # type: ignore[attr-defined]
```
dantavori, line 137: maybe we can override IntegrationScript's to_dict() instead?

demisto_sdk/commands/content_graph/objects/base_content.py
GuyAfik: good question

demisto_sdk/commands/content_graph/objects/base_content.py
GuyAfik: implemented it differently

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -36,7 +38,9 @@
     )
 )
 
-PYTHON_IMAGE_REGEX = re.compile(r"[\d\w]+/python3?:(?P<python_version>[23]\.\d+)")
+PYTHON_IMAGE_REGEX = re.compile(
```
JudahSchwartz, line 41: lets call this demisto_python_base_regex

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -400,6 +408,14 @@ def get_docker():
     return MountableDocker() if CAN_MOUNT_FILES else DockerBase()
 
 
+def _get_python_version_by_regex(image: str) -> Optional[Version]:
```
JudahSchwartz, line 411:
```suggestion
def _get_python_version_from_tag_by_regex(image: str) -> Optional[Version]:
```

demisto_sdk/commands/content_graph/common.py
```diff
@@ -281,6 +281,36 @@ class PackTags:
...
+
+    pass
+
+
+def lazy_property(property_func: Callable):
+    """
+    lazy property: specifies that this property should be added to the pydantic model lazily
+    only when the instance property is first accessed.
+
+    Use this decorator on your property in case you need it to be added to the model only if its called directly
```
JudahSchwartz, line 297: reminder to add that can only return json serializable data

demisto_sdk/commands/content_graph/tests/parsers_and_models_test.py
```diff
@@ -1570,6 +1570,50 @@ def test_repo_parser(self, repo: Repo):
...
+            parser.relationships,
+            integration_commands=["test-command"],
+            imports=["MicrosoftApiModule"],
+            tests=["test_playbook"],
+        )
+        model = Integration.from_orm(parser)
+        # make sure that the python_version is not in the model because it was called directly
+        assert "python_version" not in str(model)
+
+        assert model.python_version == expected_python_version
```
JudahSchwartz, line 1613: Nice test

JudahSchwartz (approved):
Noice

demisto_sdk/commands/common/docker_images_metadata.py
```diff
@@ -0,0 +1,101 @@
...
+)
+from demisto_sdk.commands.common.git_content_config import GitContentConfig
+from demisto_sdk.commands.common.logger import logger
+from demisto_sdk.commands.common.singleton import PydanticSingleton
+from demisto_sdk.commands.common.tools import get_remote_file_from_api
+
+DOCKER_IMAGES_METADATA_NAME = "docker_images_metadata.json"
+
+# regex to extract docker-images that are specific to content / dockerfiles
+DOCKER_IMAGE_REGEX = r"^demisto/([^\s:]+):(\d+(\.\d+)*)$"
```
kobymeir, line 19: are there images not starting with 'demisto/'? 
such as here:
https://github.com/demisto/demisto-sdk/pull/3339/files#diff-3ff89907faea6587886223530c2f5d6bba10bda00f22b4a610481f359fc96ef4R27

demisto_sdk/commands/common/tests/docker_helper_test.py
```diff
@@ -39,6 +40,21 @@ def __init__(self, attrs):
             self.attrs = attrs
 
     mocker.patch.object(docker_helper, "init_global_docker_client")
+    mocker.patch(
```
kobymeir, line 43: a powershell test case is needed, to ensure we don't populate that data.

kobymeir (changes requested):

demisto_sdk/commands/common/docker_images_metadata.py
GuyAfik: this regex is used to extract only docker-images related to **content** and not custom docker-images.
See the comment in line 18.

I have changed the name of the variable so it would be more clear.

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -467,25 +483,49 @@ def _get_python_version_from_env(env: List[str]) -> Version:
 
 
 @functools.lru_cache
-def get_python_version(image: Optional[str]) -> Version:
-    log_prompt = f"Get python version from image {image}"
-    logger.debug(f"{log_prompt} - Start")
-    if not image or "pwsh" in image or "powershell" in image:
+def get_python_version(image: Optional[str]):
```
dantavori, line 486:
```suggestion
def get_python_version(image: Optional[str]) -> Optional[Version]:
```

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -467,25 +483,49 @@ def _get_python_version_from_env(env: List[str]) -> Version:
...
+
+    Returns:
+        Version: Python version X.Y (3.7, 3.6, ..)
+    """
+    logger.debug(f"Get python version from image {image=}")
+
+    if not image:
         # When no docker_image is specified, we use the default python version which is Python 2.7.18
         logger.debug(
             f"No docker image specified or a powershell image, using default python version: {DEFAULT_PYTHON2_VERSION}"
```
dantavori, line 501: when will we have `image == None` and when will `"pwsh" in image or "powershell" in image` be True? why in the first case we return `Version(DEFAULT_PYTHON2_VERSION)` and in the second one `None`?

demisto_sdk/commands/content_graph/common.py
```diff
@@ -281,6 +281,39 @@ class PackTags:
...
+    lazy property: specifies that this property should be added to the pydantic model lazily
+    only when the instance property is first accessed.
+
+    Note:
+        make sure that the lazy property returns only serializable objects.
+
+    Use this decorator on your property in case you need it to be added to the model only if its called directly
+    """
+
+    def _lazy_decorator(self):
```
dantavori, line 303: what is the meaning of "self" outside a class scope?

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -69,6 +73,14 @@ def __new__(
         if content_type:
             content_type_to_model[content_type] = model_cls
             model_cls.content_type = content_type
+        # validates that if there are any lazy properties in the class model, add them as a class attribute
+        # so we would be able to load them during graph creation
```
dantavori, line 77: please add to docstring instead.
also for consistency let's use `model_cls` below

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -128,11 +148,13 @@ def to_dict(self) -> Dict[str, Any]:
         Returns:
             Dict[str, Any]: _description_
         """
+        self.__add_lazy_properties()
```
dantavori, line 151: same here, please explain shortly why in the docstring. I'd specifically mention that to_dict is used to create the graph nodes, which is when we want to include the lazy properties.

demisto_sdk/commands/content_graph/objects/integration_script.py
```diff
@@ -26,6 +30,22 @@ class IntegrationScript(ContentItem):
...
+    @lazy_property
+    def python_version(self) -> Optional[str]:
+        """
+        Get the python version from the script/integration docker-image in case it's a python image
+        """
+        if "python" not in self.type:
+            logger.debug(
+                f"The {self.content_type} = {self.object_id=} that uses {self.docker_image=} is not a python image"
+            )
+            return None
```
dantavori, line 42: if not already handled in `get_python_version`, it should

dantavori (commented):
nice, see (mainly cosmetic) notes below

demisto_sdk/commands/common/docker_helper.py
GuyAfik: `image = None` is for cases where we have scripts/integrations that are based on javascript, @ilaner do you remember why we set the `DEFAULT_PYTHON2_VERSION` here? 

I think its related to the platform if i recall

demisto_sdk/commands/content_graph/common.py
GuyAfik: this decorator decorates methods of the `BaseContent` class (properties), hence the self here refers to the `self` of the decorator method from the `BaseContent`

demisto_sdk/commands/content_graph/objects/integration_script.py
GuyAfik: you are absolutely right, this is already handled in `get_python_version`, hence removing it.

demisto_sdk/commands/content_graph/objects/base_content.py
GuyAfik: what do you mean by "let's use model_cls below"?

demisto_sdk/commands/common/docker_helper.py
ilaner: > `image = None` is for cases where we have scripts/integrations that are based on javascript, hence they do not use any docker image, @ilaner do you remember why we set the `DEFAULT_PYTHON2_VERSION` here?
> 
> I think its related to the platform if i recall

If the docker image is missing, the server will run the code in docker image which based on python 2 image.

demisto_sdk/commands/content_graph/common.py
```diff
@@ -281,6 +281,39 @@ class PackTags:
...
+    pass
+
+
+def lazy_property(property_func: Callable):
+    """
+    lazy property: specifies that this property should be added to the pydantic model lazily
+    only when the instance property is first accessed.
+
+    Note:
+        make sure that the lazy property returns only serializable objects.
```
ilaner, line 298: primitive instead of serializable (str, int, bool, list)

kobymeir (approved):

PR #3349, opened 2023-07-19
[content-graph] - add deprecated attribute to pack

GuyAfik:
## Description
Add the flag `deprecated` also to pack as packs can be deprecated as well and not only content items.
pack is considered deprecated if all its content items are deprecated as well.


demisto_sdk/commands/content_graph/tests/parsers_and_models_test.py
```diff
@@ -196,6 +196,7 @@ def run(
         expected_preview_only: Optional[bool] = None,
         expected_marketplaces: Optional[List[MarketplaceVersions]] = None,
         expected_content_items: Dict[str, ContentType] = {},
+        expected_deprecated: Optional[bool] = None
```
dorschw, line 199: Optional because its default is `False`?

GuyAfik: optional because its a common function between multiple tests to check that content parsing objects work as expected, in some of the tests we do not want to check them up.

The value of the `deprecated` will be determined by whether the pack name ends with `(Deprecated)`.

demisto_sdk/commands/content_graph/parsers/pack.py
```diff
@@ -208,3 +215,12 @@ def parse_content_item(self, content_item_path: Path) -> None:
...
             raise
+
+    @property
+    def deprecated(self) -> bool:
+        if regex.match(PACK_NAME_DEPRECATED_REGEX, self.name) and (
+            regex.match(DEPRECATED_NO_REPLACE_DESC_REGEX, self.description)
+            or regex.match(DEPRECATED_DESC_REGEX, self.description)
+        ):
+            return True
+        return False
```
kobymeir, line 226:
```suggestion
        return regex.match(PACK_NAME_DEPRECATED_REGEX, self.name) and (
            regex.match(DEPRECATED_NO_REPLACE_DESC_REGEX, self.description)
            or regex.match(DEPRECATED_DESC_REGEX, self.description)
        )
```

kobymeir (changes requested):

demisto_sdk/commands/content_graph/parsers/pack.py
GuyAfik: done

dantavori:
where is the deprecated field used?

GuyAfik:
> Fail validate dependency packs #3365

will be used [here](https://github.com/demisto/demisto-sdk/pull/3365) in the query to for the graph to make sure we get dependent packs which are not deprecated **only**.

kobymeir (approved):

ilaner:
Looks good, buy maybe we should add this value to pack_metadata.json?

kobymeir:
@ilaner i object, as it duplicates the field in the YML with no good reason.

dantavori:
@GuyAfik I mean, when do we use the "deprecated" field in packs? I see we have some, but when it's set to "true", how does it affect content/server?

GuyAfik:
> @GuyAfik I mean, when do we use the "deprecated" field in packs? I see we have some, but when it's set to "true", how does it affect content/server?

lets talk in slack instead of here

PR #3351, opened 2023-07-19
Graph fix

eyalpalo:
## Description
Added unit test to validate that the graph created don't have is_test param of dependencies with value None


demisto_sdk/commands/content_graph/tests/create_content_graph_test.py
```diff
@@ -897,3 +897,34 @@ def test_create_content_graph_incident_to_alert_scripts(
...
         assert (script_path / "script-setIncident.yml").exists()
         assert not (script_path / "script-setAlert.yml").exists()
+
+    def test_create_content_graph_relationships_from_metadata(
+            self,
+            repo: Repo,
+    ):
+        """
+        Given:
+            - A mocked model of a repository with a pack Core, which dependce in NonCorePack
```
dantavori, line 907:
```suggestion
            - A mocked model of a repository with a pack Core, which depends on NonCorePack according to the pack metadata
```

dantavori (approved):

PR #3355, opened 2023-07-19
Add a validation to disallow use of internal terms in customer-facing docs

MichaelYochpaz:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-6459

## Description
Customers may not be familiar with internal technical terms like "test-module", since the function is invisible to them and they know it as the "test button".
This validation insures such terms aren't in use in customer-facing docs.


demisto_sdk/commands/common/hook_validations/base_validator.py
```diff
@@ -421,3 +422,46 @@ def validate_xsiam_content_item_title(file_path):
...
+        Returns:
+            bool: True if no such terms were found, False otherwise.
+        """
+        disallowed_terms = (
+            [  # These terms are checked regardless for case (case-insensitive)
+                "test-module",
+                "test module",
+                "long-running-execution",
+            ]
+        )
```
JasBeilin, line 446: It would be better practice to put this list in a different file/ const then the code, it might grow significally with time.

demisto_sdk/commands/common/hook_validations/release_notes.py
```diff
@@ -536,5 +536,9 @@ def is_file_valid(self):
             self.validate_json_when_breaking_changes(),
             # self.has_no_markdown_lint_errors(),
             self.validate_release_notes_headers(),
+            self.validate_no_disallowed_terms_in_customer_facing_docs(
+                file_content=self.latest_release_notes,
+                file_path=self.release_notes_file_path,
```
JasBeilin, line 541: Are we sure we want to send the file content if we are already sending the path?

JasBeilin (changes requested):
Great job. Please check out my notes.

demisto_sdk/commands/common/hook_validations/release_notes.py
MichaelYochpaz: The validation actually runs on the content, `file_path` is just used for the error handler (to print which file failed the validation).

I think it'll be bad optimization-wise to reopen & load it to a string again within the validation function.

demisto_sdk/commands/common/hook_validations/base_validator.py
MichaelYochpaz: That's a good point, but I think it'll be more fitting to just move it once and if it gets large, instead of moving it now to a file with just 3 terms, which will make it slightly less readable (having to refer to another file to see what the disallowed terms are).

The ticket I opened has just 3 additional potential terms that might be added.

JasBeilin: I would use const at least, but lets close it and handle that in the future.

JasBeilin (approved):

demisto_sdk/commands/validate/tests/validators_test.py
```diff
@@ -3048,3 +3052,97 @@ def test_run_validation_using_git_validation_calls(
...
+
+    validate_manager = ValidateManager()
+
+    # Mock all validations that aren't 'validate_no_disallowed_terms_in_customer_facing_docs' to return True
+    # mocker.patch.object(ReleaseNotesValidator, "has_release_notes_been_filled_out", return_value=True)
+    # mocker.patch.object(ReleaseNotesValidator, "are_release_notes_complete", return_value=True)
+    # mocker.patch.object(ReleaseNotesValidator, "is_docker_image_same_as_yml", return_value=True)
+    # mocker.patch.object(ReleaseNotesValidator, "validate_json_when_breaking_changes", return_value=True)
+    # mocker.patch.object(ReleaseNotesValidator, "has_no_markdown_lint_errors", return_value=True)
+    # mocker.patch.object(ReleaseNotesValidator, "validate_release_notes_headers", return_value=True)
```
dorschw, line 3131: Please remove these if unused.

PR #3395, opened 2023-07-23
prepare-content, add input options

darkushin:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-7067

## Description
- Add the '-a' option to the prepare-content SDK command, to allow preparing all content packs for upload.
- Support multiple arguments through the '-i' option.


demisto_sdk/__main__.py
```diff
@@ -417,25 +421,51 @@ def prepare_content(ctx, **kwargs):
...
-        os.environ[ENV_DEMISTO_SDK_MARKETPLACE] = marketplace.lower()
-    if file_type == FileType.GENERIC_MODULE:
-        from demisto_sdk.commands.prepare_content.generic_module_unifier import (
-            GenericModuleUnifier,
+    assert any(
+        [kwargs["all"], kwargs["input"]]
+    ), "Either '-a' or '-i' parameters must be provided."
+    assert not all(
+        [kwargs["all"], kwargs["input"]]
+    ), "Only one of '-a' and '-i' parameters should be provided."
```
dorschw, line 429: we can `assert sum(kwargs["all"], kwargs["input"]), "either all or input must be provided"`, a bit cleaner :)

demisto_sdk/__main__.py
```diff
@@ -417,25 +421,51 @@ def prepare_content(ctx, **kwargs):
...
+
+    if kwargs["all"]:
+        content_DTO = from_path()
+        content_DTO.dump(
+            dir=Path(os.path.join(kwargs["output"], "tmp")),
+            marketplace=MarketplaceVersions(kwargs["marketplace"]),
```
dorschw, line 435: please use `parse_marketplace_kwargs`

demisto_sdk/__main__.py
```diff
@@ -417,25 +421,51 @@ def prepare_content(ctx, **kwargs):
...
+        return 0
 
-        # pass arguments to GenericModule unifier and call the command
-        generic_module_unifier = GenericModuleUnifier(**kwargs)
-        generic_module_unifier.merge_generic_module_with_its_dashboards()
-    else:
-        PrepareUploadManager.prepare_for_upload(**kwargs)
+    inputs = []
+    if kwargs["input"]:
+        inputs = kwargs["input"].split(",")
```
dorschw, line 441:
```suggestion
    if (input_ := kwargs["input"]):
        inputs = input_.split(",")
```

demisto_sdk/__main__.py
```diff
@@ -417,25 +421,51 @@ def prepare_content(ctx, **kwargs):
...
+
+    output_path = kwargs["output"]
+
+    for input in inputs:
```
dorschw, line 445: input is a saved word in python, please use something else.

demisto_sdk/__main__.py
```diff
@@ -417,25 +421,51 @@ def prepare_content(ctx, **kwargs):
...
+        if output_path and len(inputs) > 1:
+            path_name = os.path.basename(input)
```
dorschw, line 447: please use Pathlib over os.path.

demisto_sdk/__main__.py
```diff
@@ -417,25 +421,51 @@ def prepare_content(ctx, **kwargs):
...
+
+        if click.get_current_context().info_name == "unify":
+            kwargs["unify_only"] = True
+
+        check_configuration_file("unify", kwargs)
+        # Input is of type Path.
+        kwargs["input"] = str(input)
+        file_type = find_type(kwargs["input"])
+        if marketplace := kwargs.get("marketplace"):
+            os.environ[ENV_DEMISTO_SDK_MARKETPLACE] = marketplace.lower()
```
dorschw, line 458: can we avoid writing to the env var, by passing the marketplace?

demisto_sdk/commands/content_graph/objects/repository.py
```diff
@@ -11,10 +11,17 @@
 from demisto_sdk.commands.common.cpu_count import cpu_count
 from demisto_sdk.commands.common.logger import logger
 from demisto_sdk.commands.content_graph.objects.pack import Pack
+from demisto_sdk.commands.content_graph.parsers.repository import RepositoryParser
 
 USE_MULTIPROCESSING = False  # toggle this for better debugging
 
 
+def from_path():
```
dorschw, line 19: please document this function

demisto_sdk/__main__.py
darkushin: -i is a string and not a boolean so it can't be summed :(
Unless converting it first to a boolean which won't look nice again. Also, would need to check that sum == 1 as we don't want to allow both parameters.

demisto_sdk/__main__.py
darkushin: Not sure about that, these are parts I didn't touch so I didn't want to make any changes to them

demisto_sdk/__main__.py
dorschw: `sum(bool(kwargs["all"])),bool(kwargs["input"]))) == 1` isn't too ugly, and is less code, so I recommend but won't insist 🥲

demisto_sdk/__main__.py
```diff
@@ -423,25 +427,58 @@ def prepare_content(ctx, **kwargs):
...
+        inputs = input_.split(",")
+
+    if output_path := kwargs["output"]:
+        if "." in Path(output_path).name:  # check if the output path is a file
+            if len(inputs) > 1:
+                raise Exception(
+                    "When passing multiple inputs, the output path should be a directory and not a file."
+                )
+        else:
+            os.makedirs(output_path, exist_ok=True)
```
dorschw, line 456: create a path object for the destination path, then use `dest_path.parent.makedirs(exist_ok=True)`

demisto_sdk/__main__.py
darkushin: no problem, updated

demisto_sdk/__main__.py
```diff
@@ -423,30 +430,63 @@ def prepare_content(ctx, **kwargs):
...
-        generic_module_unifier = GenericModuleUnifier(**kwargs)
-        generic_module_unifier.merge_generic_module_with_its_dashboards()
-    else:
-        PrepareUploadManager.prepare_for_upload(**kwargs)
+    inputs = []
+    if input_ := kwargs["input"]:
+        inputs = input_.split(",")
+
+    if output_path := kwargs["output"]:
+        if "." in Path(output_path).name:  # check if the output path is a file
```
dorschw, line 451:
```suggestion
        if Path(output_path).is_file():
```

darkushin: At this point, the file doesn't exist yet, so using `Path.is_fiile()` will always return false.

demisto_sdk/commands/content_graph/objects/repository.py
```diff
@@ -11,10 +11,20 @@
...
```
dantavori, line 19: name is unclear, which path?
I think, either allow providing it or call it `all_content_repo()`

demisto_sdk/__main__.py
```diff
@@ -370,9 +371,15 @@ def extract_code(ctx, config, **kwargs):
 @click.option(
     "-i",
     "--input",
-    help="The directory path to the files or path to the file to unify",
-    required=True,
-    type=click.Path(dir_okay=True),
+    help="The directory path to the files or path to the file to unify. Supports a comma separated list.",
```
dorschw, line 374:
```suggestion
    help="Comma-separated list of paths to directories or files to unify.",
```

PR #3421, opened 2023-07-27
Add a CLI command log

MichaelYochpaz:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-7586

## Description
Add a log (visible only in log files) that specifies what command was used, with the complete arguments.
Quotes are added to arguments that contain spaces to know that they're a single arguments and not multiple separate ones.

Example output:
```
[2023-07-27T11:47:26] - [MainThread] - [DEBUG] - logger.py:332 - CLI command: demisto-sdk download -i "Michael Test" -o Packs/MichaelTest
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * Added a new validation (`BA125`) to **validate** that assures internal function names aren't in use in customer-facing docs.
 * Fixed an issue in **update-content-graph** where the neo4j service was unaccessible for non-root users.
 * Added a Sourcery hook to **pre-commit**.
+* Added a log specifying the SDK command that's in use in log files.
 
 ## 1.17.2
 * Fixed an issue where **lint** and **validate** commands failed on integrations and scripts that use docker images that are not available in the Docker Hub but exist locally.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 * Added a new validation (`BA125`) to **validate** that assures internal function names aren't in use in customer-facing docs.
 * Fixed an issue in **update-content-graph** where the neo4j service was unaccessible for non-root users.
 * Added a Sourcery hook to **pre-commit**.
-* Added a log specifying the SDK command that's in use in log files.
+* Added a log of the CLI command used when running demisto-sdk.
 
 ## 1.17.2
 * Fixed an issue where **lint** and **validate** commands failed on integrations and scripts that use docker images that are not available in the Docker Hub but exist locally.
--- a/demisto_sdk/commands/common/logger.py
+++ b/demisto_sdk/commands/common/logger.py
@@ -328,7 +328,9 @@ def logging_setup(
 
     set_demisto_logger(demisto_logger)
 
-    cli_args = [f'"{arg}"' if " " in arg else arg for arg in sys.argv[1:]]  # Add quotes to spaced arguments
+    cli_args = [
+        f'"{arg}"' if " " in arg else arg for arg in sys.argv[1:]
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 * Added a new validation (`BA125`) to **validate** that assures internal function names aren't in use in customer-facing docs.
 * Fixed an issue in **update-content-graph** where the neo4j service was unaccessible for non-root users.
 * Added a Sourcery hook to **pre-commit**.
+* Added a log specifying the SDK command that's in use in log files.
```
dorschw, line 29:
```suggestion
* Added logging of the called cli command.
```

demisto_sdk/commands/common/logger.py
```diff
@@ -328,6 +328,8 @@ def logging_setup(
 
     set_demisto_logger(demisto_logger)
 
+    cli_args = [f'"{arg}"' if " " in arg else arg for arg in sys.argv[1:]]  # Add quotes to spaced arguments
```
dorschw, line 331: `[1:]` because 0th is the path/python? I can imagine a situation where it could be useful. How dirty is it if we don't slice?

dorschw (approved):

demisto_sdk/commands/common/logger.py
MichaelYochpaz: When running from the repository, `argv[0]` is `/Users/username/dev/demisto/demisto-sdk/demisto_sdk/__main__.py`

When running from a compiled package (I installed from a test branch):
`/Users/username/Library/Caches/pypoetry/virtualenvs/demisto-sdk-686ZqX68-py3.10/bin/demisto-sdk`

We can do `Path(arg).name`, but I don't see the benefit of that, and we'll get `__main__.py` when running from code (not sure if that's good or bad, don't see a case where it matters, honestly)

CHANGELOG.md
MichaelYochpaz:
```suggestion
* Added a log of the CLI command used to run demisto-sdk.
```

demisto_sdk/commands/common/logger.py
dorschw: the python path isn't a bad thing to have in the logs, we have users working with multiple envs 
If you rather keep this log line relatively short, consider logging the `[0]` in a separate line

CHANGELOG.md
dorschw:
```suggestion
* Added a log of the CLI command used when running demisto-sdk.
```

demisto_sdk/commands/common/logger.py
```diff
@@ -328,6 +328,8 @@ def logging_setup(
...
+    demisto_logger.debug(f"CLI command: demisto-sdk {' '.join(cli_args)}")
     demisto_logger.debug(f"Python version: {sys.version}")
```
MichaelYochpaz, line 333:
```suggestion
    demisto_logger.debug(f"CLI command: demisto-sdk {' '.join(cli_args)}")
    demisto_logger.debug(f"Executable location: {sys.argv[0]}")
    demisto_logger.debug(f"Python version: {sys.version}")
```
@dorschw WDYT?

tkatzir:
Is there a way that this could leak some secrets like passwords, keys, usernames, local (internal) paths, etc.?

tkatzir:
> 

Check IntegrationLogger in CommonServerPython in Content.

A log file is sometimes sent out for support (e.g. when/if an issue is opened.)
The bash history file is never sent out of an org.

PR #3422, opened 2023-07-27
added sha1 validation for content graph parser

esharf:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-7520

## Description
Added a validation that content_graph was created with the current version of the parser.
if not it will create the graph from scratch.

### DIFF
--- a/demisto_sdk/commands/content_graph/interface/neo4j/neo4j_graph.py
+++ b/demisto_sdk/commands/content_graph/interface/neo4j/neo4j_graph.py
@@ -4,7 +4,6 @@ from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
 
 from neo4j import Driver, GraphDatabase, Session, graph
-from pydantic import ValidationError
 
 import demisto_sdk.commands.content_graph.neo4j_service as neo4j_service
 from demisto_sdk.commands.common.constants import MarketplaceVersions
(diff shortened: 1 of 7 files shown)


demisto_sdk/commands/content_graph/interface/neo4j/neo4j_graph.py
```diff
@@ -539,17 +538,9 @@ def import_graph(self, imported_path: Optional[Path] = None) -> bool:
...
-            self.marshal_graph(MarketplaceVersions.XSOAR)
-            result = True
-        except ValidationError as e:
-            logger.warning("Failed to import the content graph")
-            logger.debug(f"Validation Error: {e}")
-            result = False
-        # clear cache after loading the graph
+        has_infra_graph_been_changed = self._has_infra_graph_been_changed()
         self._id_to_obj = {}
-        return result
```
thefrieddan1, line 552: is it possible to :
self._id_to_obj = {}
return self._has_infra_graph_been_changed()

thefrieddan1 (approved):
Approved with non mandatory comment

demisto_sdk/commands/content_graph/interface/neo4j/neo4j_graph.py
esharf: I agree with your point.
but I didn't want to change it because I'm afraid of changing unnecessary things here.

ilaner (approved):
Great! Can you try to edit a file in `parsers` in this PR to make sure that the graph is created from scratch in the CI?

PR #3450, opened 2023-08-08
remove private items

dantavori:
## Related Issues
relates: https://jira-hq.paloaltonetworks.local/browse/CIAC-7149


demisto_sdk/commands/content_graph/common.py
```diff
@@ -770,3 +770,59 @@ def update(self, other: "Nodes") -> None:  # type: ignore[override]
         "fcm",
     ],
 }
+
+
+CONTENT_PRIVATE_ITEMS: dict = {
```
ShahafBenYakir, line 775: Let's add a comment that this is only temporary and attach a Jira ticket with the proper solution.

ShahafBenYakir (approved):
Looks good. see my comment.

PR #3454, opened 2023-08-09
Ignore .venv dir when using format

DinaMeylakh:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-7756

## Description
Files inside .venv directories should not be formatted.

## Local checks:
added a .venv dir with invalid some_stuff.yml to HelloWorld pack:
Before fix:
Screenshot 2023-08-09 at 14 30 51
After fix: No line containing this file name.


demisto_sdk/commands/format/format_constants.py
```diff
@@ -53,3 +53,5 @@
     FileType.GENERIC_DEFINITION,
 ]
 OLD_FILE_TYPES = [FileType.LAYOUT.value, FileType.OLD_CLASSIFIER.value]
+
+SKIP_FORMATTING_DIRS = [".venv"]
```
ShahafBenYakir, line 57: Why is ".venv" the only problem? we have many files we do not want to format inside directories ( denistomock, CSP, etc...) what am I missing here?

DinaMeylakh: these are usually handled via the excluded_files parameter

from format_manager:
```python
from demisto_sdk.commands.lint.commands_builder import excluded_files
...
...
        for file in files:
            file_path = str(Path(file))
            file_type = find_type(file_path, clear_cache=clear_cache)

            current_excluded_files = excluded_files[:]
            dirname = os.path.dirname(file_path)
            if dirname.endswith("CommonServerPython"):
                current_excluded_files.remove("CommonServerPython.py")
            if os.path.basename(file_path) in current_excluded_files:
                continue
            if any(test_dir in str(dirname) for test_dir in TESTS_AND_DOC_DIRECTORIES):
                continue
```
where 
```python
excluded_files = [
    "CommonServerPython.py",
    "demistomock.py",
    "CommonServerUserPython.py",
    "conftest.py",
    ".venv",
]
```
but since the basename of the .venv/* files is not .venv this does not hold. 
since the excluded_files are also used in other places I did not want to change this. 
and if you have a file with a name of .venv, you probably do not need to format it.

In my implementation, we get rid of these files before it even gets to this, also the .venv directory is discarded no matter where your .venv dir is located (in PackName/ or in Integrations/IntegrationName or anywhere inside the path of the -i option)

demisto_sdk/commands/format/format_module.py
```diff
@@ -203,14 +213,16 @@ def format_manager(
...
+            # Check if this is an unskippable file
+            if not any(
+                [
+                    file_path.endswith(unskippable_file)
+                    for unskippable_file in UNSKIP_FORMATTING_FILES
+                ]
+            ):
+                # If it is not an unskippable file, skip if needed
+                if os.path.basename(file_path) in SKIP_FORMATTING_FILES:
+                    continue
```
ShahafBenYakir, line 225: Do we have a unit test that check it fo all skippable and unskippable files today?

ShahafBenYakir (approved):
Looks great, onecomment

demisto_sdk/commands/format/format_module.py
DinaMeylakh: Unskippable: test_format_commonserver_skipped_files in format_integration_test.py
Skippable: test_format_integration_skipped_files in format_integration_test.py

PR #3464, opened 2023-08-14
improved demisto-sdk upgrade message

MosheEichler:
## Related Issues
fixes:

## Description
Improved the upgrade message instead of `To update, run pip3 install --upgrade demisto-sdk` => `To update, run pip3 install --upgrade demisto-sdk or poetry install (in case you work in a virtual environment)`

### DIFF
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -225,7 +225,8 @@ def main(ctx, config, version, release_notes, **kwargs):
             if last_release and __version__ != last_release:
                 logger.info(
                     f"[yellow]however version {last_release} is available.\n"
-                    f"To update, run pip3 install --upgrade demisto-sdk[/yellow]"
+                    f"To update, run pip3 install --upgrade demisto-sdk "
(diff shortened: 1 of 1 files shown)


demisto_sdk/__main__.py
```diff
@@ -225,7 +225,8 @@ def main(ctx, config, version, release_notes, **kwargs):
             if last_release and __version__ != last_release:
                 logger.info(
                     f"[yellow]however version {last_release} is available.\n"
-                    f"To update, run pip3 install --upgrade demisto-sdk[/yellow]"
+                    f"To update, run pip3 install --upgrade demisto-sdk "
+                    f"or poetry install (in case you work in a virtual environment)[/yellow]"
```
dorschw, line 229: 1. doesn't pip work in `venv`s? 🤔 
2. I think `poetry install` should be when working on `demisto/content`, or any repo that has `pyproject.toml`. Let's find a simple, correct phrasing for this case.

MosheEichler: 1. It works (poetry install will update all dependencies).
2. Correct.

So I'm closing this PR (since this is not required).

PR #3493, opened 2023-08-22
run neo4j queries in a sepreate thread

ilaner:
## Related Issues
fixes:

### DIFF
--- a/demisto_sdk/commands/content_graph/interface/neo4j/queries/common.py
+++ b/demisto_sdk/commands/content_graph/interface/neo4j/queries/common.py
@@ -1,4 +1,5 @@
 import traceback
+from concurrent.futures import ThreadPoolExecutor, TimeoutError
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict
@@ -8,6 +9,9 @@ from packaging.version import Version
 
 from demisto_sdk.commands.common.logger import logger
 from demisto_sdk.commands.content_graph.common import ContentType
+
+MAX_RETRIES_QUERY = 3
+QUERY_TIMEOUT = 60
 
 
 def labels_of(content_type: ContentType) -> str:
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/content_graph/interface/neo4j/queries/dependencies.py
```diff
@@ -85,11 +85,7 @@ def delete_deprecatedcontent_relationship(tx: Transaction) -> None:
 (:{ContentType.PACK}{{object_id: "{DEPRECATED_CONTENT_PACK}"}})
 DELETE r
 RETURN source.node_id AS source, target.node_id AS target"""
-    result = run_query(tx, query).data()
-    for row in result:
-        source = row["source"]
-        target = row["target"]
-        logger.debug(f"Deleted relationship USES between {source} and {target}")
+    run_query(tx, query).data()
```
dantavori, line 88:
```suggestion
    run_query(tx, query)
```

dantavori (approved):

demisto_sdk/commands/content_graph/interface/neo4j/queries/common.py
```diff
@@ -54,10 +58,26 @@ def to_neo4j_map(properties: dict) -> str:
...
+    result = None
     try:
         start_time: datetime = datetime.now()
         logger.debug(f"Running query:\n{query}")
-        result = tx.run(query, **kwargs)
+        # invoke a new thread and execute `tx.run in a thread. If the query times out, retry.
+        for retry in range(MAX_RETRIES_QUERY):
+            try:
+                with ThreadPoolExecutor(max_workers=1) as executor:
+                    future = executor.submit(tx.run, query, **kwargs)
```
kobymeir, line 69: please add a log line after the submit. preferably with some id of the transaction.

kobymeir (changes requested):

demisto_sdk/commands/content_graph/interface/neo4j/queries/common.py
ilaner: added a log line that the query was executed. The Transaction doesn't have an id and anyway I don't think it is very useful. In the log before we print the actual query that executed, so I think this log provides enough info for us

dansterenson (approved):

PR #3504, opened 2023-08-27
[validate, error-code] - fix bugs

GuyAfik:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-27487

## Description
see changelog / issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Removed the use of chunks when calculating content relationships.
 * Fixed an issue where the url regex in the **validate** command was wrong.
 * Fixed an issue where **pre-commit** command failed when using global environment.
-* Fixed an issue where the **validate** command would fail on external repos when trying to ignore `BA102`.
+* Fixed an issue where **validate** would fail in external repos when trying to ignore `BA102`.
 * Fixed an issue where the **error-code** command failed on some error codes.
 * Fixes an issue in **format** command where the `-i` option included files in `.venv` directories.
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 * Removed the use of chunks when calculating content relationships.
 * Fixed an issue where the url regex in the **validate** command was wrong.
 * Fixed an issue where **pre-commit** command failed when using global environment.
+* Fixed an issue where the **validate** command would fail on external repos when trying to ignore `BA102`.
```
dorschw, line 13:
```suggestion
* Fixed an issue where **validate** would fail in external repos when trying to ignore `BA102`.
```

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
...
+* Fixed an issue where the **error-code** command failed on some error codes.
```
dorschw, line 14:
```suggestion
* Fixed an issue where **error-code** failed on some error codes.
```

demisto_sdk/commands/error_code_info/error_code_info.py
```diff
@@ -36,7 +36,8 @@ def parse_function_parameters(sig: inspect.Signature):
             if get_origin(param.annotation) in [Union, Optional]:
                 param_type = get_args(param.annotation)[0]
 
-            if param_type is not str:
+            # param_type = str or param_type = "Optional[Type]" or param_type = "Union[Type1, Type2]"
+            if param_type is not str and not isinstance(param_type, str):
```
dorschw, line 40: can we use `not isinstance` instead of `is not str`?

dorschw (approved):

demisto_sdk/commands/error_code_info/error_code_info.py
GuyAfik: unfortunately not, because when the type of the argument in the error code's function is only `str` then `param_type` will evaluate to `str`, however in case where the type of the argument is `Optional[str]` for example it will evaluate to the string "Optional[str]", hence we need to handle both cases.

this is how the `annotation` method of param works unfortunately :(

PR #3505, opened 2023-08-27
SDK Parallel logger used in Content build is not parsing color tags

kobymeir:
## Related Issues
related: https://jira-hq.paloaltonetworks.local/browse/CIAC-8071
SDK Parallel logger used in Content build is not parsing color tags


demisto_sdk/commands/common/logger.py
```diff
@@ -63,6 +47,90 @@ def parse_int_or_default(value: Any, default: int) -> int:
...
+    ("darkgrey", 90),
+    ("darkred", 31),
+    ("lightblue", 94),
+    ("lightcyan", 96),
+    ("lightgreen", 92),
+    ("lightgrey", 37),
+    ("lightred", 91),
+]
+
+DEMISTO_LOG_LOOKUP = dict(  # Convert the list of tuples to a dict
```
dorschw, line 96: can we make it a dict comp, or simplify it in a different way?

dorschw (approved):

demisto_sdk/commands/common/logger.py
kobymeir: that's the simplest i could think of. i couldn't find another pattern to create 2 key per item in the list.

PR #3522, opened 2023-08-29
Fix bugs in validate -g

GuyAfik:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-27573
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-8185

## Description
1. Fixes an issue where the `validate -g` failed on`BA102` in external repos.
2. Fixes an issue where in case an external repo is not hosted in Github, the `.pack-ignore` could not be found when trying to collect it directly from Github. If the `.pack-ignore` cannot be found in Github, we would get its content from the default branch of the repository (e.g.: main/master) locally.
3. Fixes an issue where in case running `validate -g` on an empty `.pack-ignore` that was updated, the command fails.


demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -1992,22 +1992,28 @@ def get_all_files_edited_in_pack_ignore(self, modified_files: set) -> set:
             old_file_path, file_path = self.get_old_file_path(file_path)
             if not file_path.endswith(".pack-ignore"):
                 continue
-            old_file_content = get_remote_file(old_file_path, tag="master")
+            # if the repo is in Github, get the .pack-ignore content from the master branch in Github
+            # if the repo is not in Github / file cannot be found, try to take it from the latest commit on the initial branch (usually master/main)
```
dorschw, line 1996: replace `in Github` to `in remote`

PR #3568, opened 2023-09-03
validate that git is installed before executing any command.

GuyAfik:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-27651

## Description
1. validate that git is installed before executing any command and if not exit gracefully
2. make sure that for git operations we will only use the `GitUtil` class and not `Repo` class to manage git operations anymore.
3. add a ruff rule to prevent devs to import `git.repo`, but instead use `GitUtil` class 

image


demisto_sdk/__main__.py
```diff
@@ -194,12 +195,22 @@ def main(ctx, config, version, release_notes, **kwargs):
...
+
     logger = logging.getLogger("demisto-sdk")
     handle_deprecated_args(ctx.args)
 
     config.configuration = Configuration()
     import dotenv
 
+    # make sure that git is installed before using any command.
+    try:
+        subprocess.check_output(["git", "--version"])
```
dorschw, line 207: See GitUtil, we (re)raise `InvalidGitRepositoryError`, better handle it there.

GuyAfik: however, we have several places that we just the `Repo` class from github-python and not the `GitUtil` object
In order words we do not use only the `GitUtil` class for git stuff.

GuyAfik: basically it crashes when the first `git` imported is reached (which happens in `__main__.py`)

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 ## Unreleased
 * Fixed an issue where logging arguments were not in the standard kebab-case. The new arguments are: **console-log-threshold**, **file-log-threshold**, **log-file-path**.
+* Demisto-SDK will now exit gracefully with an appropriate error message when *git* is not installed.
```
dorschw, line 4: please document in `README.md`, that `Git` is required for using the SDK.

demisto_sdk/__main__.py
```diff
@@ -1,14 +1,20 @@
 # Site packages
+import sys
+
+import click
+
+try:
+    import git
+except ImportError:
+    sys.exit(click.style("Git executable cannot be found, or is invalid", fg="red"))
```
dorschw, line 9: use logger.error, then `sys.exit(1)`

demisto_sdk/commands/common/content/content.py
```diff
@@ -74,17 +72,13 @@ def git() -> Repo | None:
...
-                repo = Repo(content_path)
-                logger.debug(f"Using content path: {content_path}")
-            else:
-                repo = Repo(Path.cwd(), search_parent_directories=True)
-        except InvalidGitRepositoryError:
-            logger.debug("Git repo was not found.")
-            repo = None
+        if content_path := os.getenv("DEMISTO_SDK_CONTENT_PATH"):
+            git_util = GitUtil(Path(content_path))
+            logger.debug(f"Using content path: {content_path}")
```
dorschw, line 77:
```suggestion
```

let's move the log line into the `GitUtil` constructor, `f"GitUtil: using {content_path=}"`

demisto_sdk/commands/common/content/content.py
```diff
@@ -74,17 +72,13 @@ def git() -> Repo | None:
...
+        else:
+            git_util = GitUtil(search_parent_directories=True)
```
dorschw, line 79:
```suggestion
        return GitUtil(search_parent_directories=True)
```

demisto_sdk/commands/common/content/content.py
```diff
@@ -74,17 +72,13 @@ def git() -> Repo | None:
...
-            if content_path := os.getenv("DEMISTO_SDK_CONTENT_PATH"):
-                repo = Repo(content_path)
-                logger.debug(f"Using content path: {content_path}")
-            else:
-                repo = Repo(Path.cwd(), search_parent_directories=True)
-        except InvalidGitRepositoryError:
-            logger.debug("Git repo was not found.")
-            repo = None
+        if content_path := os.getenv("DEMISTO_SDK_CONTENT_PATH"):
+            git_util = GitUtil(Path(content_path))
```
dorschw, line 76:
```suggestion
            return GitUtil(Path(content_path))
```

demisto_sdk/commands/common/content/content.py
```diff
@@ -74,17 +72,13 @@ def git() -> Repo | None:
...
+            logger.debug(f"Using content path: {content_path}")
+        else:
+            git_util = GitUtil(search_parent_directories=True)
 
-        return repo
+        return git_util
```
dorschw, line 81:
```suggestion
```

demisto_sdk/commands/common/git_util.py
```diff
@@ -1,30 +1,34 @@
 import os
 import re
 from pathlib import Path
-from typing import Set, Tuple, Union
+from typing import Optional, Set, Tuple, Union
 
 import click
 import gitdb
-from git import InvalidGitRepositoryError, Repo
+from git import InvalidGitRepositoryError, Repo  # noqa: TID251
```
dorschw, line 8:
```suggestion
from git import InvalidGitRepositoryError, Repo  # noqa: TID251: required to create GitUtil
```

demisto_sdk/commands/common/git_util.py
```diff
@@ -1,30 +1,34 @@
...
+        search_parent_directories: bool = True,
+    ):
+        try:
+            self.repo = Repo(
+                path or Path.cwd(), search_parent_directories=search_parent_directories
+            )
+        except InvalidGitRepositoryError:
+            raise InvalidGitRepositoryError(
+                "Unable to find Repository from current working directory - aborting"
+            )
```
dorschw, line 31:
```suggestion
        except InvalidGitRepositoryError:
            logger.exception("Unable to find Repository from current working directory - aborting")
        raise
```

dorschw: also, why `from current working directory`? we can do `path = path or Path.cwd()` and then use `path.resolve()` in the error message`

demisto_sdk/commands/common/content/content.py
```diff
@@ -53,16 +54,13 @@ def from_cwd(cls) -> Content:
...
-
-        return content
+        try:
+            return Content(str(cls.git().repo.working_tree_dir))
+        except InvalidGitRepositoryError:
+            return Content(Path.cwd())
 
     @staticmethod
-    def git() -> Repo | None:
+    def git() -> GitUtil:
```
dorschw, line 63: `git()` looks odd. It would be clearer as a class attribute, I suggest taking the logic into a separate function outside the class, then in the class level doing `git_util = create_git_util()`

dorschw (approved):
Very nice

demisto_sdk/commands/common/content/content.py
GuyAfik: can't use logger in GitUtil due to circular import :(

demisto_sdk/commands/common/git_util.py
GuyAfik: agreed

demisto_sdk/commands/common/content/content.py
GuyAfik: even though i think you are right, ill change the name of the method to be `git_util`. 
I think that once we refactor the validate the code here will be completely irrelevant, i don't see the point in maintaining it much further

CHANGELOG.md
GuyAfik: on it

demisto_sdk/__main__.py
GuyAfik: we have a bit of a problem here because of circular imports 

tools.py imports GitUtil (because of current PR which uses the GitUtil class instead of repo)
common.logger.py imports tools
Gitutil imports common.logger (in this PR)

makes circular imports :/

PR #3576, opened 2023-09-04
fixing the log file rotation in build system

kobymeir:
## Description
fixing the log file rotation in build system

### DIFF
--- a/demisto_sdk/commands/lint/helpers.py
+++ b/demisto_sdk/commands/lint/helpers.py
@@ -310,10 +310,7 @@ def add_api_modules(
             copied_api_module_path = pack_path / f"{module_name}.py"
             if content_repo:  # if working in a repo
                 module_path = content_repo / api_module_path
-                try:
-                    os.symlink(src=module_path, dst=copied_api_module_path)
-                except NotImplementedError:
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/lint/helpers.py
```diff
@@ -310,7 +310,10 @@ def add_api_modules(
             copied_api_module_path = pack_path / f"{module_name}.py"
             if content_repo:  # if working in a repo
                 module_path = content_repo / api_module_path
-                shutil.copy(src=module_path, dst=copied_api_module_path)
+                try:
+                    os.symlink(src=module_path, dst=copied_api_module_path)
+                except NotImplementedError:
+                    shutil.copy(src=module_path, dst=copied_api_module_path)
```
kobymeir, line 316:
```suggestion
                shutil.copy(src=module_path, dst=copied_api_module_path)
```

ilaner (approved):

ilaner:
Can we change the default value for `DEMISTO_SDK_LOG_FILE_SIZE` to a larger value?
Currently, it is only 1MB, I think we can safely change to 10-100MB.

PR #3583, opened 2023-09-05
Adding support for OpenAI, Anthropic, Cohere, TogetherAI, Aleph Alpha, Huggingface Inference Endpoints, etc.

krrishdholakia:
Hi @GuyAfik @MichaelYochpaz 

Saw you support AI21. I'm working on litellm (simple library to standardize LLM API Calls - https://github.com/BerriAI/litellm) and was wondering if we could be helpful.

I added support for the providers listed above by replacing the HTTP API call with completion. The code is pretty similar to the OpenAI class - as litellm follows the same pattern as the openai-python sdk.

Would love to know if this helps.

Happy to add additional tests / update documentation, if the initial PR looks good to you.

### DIFF
--- a/demisto_sdk/commands/generate_outputs/generate_descriptions/generate_descriptions.py
+++ b/demisto_sdk/commands/generate_outputs/generate_descriptions/generate_descriptions.py
@@ -5,7 +5,7 @@ import os.path
 import threading
 import time
 from typing import Dict
-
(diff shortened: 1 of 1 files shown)


ishaan-jaff:
@GuyAfik @MichaelYochpaz any update on this PR ?

PR #3594, opened 2023-09-05
Updated validation to support force flag + some fixes

Shellyber:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-8262

## Description
1. Updated the release notes validation to support the new force flag template
2. Fixed an issue where the new force template showed the pack ID instead of the pack display name
3. Fixed an issue in regex that missed some wrong first-level header
4. When creating a new release notes file, it will be added to git automatically

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 * Fixed an issue where the **validate -g** failed reading a `.pack-ignore` file that contained only newlines and spaces.
 * Demisto-SDK will now exit gracefully with an appropriate error message when *git* is not installed.
 * Updated validation *RN116* to support the structure of **--force** flag in *update-release-notes* command.
+* Fixed an issue where the release notes file was not added automatically to git when using the *update-release-notes* command.
+* Fixed the structure in *update-release-notes* command when used with the **--force** flag. Now the header will display the pack display name.
 * Fixed the support in **validate** for `svg` images to have their theme suffix.
 * Fixed an issue where **modeling-rule test** command failed to properly compare types of fields.
 * Updated the **engineinfo** type in the script schema.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 ## Unreleased
 * Demisto-SDK will now exit gracefully with an appropriate error message when *git* is not installed.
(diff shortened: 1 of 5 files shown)


Shellyber:
I will add a changelog after the first review :)

ShahafBenYakir (commented):
Looks good. 
Let's add a unit test for the addition to git (it has been a regression couple of times) and also for the validation fix.

Shellyber:
> Looks good. Let's add a unit test for the addition to git (it has been a regression couple of times) and also for the validation fix.

Added the test for the git add
Already updated the test for the validation.

demisto_sdk/commands/update_release_notes/update_rn.py
```diff
@@ -429,6 +429,28 @@ def get_master_version(self) -> str:
...
+        try:
+            master_metadata = get_remote_file(self.metadata_path)
+        except Exception:
+            logger.exception(
+                f"[red]Failed fetching {self.metadata_path} from remote master branch."
+                "Using the local version (if exists), instead[/red]",
+            )
+        if master_metadata:
+            pack_display_name = master_metadata.get("name")
+        return pack_display_name
```
ShahafBenYakir, line 452: We should have the pack name inside the pack model (`pack.name`), no need to re-implement this especially not in the update.rn

Shellyber: Removed this function.
In this PR we still not using the Pack object, but we have already a function that gets the metadata.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Fixed an issue where **validate** falsely failed with error `DS108` on descriptions ending with brackets that contains a dot at the end of them.
 * Fixed an issue where the **validate -g** failed reading a `.pack-ignore` file that contained only newlines and spaces.
 * Demisto-SDK will now exit gracefully with an appropriate error message when *git* is not installed.
+* Updated validation *RN116* to support the structure of **--force** flag in *update-release-notes* command.
```
ShahafBenYakir, line 6: let's add that we fixed an issue that the rn was not added to git and that we fixed the structure of the rn when ran with -f

ShahafBenYakir (approved):
Great

PR #3604, opened 2023-09-07
Fixed an issue where **validate** falsly failed with error `DS108` on empty descriptions.

YuvHayun:
## Related Issues
fixes:

## Description
Fixed an issue where **validate** falsly failed with error `DS108` on empty descriptions.

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/content_entity_validator.py
+++ b/demisto_sdk/commands/common/hook_validations/content_entity_validator.py
@@ -793,15 +793,15 @@ class ContentEntityValidator(BaseValidator):
         line_with_missing_dot = ""
         for arg in dict_to_test.get(arg_field, []):
             stripped_description = strip_description(arg.get("description", ""))
-            if self.is_valid_description(stripped_description):
+            if self.is_invalid_description_sentence(stripped_description):
                 line_with_missing_dot += f"The argument {arg.get('name')} description should end with a period.\n"
         for output in dict_to_test.get("outputs") or []:
             stripped_description = strip_description(output.get("description", ""))
-            if self.is_valid_description(stripped_description):
+            if self.is_invalid_description_sentence(stripped_description):
                 line_with_missing_dot += f"The context path {output.get('contextPath')} description should end with a period.\n"
         return line_with_missing_dot
 
-    def is_valid_description(self, stripped_description: str) -> bool:
+    def is_invalid_description_sentence(self, stripped_description: str) -> bool:
         """
         Args:
             stripped_description: (str) a description or comment section from script / integration yml.
@@ -812,8 +812,8 @@ class ContentEntityValidator(BaseValidator):
             OR
                The description string ends with an URL.
         """
-        return (
-            stripped_description
-            and not stripped_description.endswith(".")
-            and not is_string_ends_with_url(stripped_description)
+        return not (
+            not stripped_description
+            or stripped_description.endswith(".")
+            or is_string_ends_with_url(stripped_description)
         )
--- a/demisto_sdk/commands/common/hook_validations/content_entity_validator.py
+++ b/demisto_sdk/commands/common/hook_validations/content_entity_validator.py
@@ -803,6 +803,8 @@ class ContentEntityValidator(BaseValidator):
 
     def is_valid_description(self, stripped_description):
         """
+        Args:
+            stripped_description: (str) a description or comment section from script / integration yml.
         Verifies the following::
             1. The description string exists (was extracted correctly from its field).
             2. The description string ends with a dot.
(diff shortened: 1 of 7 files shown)


demisto_sdk/commands/common/hook_validations/content_entity_validator.py
```diff
@@ -793,16 +793,27 @@ def is_line_ends_with_dot(self, dict_to_test: dict, arg_field: str):
...
         for output in dict_to_test.get("outputs") or []:
             stripped_description = strip_description(output.get("description", ""))
-            if not stripped_description.endswith(".") and not is_string_ends_with_url(
-                stripped_description
-            ):
+            if self.is_valid_description(stripped_description):
                 line_with_missing_dot += f"The context path {output.get('contextPath')} description should end with a period.\n"
         return line_with_missing_dot
+
+    def is_valid_description(self, stripped_description):
```
GuyAfik, line 804:
```suggestion
    def is_valid_description(self, stripped_description: str) -> bool:
```

demisto_sdk/commands/common/hook_validations/content_entity_validator.py
```diff
@@ -793,16 +793,29 @@ def is_line_ends_with_dot(self, dict_to_test: dict, arg_field: str):
...
+               The description string ends with an URL.
+            OR
+               The description string is empty
+        """
+        return (
+            stripped_description
+            and not stripped_description.endswith(".")
+            and not is_string_ends_with_url(stripped_description)
+            and not len(stripped_description) == 0
+        )
```
GuyAfik, line 821:
```suggestion
        return (
            not stripped_description.endswith(".")
            and not is_string_ends_with_url(stripped_description)
            and not len(stripped_description) == 0
        )
```

isn't this enough? Its already a string so if `stripped_description` evaluates to `True` it means its already not empty

demisto_sdk/commands/common/hook_validations/content_entity_validator.py
```diff
@@ -793,16 +793,29 @@ def is_line_ends_with_dot(self, dict_to_test: dict, arg_field: str):
...
+        """
+        Args:
+            stripped_description: (str) a description or comment section from script / integration yml.
+        Verifies the following::
+            1. The description string exists (was extracted correctly from its field).
+            2. The description string ends with a dot.
+            OR
+               The description string ends with an URL.
+            OR
+               The description string is empty
```
GuyAfik, line 814:
```suggestion
        Verifies the following:
            1. The description string is not a URL
            2. The description string does not end with a dot.
            3. The description string is not empty
            
        Args:
            stripped_description: (str) a description or comment section from script / integration yml.
       ```

**Review by GuyAfik — COMMENTED** [review]
@YuvHayun champ, see my comments

File: demisto_sdk/commands/common/hook_validations/content_entity_validator.py (hunk 3 above)
Comment (YuvHayun) [reply demisto_sdk/commands/common/hook_validations/content_entity_validator.py:814]: It checks for either one of these conditions to hold, so I think leaving the 'OR's is more informational @GuyAfik

File: demisto_sdk/commands/common/hook_validations/content_entity_validator.py (hunk 2 above)
Comment (YuvHayun) [reply demisto_sdk/commands/common/hook_validations/content_entity_validator.py:821]: good point

File: demisto_sdk/commands/common/hook_validations/content_entity_validator.py (hunk 3 above)
Comment (GuyAfik) [reply demisto_sdk/commands/common/hook_validations/content_entity_validator.py]: but you have AND between them and not OR, so what am i missing here?

Comment (GuyAfik) [reply demisto_sdk/commands/common/hook_validations/content_entity_validator.py]: in your docstring it says the opposite, verifies that the string ends with an URL or that the string is empty,
but you actually validate the opposite of what you write in the doscstring, so it does not make sense to me

Comment (YuvHayun) [reply demisto_sdk/commands/common/hook_validations/content_entity_validator.py]: I changed to OR but I'll rephrase it

File: demisto_sdk/commands/common/hook_validations/content_entity_validator.py [hunk 4]
```diff
@@ -793,16 +793,25 @@ def is_line_ends_with_dot(self, dict_to_test: dict, arg_field: str):
...
+        Return False if none of the following conditions are met:
+        - The description string doesn't exist/empty.
+        - The description string ends with a dot.
+        - The description string ends with an URL.
+        """
+        return not (
+            not stripped_description
+            or stripped_description.endswith(".")
+            or is_string_ends_with_url(stripped_description)
+        )
```
YuvHayun, line 817: @GuyAfik

GuyAfik: I think the previous implementation is better, more readable IMO

GuyAfik (approved):
see my comment, and then merge 👑

PR #3611, opened 2023-09-10
Fix DS108 in cases of descriptions ending with a dot inside

YuvHayun:
## Description
Fixed an issue where **validate** falsely failed with error `DS108` on descriptions ending with brackets with a dot inside.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,11 @@
 # Changelog
 ## Unreleased
-* Fixed an issue where **validate** falsely failed with error `DS108` on descriptions ending with brackets with a dot inside.
+* Fixed an issue where **validate** falsely failed with error `DS108` on descriptions ending with brackets that contains a dot at the end of them.
 * Demisto-SDK will now exit gracefully with an appropriate error message when *git* is not installed.
 
 ## 1.20.2
 * Updated the **pre-commit** command to run on all python versions in one run.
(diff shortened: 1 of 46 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -3934,3 +3934,18 @@ def parse_int_or_default(value: Any, default: int) -> int:
         return int(value)
     except (ValueError, TypeError):
         return default
+
+
+def is_sentence_ends_with_bracket(description):
```
GuyAfik, line 3939:
```suggestion
def is_sentence_ends_with_bracket(description: str):
```

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Changelog
 ## Unreleased
+* Fixed an issue where **validate** falsely failed with error `DS108` on descriptions ending with brackets with a dot inside.
```
GuyAfik, line 3:
```suggestion
* Fixed an issue where **validate** falsely failed with error `DS108` on descriptions ending with brackets that contains a dot at the end of them.
```

GuyAfik (approved):
@YuvHayun do we consider this case valid?

Shouldn't the dot be at the end anyway?

YuvHayun:
> @YuvHayun do we consider this case valid?
> 
> Shouldn't the dot be at the end anyway?

No, as per @ShirleyDenkberg it can be either inside or outside the brackets.

PR #3615, opened 2023-09-10
27596 UUID fix

maimorag:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-27596

## Description
Adding boolean flag auto-replace-uuids which defaults to true for BC, and for the customer's needs, they will set it to false.


demisto_sdk/__main__.py
```diff
@@ -1474,6 +1474,13 @@ def upload(ctx, **kwargs):
     is_flag=True,
     default=False,
 )
+@click.option(
+    "-aruuids",
+    "--auto-replace-uuids",
+    help="Whether to run demisto-sdk format on downloaded files or not",
+    is_flag=True,
+)
+
```
YuvHayun, line 1483: I think that we should add the opposite flag - 'no-uuid-replacement' (or something like that..) which setting to True will avoid UUID replacement since we don't want to affect the existing functionality (which is defaultly change the UUID).

Let's change that and change the other files accordingly.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Added a new validation (`PA137`) to **validate**, ensuring the absence of non-ig
 * Updated the **format** command to use the content graph instead of the id_set file.
 * Updated **format** command not to fail on unexpected values that returns from the graph, and just add it to the log.
 * Removed a redundant debug log on the `tools.get_file` function.
+* Adding a new boolean flag *auto-replace-uuids* to **download** command.
```
YuvHayun, line 20: After changing the flag name, make sure to add a description about the flag here:
Use this flag to avoid UUID replacements when downloading (which content items it affects) using **download** command.

demisto_sdk/commands/download/downloader.py
```diff
@@ -157,6 +157,7 @@ def __init__(
         item_type: str = "",
         init: bool = False,
         keep_empty_folders: bool = False,
+        auto_replace_uuids: bool = True,
```
YuvHayun, line 160: Make sure to add this argument to the doc strings.

YuvHayun (changes requested):
looks good, see my comments.
Also, did you manage to reproduce and test the fixes yourself?

demisto_sdk/__main__.py
dorschw: we can have both, but default=on. See other flags, we usually do `--do-thing/--no-do-thing`

CHANGELOG.md
dorschw:
```suggestion
* Added the *auto-replace-uuids* flag to the **download** command. Use it to ... <what Yuval suggested>
```

demisto_sdk/__main__.py
maimorag: I understand but according to dan in the ticket:
'we should have a boolean flag auto-replace-uuids which defaults to true for BC, and for the customer's needs, they will set it to false.'

maimorag: changing it to replace_uuids

maimorag:
Thanks @YuvHayun. 
Yes I tested the changes and verified it changes the UUID unless it is set to False

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Added a new validation (`PA137`) to **validate**, ensuring the absence of non-ig
 * Updated the **format** command to use the content graph instead of the id_set file.
 * Updated **format** command not to fail on unexpected values that returns from the graph, and just add it to the log.
 * Removed a redundant debug log on the `tools.get_file` function.
+* Added the *auto-replace-uuids* flag to the **download** command. Use it to ... <what Yuval suggested>
```
dorschw, line 20: `<...>` 🥲

maimorag: After applying the [suggested change](https://github.com/demisto/demisto-sdk/pull/3615#discussion_r1321108070).
I was fixed a commit after

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Added a new validation (`PA137`) to **validate**, ensuring the absence of non-ig
 * Updated the **format** command to use the content graph instead of the id_set file.
 * Updated **format** command not to fail on unexpected values that returns from the graph, and just add it to the log.
 * Removed a redundant debug log on the `tools.get_file` function.
+* Added the *auto-replace-uuids* flag to the **download** command. Use it to avoid UUID replacements when downloading using download command.
```
YuvHayun, line 20:
```suggestion
* Added the *auto-replace-uuids* flag to the **download** command. set this flag to False to avoid UUID replacements when downloading using download command.
```

YuvHayun (changes requested):
Also please add unit tests.

eyalpalo (changes requested):
Looks good, please add unit test, and the documentation @dorschw asked.

maxgubler:
If not `auto_replace_uuids`, isn't it also unnecessary to download_playbook_yaml? Skipping that would keep the UUIDs that are part of the content bundle (and speed up content downloads massively).

https://github.com/demisto/demisto-sdk/blob/983b4f1af5d32d1458a6d2789f8162ff1f68f0a8/demisto_sdk/commands/download/downloader.py#L353-L355

https://github.com/demisto/demisto-sdk/blob/983b4f1af5d32d1458a6d2789f8162ff1f68f0a8/demisto_sdk/commands/download/downloader.py#L283-L287

eyalpalo (approved):

maimorag:
> If not `auto_replace_uuids`, isn't it also unnecessary to download_playbook_yaml? Skipping that would keep the UUIDs that are part of the content bundle (and speed up content downloads massively).
> 
> https://github.com/demisto/demisto-sdk/blob/983b4f1af5d32d1458a6d2789f8162ff1f68f0a8/demisto_sdk/commands/download/downloader.py#L353-L355
> 
> https://github.com/demisto/demisto-sdk/blob/983b4f1af5d32d1458a6d2789f8162ff1f68f0a8/demisto_sdk/commands/download/downloader.py#L283-L287

Yes, this will be addressed later on.

maxgubler:
> Yes, this will be addressed later on.

Got it. Really that whole bit of api calling loop needs to be addressed for its inefficiency alone.

YuvHayun (approved):

CHANGELOG.md
```diff
@@ -138,6 +138,7 @@
 * Updated the **format** command to use the content graph instead of the id_set file.
 * Updated **format** command not to fail on unexpected values that returns from the graph, and just add it to the log.
 * Removed a redundant debug log on the `tools.get_file` function.
+* Added the *auto-replace-uuids* flag to the **download** command. set this flag to False to avoid UUID replacements when downloading using download command.
```
GuyAfik, line 141: please update changelog location

PR #3635, opened 2023-09-13
Unicode bug fixes

GuyAfik:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-28060


demisto_sdk/commands/common/handlers/json/ujson_handler.py
```diff
@@ -22,19 +22,19 @@ def __init__(self):
 
     def loads(self, s: AnyStr):
         try:
-            return ujson.loads(s)
+            return self.json.loads(s)
```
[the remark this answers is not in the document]
GuyAfik: bonus

demisto_sdk/commands/common/content/objects/pack_objects/pack_metadata/pack_metadata.py
```diff
@@ -522,11 +526,12 @@ def dump_metadata_file(self, dest_dir: Union[Path, str] = "") -> List[Path]:
             if self.preview_only:
                 file_content["previewOnly"] = True
 
-        new_metadata_path = os.path.join(dest_dir, "metadata.json")
-        with open(new_metadata_path, "w") as metadata_file:
-            json.dump(file_content, metadata_file, indent=4)
+        new_metadata_path = Path(str(dest_dir)).joinpath("metadata.json")
```
dorschw, line 529:
```suggestion
        new_metadata_path = Path(dest_dir, "metadata.json")
```

demisto_sdk/commands/common/tools.py
```diff
@@ -3237,12 +3236,11 @@ def get_mp_types_from_metadata_by_item(file_path):
...
     try:
-        with open(metadata_path) as metadata_file:
-            metadata = json.load(metadata_file)
-            marketplaces = metadata.get(MARKETPLACE_KEY_PACK_METADATA)
-            if not marketplaces:
-                return [MarketplaceVersions.XSOAR.value]
-            return marketplaces
+        metadata = get_file(metadata_path)
+        marketplaces = metadata.get(MARKETPLACE_KEY_PACK_METADATA)
+        if not marketplaces:
```
dorschw, line 3241:
```suggestion
        if not (marketplaces := get_file(metadata_path).get(MARKETPLACE_KEY_PACK_METADATA)):
```
🤡

demisto_sdk/commands/common/content/objects/pack_objects/pack_metadata/pack_metadata.py
GuyAfik: nice

demisto_sdk/commands/common/content/objects/abstract_objects/json_object.py
```diff
@@ -46,15 +47,14 @@ def _fix_path(path: Union[Path, str]) -> Path:
...
+            self._as_dict = get_file(self._path)
         except ValueError as e:
             raise exc.ContentSerializeError(self, self.path, str(e))
 
     def _serialize(self, dest_dir: Path):
         """Dump dictionary to json file"""
         dest_file = self._create_target_dump_dir(dest_dir) / self.normalize_file_name()
-        with open(dest_file, "w") as file:
-            json.dump(self.to_dict(), file)
+        safe_write_unicode(lambda f: json.dump(self.to_dict(), f), dest_file)
```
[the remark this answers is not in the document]
GuyAfik: on it, will do it for yml too

demisto_sdk/commands/common/tools.py
```diff
@@ -820,6 +822,7 @@ def get_file(
     clear_cache: bool = False,
     return_content: bool = False,
     keep_order: bool = False,
+    should_raise: bool = False,
```
dorschw, line 825:
```suggestion
    raise_on_error: bool = False,
```

demisto_sdk/commands/common/tools.py
```diff
@@ -2818,9 +2823,75 @@ def compare_context_path_in_yml_and_readme(yml_dict, readme_content):
     return different_contexts
 
 
-def write_yml(yml_path: str, yml_data: Dict):
-    with open(yml_path, "w") as f:
-        yaml.dump(yml_data, f)  # ruamel preservers multilines
+def safe_write_unicode_yml_or_json(
```
dorschw, line 2826:
```suggestion
def write_dict(
```

demisto_sdk/commands/common/tools.py
```diff
@@ -2818,9 +2823,75 @@ def compare_context_path_in_yml_and_readme(yml_dict, readme_content):
...
+    indent: int = 0,
+    sort_keys: bool = False,
+    **kwargs,
+):
+    """
+    Write unicode content into a json/yml file.
+    """
+    path = Path(path)
+
+    if isinstance(handler, YAML_Handler):
```
dorschw, line 2839: Can we keep handler default=`None`, and take the `path.suffix` to use the appropriate default handler, instead of forcing users to pass the right one? Can save us code and confusion.

demisto_sdk/commands/common/tools.py
```diff
@@ -2818,9 +2823,75 @@ def compare_context_path_in_yml_and_readme(yml_dict, readme_content):
...
+            path,
+            json_data=data,
+            handler=handler,
+            indent=indent,
+            sort_keys=sort_keys,
+            **kwargs,
+        )
+
+
+def safe_write_unicode_yml(
```
dorschw, line 2859: what's the difference between this one and the `_json` one? They seem to do pretty much the same thing.

demisto_sdk/commands/common/tools.py
GuyAfik: The reason I chose this way is because I am not sure that the suffix of the dumped file will always be `.json` or `.yml`

this is a better way to make sure we won't have unnecessary bugs IMO

demisto_sdk/commands/common/tools.py
GuyAfik: 1. one is for yml, while the other is for json
2. the default handler in each case is different, `yaml` for dumpring yml and `json` for dumping `json`

Even though they are very similar, I think we **should have** separate logic for each file type

demisto_sdk/commands/common/tools.py
GuyAfik: will use only `write_dict` for all

demisto_sdk/commands/common/tools.py
GuyAfik: will use only `write_dict` for all

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * Updated the **modeling-rules init & test** commands to support RULE section fields.
 * Stability improvements for **graph create** and **graph update** commands.
 * Fixed the *metadata* type in the XSIAM dashboard schema to *map*, with possible values: **lazy_load** and **cache_ttl**
+* Fixed an issue where using **prepare-content**, **upload**, **zip-packs** and **download** on non-unicode content caused errors.
```
dorschw, line 21:
```suggestion
* Fixed an issue where using **prepare-content**, **upload**, **zip-packs** and **download** on machines with default encoding other than unicode caused errors.
```

demisto_sdk/commands/common/tools.py
```diff
@@ -820,6 +822,7 @@ def get_file(
     clear_cache: bool = False,
     return_content: bool = False,
     keep_order: bool = False,
+    raise_on_error: bool = False,
```
dorschw, line 825: great, please document (docstring) the fact that we used to just return empty on error, and now allow catching errors

demisto_sdk/commands/common/tools.py
```diff
@@ -2818,9 +2823,23 @@ def compare_context_path_in_yml_and_readme(yml_dict, readme_content):
...
+    handler: Optional[XSOAR_Handler] = None,
+    indent: int = 0,
+    sort_keys: bool = False,
+    **kwargs,
+):
+    """
+    Write unicode content into a json/yml file.
+    """
+    path = Path(path)
+    file_handler = handler or (json if path.suffix.lower() == ".json" else yaml)
```
dorschw, line 2838: the `else` is risky (a `.md` for instance would be given the yaml loader), i'd replace it with two `if` and raise if it's not a supported suffix

demisto_sdk/commands/prepare_content/generic_module_unifier.py
```diff
@@ -79,8 +82,7 @@ def merge_generic_module_with_its_dashboards(self) -> Dict:
 
         Returns: the unified GenericModule
         """
-        with open(self.input_path) as f:
-            generic_module = json.load(f)
+        generic_module = get_file(self.input_path)
```
dorschw, line 85: we need `raise_on_error=True` in every place where an error would be risen (`open` would raise here), otherwise we may face odd bugs

dorschw (approved):
Very nice, see comments

demisto_sdk/commands/common/tools.py
```diff
@@ -2835,10 +2840,17 @@ def write_dict(
...
+        if suffix == ".json":
+            handler = json
+        elif suffix in {".yaml", ".yml"}:
+            handler = yaml
+        else:
+            raise ValueError(f"The file {path} is neither json/yml")
 
     safe_write_unicode(
-        lambda f: file_handler.dump(data, f, indent, sort_keys, **kwargs), path
+        lambda f: handler.dump(data, f, indent, sort_keys, **kwargs), path  # type: ignore[union-attr]
```
dorschw, line 2853: why the type-ignore?

dorschw (approved):

demisto_sdk/commands/common/tools.py
GuyAfik: looks like mypy doesn't recognize that handler will not be `None` in any situation, perhaps its going by the function signature?

PR #3648, opened 2023-09-19
add git repo fixture to TestSuite

GuyAfik:
## Description
since our code relies heavily on git operations (GitUtil), creating the content repo temporary folders with git initialized would be helpful in cases such as:

* Save us mocks for the `GitUtil` class.
* Would allow us to make use of git operations and to mimic real SDK behaviour which does a lot of git operation stuff.


TestSuite/repo.py
```diff
@@ -79,10 +80,23 @@ def __init__(self, tmpdir: Path):
...
     def __del__(self):
         shutil.rmtree(self.path, ignore_errors=True)
 
+    @classmethod
+    def init_as_git_repo(cls, tmpdir: Path) -> "Repo":
+        content_repo = cls(tmpdir, as_git_repo=True)
+        # commit all files
+        content_repo.git_util.commit_files("Initial Commit")
+        content_repo.git_util.repo.create_head("master").checkout()
+        return content_repo
```
dorschw, line 98: why both this and the `if` in `__init__`?

GuyAfik: updated the implementation

ilaner (approved):
good job

PR #3650, opened 2023-09-19
Updated the call rule regex to require a space

MosheEichler:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-8443

## Description
Updated the call rule regex patern from `r"call\s*(?P<rule_name>\w+)"` to `r"call\s+(?P<rule_name>\w+)"` in order to avoid something like this `callblabla`


bziser (commented):
please add a test which simulates the bug

kobymeir (approved):

bziser (approved):

PR #3662, opened 2023-09-22
[lint] Added support for sharing test_data folder for all automations in a pack

ennozdd:
## Related Issues
fixes: In cases where many automations are relying on the same set of test files, user has to copy and paste each test file into all the automations, this pr solves the issue by copying each test file into the automation's test_data folder


dorschw:
Hi @ennozdd, thanks for creating a pull request. 
After internal discussion, we decided to only allow `test_data` in integration/script folders.

PR #3683, opened 2023-09-28
introduce the file classes for file operations

GuyAfik:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-8895

## Description
This PR creates new services for handling files, this includes support for binary files, yml/json files, txt files and ini files.

1. Added the option to read a file from local file system.
2. Added the option to read a file from a specific commit/branch (remote or local)
3. Added the option to read a file from github api.
4. Added the option to read a file from gitlab api.
5. Added the option to read a file from any source via http request.
6. Added the option to read a file that is already loaded into memory (without path).
7. Implemented the `from_content_path` in `GitUtil` to retrieve the correct content path. following that we can deprecate `get_content_path` from `tools.py`

Reasons to switch using the File classes:
* one source of truth for files (only one implementation for each file type)
* handles any encoding of any files with read/write operations
* very simple to use, no need to create an instance, being done automatically
* makes the entire operations with reading files from git and api very easy (with remotes / without remotes)
* easy to fix any issues, need to change only single source of code
* possible read files as other types if needed
* handles automatically the path of reading/writing files instead of using get_content_path function
* possibly will allow us to support Windows.

TODO (update File Classes)
* add clear cache :heavy_check_mark:
* keep_oder additions to yml (this can be done by passing the correct yml_handler) :heavy_check_mark:
* implement unicode handling in the write method and make _write abstract :heavy_check_mark:
* add implementation to retrieve files from gitlab/github api and from general http request :heavy_check_mark:
* add implementation to read the file from its content (bytes)  :heavy_check_mark:


TestSuite/integration.py
```diff
@@ -59,6 +62,8 @@ def build(
         description: Optional[str] = None,
         changelog: Optional[str] = None,
         image: Optional[bytes] = None,
+        commands_txt: Optional[str] = None,
+        test: Optional[str] = None,
     ):
```
ilaner, line 67: why in this pr?

demisto_sdk/commands/common/files/file.py
```diff
@@ -0,0 +1,373 @@
...
+            )
+
+        input_path = git_util.repo.working_dir / input_path
+        if not input_path.exists():
+            raise FileNotFoundError(f"File {input_path} does not exist")
+
+        return input_path
+
+    @property
+    def input_file_content(self) -> bytes:
```
ilaner, line 67: Lets call it `content` and add a new method for `text`.
`content` will call `read_bytes` and `text` will call `read_text`

WDYT?

demisto_sdk/commands/common/files/file.py
```diff
@@ -0,0 +1,373 @@
...
+        if suffix := self.input_path.suffix.lower():
+            return suffix[1:]
+        return suffix
+
+    @property
+    def input_path_original_encoding(self) -> Optional[str]:
+        return UnicodeDammit(self.input_file_content).original_encoding
+
+    @property
+    def input_file_size(self) -> int:
```
ilaner, line 83: I think size is better name

demisto_sdk/commands/common/files/file.py
```diff
@@ -0,0 +1,373 @@
...
+            raise HttpFileReadError(url, exc=e)
+
+        try:
+            return cls.read_from_file_content(response.content, handler=handler)
+        except FileContentReadError as e:
+            logger.exception(f"Could not read file from {url} as {cls.__name__} file")
+            raise HttpFileReadError(url, exc=e)
+
+    @classmethod
+    def write_file(
```
ilaner, line 319: what is the difference between this and `write`?

ilaner (changes requested):
Good job.
Please see comments

demisto_sdk/commands/common/files/file.py
GuyAfik: but what if its not a text file? putting the text method in `File` is not relevant for **all** files.

TestSuite/integration.py
GuyAfik: because i needed it to create some examples of content items that i wanted to load and check, do you think it should be added in a different PR? its was mainly added for UTs.

demisto_sdk/commands/common/files/file.py
GuyAfik: `write_file` already creates the instance for you, so you can just call for example `TextFile.write_file(...)` and it would write whatever you need in one-line.

demisto_sdk/commands/common/files/yml_file.py
```diff
@@ -0,0 +1,13 @@
+from pydantic import Field, validator
+
+from demisto_sdk.commands.common.files.json_yml_file import JsonYmlFile
+from demisto_sdk.commands.common.handlers import DEFAULT_YAML_HANDLER as yaml
+from demisto_sdk.commands.common.handlers import YAML_Handler
+
+
+class YmlFile(JsonYmlFile):
```
JudahSchwartz, line 8: why isnt this in the files directory ?

demisto_sdk/commands/common/files/json_yml_file.py
```diff
@@ -0,0 +1,53 @@
...
+from pathlib import Path
+from typing import Any, Optional, Type, Union
+
+from pydantic import Field, validator
+
+from demisto_sdk.commands.common.files.text_file import TextFile
+from demisto_sdk.commands.common.handlers.xsoar_handler import XSOAR_Handler
+
+
+class JsonYmlFile(TextFile, ABC):
```
JudahSchwartz, line 13: thats not a great name. Maybe "StructuredDataFile" or something that better indicates what this is meant to be

JudahSchwartz (approved):
looks awesome. Add docstring

demisto_sdk/commands/common/files/yml_file.py
GuyAfik: nice catch, just by mistake, ill fix it

GuyAfik: we got confused, it is indeed in the files directory

ilaner (approved):
Good job!

demisto_sdk/commands/common/files/json_yml_file.py
GuyAfik: this is why i called it `HandlerFile` before because its a file that should be used by `xsoar_handler` 
In the graph for example we have classes called IntegrationScript

so why do you think J`sonYmlFile` isn't good enough, do you have maybe other suggestion?

PR #3295, opened 2023-07-13
[Download Command] Rewrite & Optimization

MichaelYochpaz:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/XSUP-25632, https://jira-dc.paloaltonetworks.com/browse/CIAC-6563

## Description
Optimize 'download' command's performance, and add logs.

PR #3698, opened 2023-10-02
Fix an issue where unified integrations / scripts with a period in th…

MichaelYochpaz:
## Related Issues
This PR fixes the issue mentioned [here](https://github.com/demisto/demisto-sdk/pull/3295/files#r1290410676), which seems to be mainly noticeable when downloading content.

## Description
This PR fixes an issue when splitting integrations with a period in their name (ex: "Test v1.0") would result in the python file being named "Test v1.py" due to `.0` being detected as a suffix by `pathlib.Path`.


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Fixed an issue where **validate** falsely failed with error `DS108` on descriptions ending with new lines followed by square/curly brackets.
 * Fixed an issue where **graph** commands would not clean their temporary files properly, causing successive commands to fail.
 * Fixed an issue where an error log message changed the terminal color.
+* Fixed an issue where in some cases, splitting unified integrations / scripts with a period in their name would result in an incorrect filename being used for the code file.
```
GuyAfik, line 12: which cases? can it be more specific and still be a short log?

move it to `Unreleased`

demisto_sdk/tests/test_files/integration-Zoom-v1.0.yml
```diff
@@ -0,0 +1,409 @@
+commonfields:
+  id: Zoom-v1.0
+  version: -1
```
GuyAfik, line 3: can we avoid adding new test files?
can you use `TestSuite` here instead maybe?

CHANGELOG.md
MichaelYochpaz: Updated it, let me know what if this is better.

demisto_sdk/tests/test_files/integration-Zoom-v1.0.yml
MichaelYochpaz: Fixed on c9d835df9f7357f35ef6dac69b6e88e9054fb9ae

PR #3709, opened 2023-10-05
Add `safe_write_unicode` function to tools

MichaelYochpaz:
## Description
This PR adds a few changes to the `tools` module made originally in the [download-rewrite PR](https://github.com/demisto/demisto-sdk/pull/3295).
This PR adds the `safe_write_unicode` function, which is needed for the new `download` rewrite, since we open files in memory and can't use the previous `_read_file` function (which opens files from disk).
It also adds support for using `pathlib.Path` on the `get_yml_paths_in_dir` and `get_child_directories` functions, and a few other minor fixes & changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 ## Unreleased
 * Added the *auto-replace-uuids* flag to the **download** command. set this flag to False to avoid UUID replacements when downloading using download command.
 * **format** command will run without the content graph if graph creation fails.
+* Replaced the `tools._read_file` function with a more generic `tools.safe_read_unicode` function.
+* Added `pathlib.Path` support to the `tools.get_yml_paths_in_dir` and `tools.get_child_directories` functions.
 
 ## 1.20.7
 * Fixed an issue where unified integrations / scripts with a period in their name would not split properly.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 ## Unreleased
 * Added the *auto-replace-uuids* flag to the **download** command. set this flag to False to avoid UUID replacements when downloading using download command.
 * **format** command will run without the content graph if graph creation fails.
-* Replaced the `tools._read_file` function with a more generic `tools.safe_read_unicode` function.
-* Added `pathlib.Path` support to the `tools.get_yml_paths_in_dir` and `tools.get_child_directories` functions.
+* Internal: Replaced the `tools._read_file` function with a more generic `tools.safe_read_unicode` function.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
 ## Unreleased
+* Replaced the `tools._read_file` function with a more generic `tools.safe_read_unicode` function.
+* Added `pathlib.Path` support to the `tools.get_yml_paths_in_dir` and `tools.get_child_directories` functions.
```
dorschw, line 4: please add the `Internal:` prefix

demisto_sdk/commands/common/tools.py
```diff
@@ -1,11 +1,13 @@
+from __future__ import annotations
```
dorschw, line 1: why?

demisto_sdk/commands/common/tools.py
```diff
@@ -758,36 +766,39 @@ def get_last_remote_release_version():
                 f'{exc_msg[exc_msg.find(">") + 3:-3]}.\n'
                 f"This may happen if you are not connected to the internet."
             )
-        logger.info(
-            f"[yellow]Could not get latest demisto-sdk version.\nEncountered error: {exc_msg}[/yellow]"
+        logger.warning(
+            f"Could not find the latest version of 'demisto-sdk'.\nError: {exc_msg}"
```
dorschw, line 770: why remove `yellow` as part of this PR?

dorschw (approved):

demisto_sdk/commands/common/tools.py
MichaelYochpaz: It allows using `|` for parameters typing like [here](https://github.com/demisto/demisto-sdk/pull/3709/files#diff-19219bb189e9e2c49968f53fb5cad06444a41c9502d62d4dade43e864a87cf59R693) on Python versions that are 3.7+ (without it, it's only allowed on 3.10+).

demisto_sdk/commands/common/tools.py
MichaelYochpaz: Using a `warning` log level already uses a yellow color automatically, I already checked and confirmed it.
These are just some remains from when we didn't do that automatically and then these tags were needed.

CHANGELOG.md
MichaelYochpaz:
```suggestion
Internal: * Replaced the `tools._read_file` function with a more generic `tools.safe_read_unicode` function.
Internal: * Added `pathlib.Path` support to the `tools.get_yml_paths_in_dir` and `tools.get_child_directories` functions.
```
Like this?

PR #3711, opened 2023-10-07
Fix download listing for custom "reputation" content items

MichaelYochpaz:
## Related Issues
relates: https://jira-hq.paloaltonetworks.local/browse/CIAC-3583

## Description
Fix an issue where custom `reputation` content items would be listed with an incorrect name (a UUID) instead of their actual display name when using the `-lf` flag for the `download` command.
This PR doesn't resolve the issue yet, as the fixed method here (`get_display_name`) is not currently in use by the `download` command, but it will be once the [download rewrite PR](https://github.com/demisto/demisto-sdk/pull/3295) will be merged.


demisto_sdk/commands/common/tools.py
```diff
@@ -3530,47 +3530,52 @@ def get_display_name(file_path, file_data={}) -> str:
...
-    elif "brandName" in file_data:
-        name = file_data.get("brandName", None)
-    elif "id" in file_data:
-        name = file_data.get("id", None)
-    elif "trigger_name" in file_data:
-        name = file_data.get("trigger_name")
-    elif "rule_name" in file_data:
-        name = file_data.get("rule_name")
+    _file_path = Path(file_path)
+    _file_data: dict = file_data or {}
```
MichaelYochpaz, line 3534: I've added a separate `_file_data` parameter since mypy complained for the section below that `file_data` can be `None` (due to the typing of the function - `file_data: dict | None = None`)

bziser (approved):
looks good, 
@GuyAfik please take a look too

demisto_sdk/commands/common/tools.py
```diff
@@ -3530,47 +3530,52 @@ def get_display_name(file_path, file_data={}) -> str:
...
-
+        return _file_data["layout"]["id"]
+    elif "name" in _file_data:
+        return _file_data["name"]
+    elif "TypeName" in _file_data:
+        return _file_data["TypeName"]
+    elif "brandName" in _file_data:
+        return _file_data["brandName"]
+    elif "reputationCommand" in _file_data and "details" in _file_data:
+        return _file_data["details"]
```
MichaelYochpaz, line 3554: GitHub kind of messed up the diffs, so these lines (which weren't there before) are the fix.
(The display name of reputation content items is within the `details` field. Tested it by creating a new item on the UI.)

demisto_sdk/commands/common/tools.py
GuyAfik: then just do `file_path = Path(file_path)` - no need for new variable if you define it to be either `str | Path`
also, you can do for `_file_data`
```
if not file_data:
    file_data = {}
```

also an option instead of creating new var

demisto_sdk/commands/init/contribution_converter.py
```diff
@@ -780,7 +780,7 @@ def format_user_input(self) -> Dict[str, str]:
             if line.startswith(entity_identifier):
                 entity_name = line.lstrip(entity_identifier)
                 if items_path.get(entity_name):
-                    entity_name = get_display_name(items_path.get(entity_name))
+                    entity_name = get_display_name(items_path[entity_name])
```
GuyAfik, line 783: there is always `entity_name` key in the `items_path` dict?

MichaelYochpaz: The condition is:
```
if items_path.get(entity_name):
    entity_name = get_display_name(items_path[entity_name])
```
So this line is executed only if `items_path.get(entity_name)` equals true.

GuyAfik (approved):
@MichaelYochpaz please add a UT

PR #3720, opened 2023-10-11
Add pre-loaded data support for `YmlSplitter`

MichaelYochpaz:
## Description
Added support for using preloaded data when using the `YmlSplitter`, to avoid having to re-parse the file for no reason.
Used by the WIP download command rewrite.

### DIFF
--- a/demisto_sdk/commands/split/ymlsplitter.py
+++ b/demisto_sdk/commands/split/ymlsplitter.py
@@ -76,7 +76,9 @@ class YmlSplitter:
         self.lines_inserted_at_code_start = 0
         self.config = configuration or Configuration()
         self.auto_create_dir = not no_auto_create_dir
-        self.yml_data = input_file_data if input_file_data is not None else get_yaml(self.input)
+        self.yml_data = (
+            input_file_data if input_file_data is not None else get_yaml(self.input)
+        )
         self.api_module_path: Optional[str] = None
 
     def get_output_path(self):
--- a/demisto_sdk/commands/split/ymlsplitter.py
+++ b/demisto_sdk/commands/split/ymlsplitter.py
@@ -41,7 +41,7 @@ class YmlSplitter:
     Attributes:
         input (str): input yml file path
         output (str): output path
-        loaded_data (dict | None): preloaded YAML data. if not provided, data will be loaded from input
+        input_file_data (dict | None): preloaded YAML data. if not provided, data will be loaded from input
         no_demisto_mock (bool): whether to add an import for demistomock
         no_common_server (bool): whether to add an import for common server
         no_auto_create_dir (bool): whether to create a dir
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/split/ymlsplitter.py
```diff
@@ -72,7 +76,7 @@ def __init__(
         self.lines_inserted_at_code_start = 0
         self.config = configuration or Configuration()
         self.auto_create_dir = not no_auto_create_dir
-        self.yml_data = get_yaml(self.input)
+        self.yml_data = loaded_data if loaded_data is not None else get_yaml(self.input)
```
bziser, line 79:
```suggestion
        self.yml_data = loaded_data if loaded_data else get_yaml(self.input)
```
why not like this?

MichaelYochpaz: Because I wanted to avoid re-loading the file in an edge case where `loaded_data` is an empty dict (`{}`).

MichaelYochpaz:
@GuyAfik I don't open a file to get those dicts, I extract & load them on memory from a tar file or receive a dict from API calls (for system content items).

demisto_sdk/commands/split/ymlsplitter.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
GuyAfik, line 1: what is this?

demisto_sdk/commands/split/ymlsplitter.py
```diff
@@ -54,6 +57,7 @@ def __init__(
         input: str,
         output: str = "",
         file_type: str = "",
+        loaded_data: dict | None = None,
```
GuyAfik, line 60: if you add here predefined loaded_data, why do we need the `input_file` to be required?

demisto_sdk/commands/split/ymlsplitter.py
```diff
@@ -39,6 +41,7 @@ class YmlSplitter:
     Attributes:
         input (str): input yml file path
         output (str): output path
+        loaded_data (dict | None): preloaded YAML data. if not provided, data will be loaded from input
```
GuyAfik, line 44:
```suggestion
        input_file_data (dict | None): preloaded YAML data. if not provided, data will be loaded from input
```

GuyAfik (commented):
add UT

demisto_sdk/commands/split/ymlsplitter.py
MichaelYochpaz: Allows using `|` for typing instead of`Union` in Python versions 3.7-3.9 (without it, it's only possible on versions 3.10+)

demisto_sdk/commands/split/ymlsplitter.py
MichaelYochpaz: I'm assuming you mean `input`? It seems to be used for determining the output path:
`self.output = (Path(output) if output else Path(self.input.parent)).resolve()`

I also want to add this parameter without breaking BC and potentially breaking stuff in other places.

demisto_sdk/commands/split/ymlsplitter.py
MichaelYochpaz: Renamed on 1f44a4149e3b9243825d7bf4ef1554a3525abf51

GuyAfik (approved):

PR #3724, opened 2023-10-15
Duplicate error codes

dorschw:
## Related Issues
fixes: https://jira-hq.paloaltonetworks.local/browse/CIAC-8645

## Description
Prevent errors from sharing the error code.


YuvHayun (changes requested):
Nice idea!
Please add docs for this unit test, and change one of the error codes (make sure we don't break anything related to content), and make sure to update the changelogs.

YuvHayun (approved):
Nice!

PR #3730, opened 2023-10-16
Added the update demisto-sdk automatically script

MosheEichler:
## Related Issues
related: https://jira-hq.paloaltonetworks.local/browse/CIAC-8464

## Description
Added a new scirpt to udpate the demsito-sdk versionץ
The script is useful when you want to update the demisto-sdk in your local environment (content & SDK repositories).

## How to use it
The script gets a list of local repositories to update the version.
```
update_demisto_sdk_version.sh  ~/dev/demisto/content ~/dev/demisto/demisto-sdk
```


demisto_sdk/scripts/update_demisto_sdk_version.sh
```diff
@@ -0,0 +1,38 @@
...
+# Save current path
+current_path="$PWD"
+
+# Loop through paths
+for repository in "${repositories[@]}"
+do
+  # Validate path
+  if [ ! -d "$repository" ]; then
+    echo "Error: $repository is not a valid directory."
+    exit 1
```
tkatzir, line 21: You fail on the first failed repo dir?
You may only update some of the required repos.
Maybe better to check all repos are ok, and then try to handle them one by one?

demisto_sdk/scripts/update_demisto_sdk_version.sh
```diff
@@ -0,0 +1,38 @@
...
+  fi
+
+  # CD to path
+  cd "$repository" || exit
```
tkatzir, line 25: That way the repos should be given as absolute paths.
You could use `pushd` (instead of `cd`), and then the repos could be either absolute or relative

demisto_sdk/scripts/update_demisto_sdk_version.sh
```diff
@@ -0,0 +1,38 @@
...
+  git stash
+  git checkout master
+  git pull
+  poetry install
+  git stash pop
+  git checkout "$current_content_branch"
+done
+
+# CD back to original path
+cd "$current_path" || exit
```
tkatzir, line 38: If you did `pushd` before, you could do `popd` here.

tkatzir (commented):
Very good!
Some comments.

demisto_sdk/scripts/update_demisto_sdk_version.sh
```diff
@@ -0,0 +1,59 @@
...
+  if [ ! -d "$repository" ]; then
+    invalid_repositories+=("$repository")
+  else
+    valid_repositories+=("$repository")
+  fi
+done
+
+# Make sure there are valid repositories
+if [ ${#valid_repositories[@]} -eq 0 ]; then
+  echo "There are no valid repositories found"
```
tkatzir, line 29:
```suggestion
  echo "Error, No valid repositories were found"
```

demisto_sdk/scripts/update_demisto_sdk_version.sh
```diff
@@ -0,0 +1,59 @@
...
+  git checkout "$current_content_branch"
+done
+
+# Move back to original path
+popd || exit
+
+# Check if invalid_repositories is not empty and print its items
+if [ ${#invalid_repositories[@]} -gt 0 ]; then
+  echo "The following invalid repositories are found:"
+  for invalid_repo in "${invalid_repositories[@]}"
```
tkatzir, line 55: Can you add the reason why they're invalid repositories?
Right now we only check if they're directories or not, but maybe in the future we would add more checks?

tkatzir (approved):
Very good!
Approved, with minor comments.

demisto_sdk/scripts/update_demisto_sdk_version.sh
MosheEichler: Screenshot 2023-11-27 at 9 37 37

MosheEichler: Screenshot 2023-11-27 at 9 38 37

PR #3761, opened 2023-10-31
updated infra to match the requests for the new validator

YuvHayun:
## Related Issues
fixes:

## Description
As a part of this PR:
- Added support for save() method to save changes made to the content_item in the original file.
- Added fields that will become handy for validations such as pack's ignored errors and support level.
- Added BaseContentModel class to differentiate between graph objects that are validatable and those that are not (such as commands).
- Added the functionality to create a BaseContent object from a specific git commit.

### DIFF
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -3984,7 +3984,7 @@ def get_all_repo_pack_ids() -> list:
     return [path.name for path in (Path(get_content_path()) / PACKS_DIR).iterdir()]
 
 
-def get(obj: dict, paths: Union[str, List[str]], defaultParam=None):
+def get_value(obj: dict, paths: Union[str, List[str]], defaultParam=None):
     """Extracts field value from nested object
     Args:
       obj (dict): The object to extract the field from
@@ -4025,12 +4025,12 @@ def find_correct_key(data: dict, keys: List[str]) -> str:
         str: Either the path where the given data object has a value at or the last option.
     """
     for key in keys:
-        if get(data, key, None):
+        if get_value(data, key, None):
             return key
     return keys[-1]
 
 
-def set_val(data: dict, paths: Union[str, List[str]], value) -> None:
+def set_value(data: dict, paths: Union[str, List[str]], value) -> None:
     """Updating a data object with given value in the given key.
     If a list of keys is given, will find the right path to update based on which path acctually has a value.
     Args:
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -846,7 +846,9 @@ def get_file(
     if git_sha:
         if file_path.is_absolute():
             file_path = file_path.relative_to(get_content_path())
-        return get_remote_file(str(file_path), tag=git_sha)
+        return get_remote_file(
+            str(file_path), tag=git_sha, return_content=return_content
+        )
 
     type_of_file = file_path.suffix.lower()
 
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -3974,3 +3974,63 @@ def parse_int_or_default(value: Any, default: int) -> int:
 
 def get_all_repo_pack_ids() -> list:
     return [path.name for path in (Path(get_content_path()) / PACKS_DIR).iterdir()]
+
+
+def get(obj: dict, fields: Union[str, List[str]], defaultParam=None):
+    """Extracts field value from nested object
+    Args:
+      obj (dict): The object to extract the field from
+      field (Union[str,List[str]]): The field or a list of possible fields to extract from the object, given in dot notation
+      defaultParam (object): The default value to return in case the field doesn't exist in obj
+    Returns:
+      str: The value of the extracted field
+    """
+    if isinstance(fields, str):
+        fields = [fields]
+    temp_obj = obj
+    for field in fields:
+        parts = field.split(".")
+        temp_obj = obj
+        success = True
+        for part in parts:
+            if temp_obj and part in temp_obj:
+                temp_obj = temp_obj[part]
+            else:
+                success = False
+                continue
+        if success:
+            return temp_obj
+    return defaultParam
+
+
+def find_correct_key(data: dict, keys: List[str]) -> str:
+    """Given a data object and a list of possible paths, finding the path where the object holds a value in that path.
+    Args:
+        data (dict): The object that holds the keys.
+        keys (List[str]) List of possible paths.
+    Returns:
+        str: Either the path where the given data object has a value at or the last option.
+    """
+    for key in keys:
+        if get(data, key, None):
+            return key
+    return keys[-1]
+
+
+def set_val(data: dict, keys: Union[str, List[str]], value) -> None:
+    """Updating a data object with given value in the given key.
+    If a list of keys is given, will find the right path to update based on which path acctually has a value.
+    Args:
+        data (dict): the data object to update.
+        keys (Union[str,List[str]]): the path or list of possible paths to update.
+        value (_type_): the value to update.
+    """
+    if isinstance(keys, list):
+        keys = find_correct_key(data, keys)
+    if get(data, keys, None):
+        keys = keys.split(".")
+        dic = data
+        for k in keys[:-1]:
+            dic = dic.get(k, {})
+
+        dic[keys[-1]] = value
--- a/demisto_sdk/commands/common/tools.py
+++ b/demisto_sdk/commands/common/tools.py
@@ -3976,7 +3976,7 @@ def get_all_repo_pack_ids() -> list:
     return [path.name for path in (Path(get_content_path()) / PACKS_DIR).iterdir()]
 
 
-def get(obj: dict, fields: Union[str, List[str]], defaultParam=None):
+def get(obj: dict, paths: Union[str, List[str]], defaultParam=None):
     """Extracts field value from nested object
     Args:
       obj (dict): The object to extract the field from
@@ -3985,17 +3985,22 @@ def get(obj: dict, fields: Union[str, List[str]], defaultParam=None):
     Returns:
       str: The value of the extracted field
     """
-    if isinstance(fields, str):
-        fields = [fields]
-    temp_obj = obj
-    for field in fields:
-        parts = field.split(".")
+    if isinstance(paths, str):
+        paths = [paths]
+    for path in paths:
+        keys = path.split(".")
         temp_obj = obj
         success = True
-        for part in parts:
-            if temp_obj and part in temp_obj:
-                temp_obj = temp_obj[part]
-            else:
+        for key in keys:
+            try:
+                if "[" in key and "]" in key:
+                    # Handle list indexing
+                    list_key, index = key.split("[")
+                    index = int(index.strip("]"))  # type: ignore
+                    temp_obj = temp_obj[list_key][index]
+                else:
+                    temp_obj = temp_obj[key]
+            except (AttributeError, KeyError, IndexError):
                 success = False
                 continue
         if success:
@@ -4017,7 +4022,7 @@ def find_correct_key(data: dict, keys: List[str]) -> str:
     return keys[-1]
 
 
-def set_val(data: dict, keys: Union[str, List[str]], value) -> None:
+def set_val(data: dict, paths: Union[str, List[str]], value) -> None:
     """Updating a data object with given value in the given key.
     If a list of keys is given, will find the right path to update based on which path acctually has a value.
     Args:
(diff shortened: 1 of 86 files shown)


demisto_sdk/commands/common/tools.py
```diff
@@ -834,24 +834,26 @@ def get_file(
...
-
     if raise_on_error = False, this function will return empty dict
     """
     if clear_cache:
         get_file.cache_clear()
     file_path = Path(file_path)  # type: ignore[arg-type]
+    if git_sha:
+        if file_path.is_absolute():
+            file_path = file_path.relative_to(get_content_path())
+        return get_remote_file(str(file_path), tag=git_sha)
```
ilaner, line 849: please provide `return_content` to the function

demisto_sdk/commands/common/tools.py
```diff
@@ -3974,3 +3982,84 @@ def parse_int_or_default(value: Any, default: int) -> int:
...
+                if "[" in key and "]" in key:
+                    # Handle list indexing
+                    list_key, index = key.split("[")
+                    index = int(index.strip("]"))  # type: ignore
+                    temp_obj = temp_obj[list_key][index]
+                else:
+                    temp_obj = temp_obj[key]
+            except (AttributeError, KeyError, IndexError):
+                success = False
+                continue
```
ilaner, line 4013: why should we continue and not return?

demisto_sdk/commands/common/tools.py
```diff
@@ -3974,3 +3982,84 @@ def parse_int_or_default(value: Any, default: int) -> int:
 
 def get_all_repo_pack_ids() -> list:
     return [path.name for path in (Path(get_content_path()) / PACKS_DIR).iterdir()]
+
+
+def get(obj: dict, paths: Union[str, List[str]], defaultParam=None):
```
ilaner, line 3987: let's rename to `get_value`

demisto_sdk/commands/content_graph/content_graph_builder.py
```diff
@@ -70,9 +70,8 @@ def _create_content_dtos(self, packs: Optional[List[str]]) -> List[ContentDTO]:
             leave=True,
         ) as progress_bar:
             for packs_batch in more_itertools.chunked(packs_to_parse, PACKS_PER_BATCH):
-                repository_parser.parse(packs_batch)
+                repository_parser.parse(packs_batch, progress_bar)
```
ilaner, line 73: revert

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -86,7 +87,7 @@ def __new__(
         return model_cls
 
 
-class BaseContent(ABC, BaseModel, metaclass=BaseContentMetaclass):
+class BaseContentModel(ABC, BaseModel, metaclass=BaseContentMetaclass):
```
ilaner, line 90: let's rename:
This base class will be `BaseContent`
The subclass will be `BaseContentWithPath`

@dantavori what do you think?
The idea is to create a class between `BaseContent` and the Pack/Content (without commands), so we can use the `path` attribute

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -130,6 +131,14 @@ def __getstate__(self):
             "__fields_set__": self.__fields_set__,
         }
 
+    @abstractmethod
```
ilaner, line 134: this function should be only on the subclass (BaseContentWithPath)

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -160,32 +169,84 @@ def to_dict(self) -> Dict[str, Any]:
...
+        if relationship.content_item_to == self:
+            # skip adding circular dependency
+            return
+        self.relationships_data[relationship_type].add(relationship)
+
+
+class BaseContent(BaseContentModel):
+    mapping: dict = Field({}, exclude=True)
+    path: Path
+    git_status: Optional[str]
```
ilaner, line 196: use enum for git status instead of str

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -160,32 +169,84 @@ def to_dict(self) -> Dict[str, Any]:
...
+    old_file_path: Optional[Path]
```
ilaner, line 197: just `old_path`

demisto_sdk/commands/content_graph/objects/content_item.py
```diff
@@ -183,6 +200,18 @@ def handler(self) -> XSOAR_Handler:
     def data(self) -> dict:
         return get_file(self.path, keep_order=False)
 
+    @property
+    def original_data(self) -> dict:
```
ilaner, line 204: can we think about a better name?

demisto_sdk/commands/content_graph/objects/content_item.py
```diff
@@ -183,6 +200,18 @@ def handler(self) -> XSOAR_Handler:
...
+        return get_file(self.path, keep_order=True)
+
+    def save(self):
```
ilaner, line 207: there is some code duplication between the `save` methods in Pack and in Content Item, which could lead to bugs in the future. There is even a bug here, since you miss handling the `marketplaces` field.

I suppose implementing a method in the `BaseContentWithPath` which receives the data and the path to write to, and the subclass calls it. For instance:
```
def save(self):
            self._save(self.path, self.original_data)
```
```
def save(self):
      self._save(self.metadata_path, self.original_data)
```
Where we add `metadata_path` and `original_data` to the `Pack` object.

demisto_sdk/commands/content_graph/objects/pack.py
```diff
@@ -497,3 +508,23 @@ def to_nodes(self) -> Nodes:
             self.to_dict(),
             *[content_item.to_dict() for content_item in self.content_items],
         )
+
+    def save(self):
```
ilaner, line 512: see the comment in the other `save` impl

demisto_sdk/commands/content_graph/parsers/pack.py
```diff
@@ -304,3 +316,33 @@ def deprecated(self) -> bool:
         ):
             return True
         return False
+
+    def parse_ignored_errors(self):
+        """Sets the pack's ignored_errors field."""
+        self.ignored_errors_dict = get_pack_ignore_content(self.path.name) or {}  # type: ignore
```
ilaner, line 322: why the type ignore here?

demisto_sdk/commands/content_graph/parsers/pack.py
```diff
@@ -238,11 +245,16 @@ def __init__(self, path: Path) -> None:
...
-            self.contributors: List[str] = get_json(path / PACK_CONTRIBUTORS_FILENAME)
+            self.contributors: List[str] = get_json(
+                path / PACK_CONTRIBUTORS_FILENAME, git_sha=git_sha
+            )
         except FileNotFoundError:
             logger.debug(f"No contributors file found in {path}")
         logger.debug(f"Parsing {self.node_id}")
         self.parse_pack_folders()
+        self.parse_ignored_errors()
+        logger.debug(f"Successfully parsed {self.node_id}")
```
ilaner, line 256: delete

demisto_sdk/commands/content_graph/parsers/parsing_rule.py
```diff
@@ -10,13 +12,21 @@
...
+
+    @cached_property
+    def mapping(self):
+        super().mapping.update({"object_id": "id"})
+        return super().mapping
 
     @property
     def object_id(self) -> Optional[str]:
-        return self.yml_data.get("id")
+        return get(self.yml_data, self.mapping.get("object_id", ""))
```
ilaner, line 29: we can remove all those properties in all the subclasses.
The impl is identical to YAMLContentItemParser.

demisto_sdk/commands/content_graph/parsers/repository.py
```diff
@@ -3,6 +3,8 @@
 from pathlib import Path
 from typing import Iterator, List, Optional
 
+from tqdm import tqdm
```
ilaner, line 6: revert the changes here (handled by another PR)

ilaner (changes requested):
Very good job.
Please see notes

demisto_sdk/commands/common/tools.py
YuvHayun: Because there're cases where the mapping hold several keys options for one item, so we want to be able to find the one where the value is really at.

demisto_sdk/commands/content_graph/parsers/pack.py
YuvHayun: you mean only the logger line right?

demisto_sdk/commands/content_graph/parsers/pack.py
```diff
@@ -217,16 +222,18 @@ class PackParser(BaseContentParser, PackMetadataParser):
...
 
-    def __init__(self, path: Path) -> None:
+    def __init__(self, path: Path, git_sha: Optional[str] = None) -> None:
         """Parses a pack and its content items.
 
         Args:
             path (Path): The pack path.
         """
+        if path.name == PACK_METADATA_FILENAME:
+            path = path.parent
```
dantavori, line 232: why is it needed?

demisto_sdk/commands/content_graph/parsers/xsiam_dashboard.py
```diff
@@ -16,28 +15,32 @@ class XSIAMDashboardParser(
...
-            or DEFAULT_CONTENT_ITEM_FROM_VERSION
+    @cached_property
+    def mapping(self):
+        super().mapping.update(
+            {
+                "object_id": [
+                    "dashboards_data[0].global_id",
+                    "dashboards_data[0].id",
+                    "id",
+                    "global_id",
```
dantavori, line 33: ?

demisto_sdk/commands/content_graph/tests/test_data/script.yml
dantavori, line 1: where is it used?

dantavori (commented):
good job, need to update the documentation with this changes, here (under content_graph/README.md and confluence)
please go over the fields/mapping again before you merge to ensure we didn't miss anything

demisto_sdk/commands/content_graph/tests/test_data/script.yml
YuvHayun: It's a part of the testings for the refactor tests, I can add it on the other PR if you prefer

demisto_sdk/commands/content_graph/parsers/xsiam_dashboard.py
YuvHayun: there are multiple options for where this key might be.. get_value will know how to find the right one.

demisto_sdk/commands/content_graph/parsers/pack.py
YuvHayun: We want to be able to parse a pack & pack_metadata from 'from_path' (-i inputs) so in some cases we receive pack_metadata as an input and this is how we handle this case.

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -160,32 +165,102 @@ def to_dict(self) -> Dict[str, Any]:
...
+        raise NotImplementedError()
+
     @staticmethod
     @lru_cache
-    def from_path(path: Path) -> Optional["BaseContent"]:
+    def from_path(
+        path: Path,
+        git_status: Optional[GitStatuses] = None,
+        old_file_path: Optional[Path] = None,
+        git_sha: Optional[str] = None,
```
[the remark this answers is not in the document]
YuvHayun: We won't call it altogether
But the idea is, that if we rename a file, then we create an object with the path, the git_status=Renamed, old_file_path = the file before the renaming.
Then when we want to create an object for the file before the renaming path = the old_file_path of the other file and git_sha will be the prev_ver (probably master)

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -160,32 +165,102 @@ def to_dict(self) -> Dict[str, Any]:
...
+    def add_relationship(
+        self, relationship_type: RelationshipType, relationship: "RelationshipData"
+    ) -> None:
+        if relationship.content_item_to == self:
+            # skip adding circular dependency
+            return
+        self.relationships_data[relationship_type].add(relationship)
+
+
+class BaseContentWithPath(BaseContent):
```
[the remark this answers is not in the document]
ilaner: BaseContent: the base class of all the nodes - Packs, Content Items, Commands
BaseContentWithPath: the base class of Packs and Content Items. Because they have `path` they share a lot of behavior

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -160,32 +165,109 @@ def to_dict(self) -> Dict[str, Any]:
...
-    def from_path(path: Path) -> Optional["BaseContent"]:
+    def from_path(
+        path: Path,
+        git_status: Optional[GitStatuses] = None,
+        old_file_path: Optional[Path] = None,
+        git_sha: Optional[str] = None,
+    ) -> Optional["BaseContentWithPath"]:
         logger.debug(f"Loading content item from path: {path}")
+        if git_status in (GitStatuses.MODIFIED, GitStatuses.RENAMED):
+            obj = BaseContentWithPath.from_path(path, git_status)
```
ilaner, line 244: please comment what we are doing here, as it is not that obvious

demisto_sdk/commands/content_graph/objects/base_content.py
```diff
@@ -206,39 +288,15 @@ def from_path(path: Path) -> Optional["BaseContent"]:
             logger.error(f"Could not parse content item from path: {path}")
             return None
         try:
-            return model.from_orm(content_item_parser)
+            obj = model.from_orm(content_item_parser)
+            obj.git_status = git_status
```
ilaner, line 292: let's go back to `return model.from_orm(content_item_parser)`, and set the git_status in the first `if` condition

ilaner (approved):
Great!

demisto_sdk/commands/content_graph/parsers/xsiam_dashboard.py
dantavori: can you please share a path to an existing XSIAM dashboard that has "id" or "global_id" fields in the root level of the json?

YuvHayun: Oh my bad, looking into this again, I got confused because in the super class we'd extract dashboards_data and then call id / global_id.. removing.

PR #3744, opened 2023-10-26
Collect the python file along with the yml file in demisto-sdk pre-commit

RotemAmit:
## Related Issues
fixes:https://jira-dc.paloaltonetworks.com/browse/CIAC-8303

## Description: When a yml file is given to demisto-sdk pre-commit command, the pre-commit will also check the matching python file.
For example, when running `demisto-sdk pre-commit -i Packs/Discord/Integrations/Discord/Discord.yml`, pre-commit will run on the files Packs/Discord/Integrations/Discord/Discord.py and Packs/Discord/Integrations/Discord/Discord.yml.

PR #3764, opened 2023-11-02
In preprocess_files function check that the py file exists.

RotemAmit:
## Related Issues
fixes:

## Description: 
When running the pre-commit command with only a yml file as input, we want to collect the python file as well. Before we do that we want to make sure that it exists.

### DIFF
--- a/demisto_sdk/commands/pre_commit/pre_commit_command.py
+++ b/demisto_sdk/commands/pre_commit/pre_commit_command.py
@@ -436,8 +436,9 @@ def preprocess_files(
             files_to_run.add(file)
             # if the current file is a yml file, add the matching python file to files_to_run
             if str(file).endswith("yml"):
-                str_py_file_path = str(file).replace("yml", "py")
-                files_to_run.add(Path(str_py_file_path))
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -441,10 +441,11 @@ def preprocess_files(
             files_to_run.update({path for path in file.rglob("*") if path.is_file()})
         else:
             files_to_run.add(file)
-            # if the current file is a yml file, add the matching python file to files_to_run
+            # If the current file is a yml file, add the matching python file to files_to_run
             if str(file).endswith("yml"):
```
ilaner, line 445:
```suggestion
            if file.suffix == ".yml"
```

GuyAfik:
@RotemAmit no need for changelog here?

eyalpalo:
@GuyAfik This is a fix to an already document change Rotem did, it is in the same release, so I think we don't need a new entry in the changelog

RotemAmit:
@GuyAfik This is the related [PR](https://github.com/demisto/demisto-sdk/pull/3744).

PR #3766, opened 2023-11-02
added sections to schema

dantavori:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-8295

## Description
added support for inputsSections and outputsSections in the playbook yml.


demisto_sdk/commands/common/schemas/playbook.yml
```diff
@@ -311,6 +319,23 @@ schema;input_schema:
...
 
+# playbook->inputSections
+schema;inputsSection_schema:
+  type: map
+  mapping:
+    name:
+      type: str
+      required: true
+    id:
+      type: str
```
ShahafBenYakir, line 330: This is optional I understand? 
I think we need to make sure the ID will not be exported when exporting the playbook yml

demisto_sdk/commands/common/tests/playbook_test.py
```diff
@@ -580,6 +580,12 @@ class TestPlaybookValidator:
         (INVALID_PLAYBOOK_INPUTS_USE, False, True),
     ]
 
+    def test_valid_schema(self, playbook):
```
ShahafBenYakir, line 583: Not sure I understand what this checks that was not checked so far?

ShahafBenYakir (approved):
Before merging, let's open a PR in content (having this SDK branch in the poetry) with this changes to some playbook and see that build passes

demisto_sdk/commands/common/tests/playbook_test.py
dantavori: Yes there's a bit of a mess... I searched for `is_valid_scheme` in the repo and here's what we have:
* In `structure_test.py`:
   * For playbooks there's only `test_scheme_validation_playbook` which does not use TestSuite
   * There are 9 schema tests (with TestSuite) for incident/indicator fields and reputation
   * There are schema tests (with TestSuite) for all **XSIAM** content types
* In different test files, there are 3 tests for generic fields, and 1 test for layout rule

Except for these there are no tests for schemas at all.

demisto_sdk/commands/common/schemas/playbook.yml
dantavori: It shouldn't be exported from XSOAR when we download it, I think we can remove this, but I'll ensure.

demisto_sdk/commands/common/tests/playbook_test.py
ShahafBenYakir: Meaning we were not validating schema for playbooks?

dantavori: we validate in `test_scheme_validation_playbook`

PR #3776, opened 2023-11-09
Add a flag to enable the option to run pre-commit without docker hooks

RotemAmit:
## Related Issues
fixes:https://jira-dc.paloaltonetworks.com/browse/CIAC-8830

## Description
Add the flag `--docker/--no-docker` in order to enable the option to run `demisto-sdk pre-commit` without docker hooks.


JudahSchwartz:
we dont have a way to enable/disable specific hooks. is that not behavior we want? For example disable pylint-in-docker only but yes to run powershell tests

@eyalpalo @kobymeir

demisto_sdk/__main__.py
```diff
@@ -3463,6 +3463,12 @@ def update_content_graph(
     nargs=-1,
     type=click.Path(exists=True, resolve_path=True, path_type=Path),
 )
+@click.option(
+    "--docker/--no-docker",
+    help="Whether to run docker based hooks or not.",
+    default=False,
```
eyalpalo, line 3469: Lets change the default to True as @JudahSchwartz suggested.

demisto_sdk/commands/pre_commit/hooks/docker.py
```diff
@@ -194,15 +194,18 @@ class DockerHook(Hook):
     This class will make common manipulations on commands that need to run in docker
     """
 
-    def prepare_hook(self, files_to_run: Iterable):
+    def prepare_hook(self, files_to_run: Iterable, run_docker_hooks: bool = False):
```
eyalpalo, line 197: If we have default value in the command option, do we need this default?

eyalpalo (changes requested):
Nice job!

eyalpalo (approved):

PR #3784, opened 2023-11-13
Collect pwsh files as part of the demisto-sdk pre-commit

RotemAmit:
## Related Issues
fixes:https://jira-dc.paloaltonetworks.com/browse/CIAC-7993

## Description
* In order to add the command pwsh-test to pre-commmit, it was needed to add them to the rest of the files instead of excluding them.
* A bug fix - when adding js files to language_to_files dictionary, their python_version was initialized with 2.7 instead of being None, since they are not python files. The fix is in the function python_version() in the file demisto_sdk/commands/content_graph/objects/integration_script.py.


demisto_sdk/commands/common/content_constant_paths.py
```diff
@@ -24,6 +24,7 @@
 PYTHONPATH = [
```
JudahSchwartz, line 24: dont think CommonServerPowerShell should be in python path

PR #3795, opened 2023-11-15
Fix tag parser missing flag

omerKarkKatz:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-8917

## Description
Fixed an issue that the tag parser regex was missing a flag allowing new lines when using the .* char.


adi88d (approved):
NICE!

PR #3797, opened 2023-11-16
clients: add uts + fix auth id bug

GuyAfik:
## Description
* Fixed an issue where *auth_id* was not tested properly when initializing `ClientConfig`
* Added UTs for initializing `ClientConfigs` to avoid breaking the functionality there

### DIFF
--- a/demisto_sdk/commands/common/clients/configs.py
+++ b/demisto_sdk/commands/common/clients/configs.py
@@ -70,10 +70,10 @@ class XsoarClientConfig(BaseModel):
 class XsoarSaasClientConfig(XsoarClientConfig):
     auth_id: str = Field(default=os.getenv(AUTH_ID), description="XSOAR/XSIAM Auth ID")
 
-    @validator("api_key", always=True)
+    @validator("auth_id", always=True)
     def validate_auth_params(cls, v, values):
         if not v:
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/common/clients/configs.py
```diff
@@ -70,10 +70,10 @@ def __eq__(self, other):
 class XsoarSaasClientConfig(XsoarClientConfig):
     auth_id: str = Field(default=os.getenv(AUTH_ID), description="XSOAR/XSIAM Auth ID")
 
-    @validator("api_key", always=True)
+    @validator("auth_id", always=True)
     def validate_auth_params(cls, v, values):
         if not v:
-            raise ValueError("api_key is required for SaaS")
+            raise ValueError("auth_id is required for xsoar-saas/xsiam")
```
ilaner, line 76: Why?
`auth_id` is already required without a custom validator (line 71)
However, the `api_key` is defined as optional in `XsoarClientConfig`, so we overwrite the validator to make it required.

GuyAfik: kept `api_key` validation and added `auth_id` validation

ilaner (approved):

PR #3800, opened 2023-11-16
YR/Add-a-graph-flag-to-the-upload-command/XSUP-30181

RosenbergYehuda:
## Related Issues
fixes:


GuyAfik:
@RosenbergYehuda isn't there an option to calculate dependancies for a specific pack rather then call the graph and add a new argument? That would be the preferable solution and it would make the performance better
cc: @ilaner

PR #3804, opened 2023-11-19
[clients] - validate that base-api-url is indeed the api-url

GuyAfik:
## Description
Validates that the `DEMISTO_BASE_URL` is indeed a valid api url and its not by mistake a reference for the base URL.


ilaner (approved):
nice

PR #3809, opened 2023-11-20
[clients] - minor refactors

GuyAfik:
## Description
* add additional logs to client classes
* set env vars to be the default values for __init__ funcs
* add poll-incident-state method
* make verify ssl as part of the configs
* add ip and fqdn properties


ilaner (approved):
Good job

PR #3812, opened 2023-11-20
Add unit+integration+graph tests Github action

GuyAfik:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-7032

## Description
See issue's comment


JudahSchwartz (approved):

.github/actions/setup_environment/action.yml
```diff
@@ -0,0 +1,42 @@
...
+    - name: Set up Python ${{ inputs.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ inputs.python-version }}
+
+    - name: Install SDK dependencies with poetry
+      run: |
+        sudo curl -sSL https://install.python-poetry.org | python3 -
+        poetry --version
+        poetry check --lock
```
ilaner, line 29: I think it's worth moving `poetry check` to a different step, so it the poetry.lock and pyproject.toml are inconsistent it will fail on this step.
In addition, we can use install poetry action instead. I believe they take care on cache and so on

ilaner (approved):
looks really good!
see minor comment about `poetry install`

.github/actions/setup_environment/action.yml
GuyAfik: done

PR #3813, opened 2023-11-21
Add validations batch

YuvHayun:
## Related Issues
fixes:


demisto_sdk/commands/validate/validators/BA_validators/BA116_cli_name_should_equal_id.py
```diff
@@ -0,0 +1,47 @@
...
+                message=self.error_message.format(
+                    content_item.cli_name, content_item.object_id
+                ),
+                content_object=content_item,
+            )
+            for content_item in content_items
+            if content_item.object_id != content_item.cli_name
+        ]
+
+    def fix(self, content_item: ContentTypes) -> FixResult:
```
JudahSchwartz, line 41: why does is_valid get a list, but fix gets a single item? Is that something we can fix in the infra? Seems very counter intuitive

demisto_sdk/commands/validate/tests/BA_validators_test.py
```diff
@@ -0,0 +1,107 @@
...
+@pytest.mark.parametrize(
+    "content_item, expected_name",
+    [
+        (
+            create_wizard_object({"id": "should_fix"}),
+            "should_fix",
+        ),
+    ],
+)
+def test_IDNameValidator_fix(content_item, expected_name):
```
JudahSchwartz, line 89: theres only 1 item in this list, why parameterize? Cases comment doesnt seem relevant

demisto_sdk/commands/validate/tests/DO_validators_test.py
```diff
@@ -0,0 +1,65 @@
...
+from demisto_sdk.commands.validate.tests.test_tools import (
+    create_integration_object,
+    create_script_object,
+)
+from demisto_sdk.commands.validate.validators.DO_validators.DO108_docker_image_exist import (
+    DockerImageExistValidator,
+)
+
+
+@pytest.mark.parametrize(
```
JudahSchwartz, line 12: to all of these test files we should add a few tests to make sure the message is as expected

demisto_sdk/scripts/init_validation_script.py
```diff
@@ -452,8 +452,8 @@ def generate_supported_content_types_section(self):
                 for content_type in self.content_types
             ]
             self.supported_content_types = (
-                f"ContentTypes = Union{str(supported_content_types)}"
-            )
+                f"ContentTypes = Union{(supported_content_types)}"
+            ).replace("'", "")
```
JudahSchwartz, line 456: whats this change?

JudahSchwartz (approved):
@dantavori please check the changes made to graph work

@YuvHayun looks great!

JudahSchwartz:
@YuvHayun please ping me to look over changelog

YuvHayun:
> @YuvHayun please ping me to look over changelog

I didn't document it on purpose.. since the refactor is not officially out yet.

demisto_sdk/commands/validate/validators/BA_validators/BA116_cli_name_should_equal_id.py
YuvHayun: is_valid gets a list because some validations can be done on the whole list (for example graph validations need 1 execution with the received list)
however, each item needs a fix on its own..
We could change that so that fix gets a list as well but that actually makes it harder in my opinion as it can't be done in list comp so users will need to:
1. create a results list,
2. iterate on all items
3. for each item fix and append the fix result to the list
4. return the list
Rather than fix, and return the fix result.

demisto_sdk/scripts/init_validation_script.py
YuvHayun: Putting str on the whole list would print the content types with "'" around them, and we need the actual object rather than the string of its class name.

demisto_sdk/commands/validate/tests/BA_validators_test.py
YuvHayun: Forgot to change after copy-paste, changing now.

demisto_sdk/commands/validate/validators/IN_validators/IN130_is_integration_runable.py
```diff
@@ -0,0 +1,37 @@
...
+    ValidationResult,
+)
+
+ContentTypes = Integration
+
+
+class IsIntegrationRunnableValidator(BaseValidator[ContentTypes]):
+    error_code = "IN130"
+    description = "validate that the integration is runable"
+    error_message = "Could not find any runnable command in the integration.\nMust have at least one command, `isFetch: true`, `feed: true`, `longRunning: true`."
```
dantavori, line 17:
```suggestion
    error_message = "Could not find any runnable command in the integration.\nMust have at least one of: a command under the `commands` section, `isFetch: true`, `feed: true`, or `longRunning: true`."
```

demisto_sdk/commands/validate/validators/DO_validators/DO108_docker_image_exist.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class DockerImageExistValidator(BaseValidator[ContentTypes]):
+    error_code = "DO108"
+    description = "Validate that the given content item has a docker_image."
+    error_message = (
+        "The {0} {1} is missing a docker image, please make sure to add one."
+    )
+    related_field = "Docker image"
+    is_auto_fixable = False
```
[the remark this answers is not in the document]
YuvHayun: I think that's a great idea.
It's worth changing to that and informing the users.

PR #3823, opened 2023-11-27
[pre-commit] do not delete properties which set to False

ilaner:
Currently, we delete properties in the configuration which are set to False.
For instance, in the template we set sometimes the `pass_filenames` to False. Right now, we delete this flag which makes the default is True.


JudahSchwartz (approved):
nice catch

PR #3824, opened 2023-11-27
[pre-commit] support skip for modes

ilaner:
This supports skip property for modes.
If provided `skip:<mode>`, the selected hook will not be genreated at all.
## Related Issues
related: https://jira-dc.paloaltonetworks.com/browse/CIAC-9037?filter=-2


demisto_sdk/commands/pre_commit/hooks/hook.py
```diff
@@ -80,7 +80,8 @@ def filter_files_matching_hook_config(
             and (not exclude_pattern or not re.search(exclude_pattern, str(file)))
         }  # only exclude if defined
 
-    def _get_property(self, name, default=None):
+    @staticmethod
+    def get_property(hook: dict, mode: str, name: str, default=None):
         """
         Will get the given property from the base hook, taking mode into account
         Args:
```
mmhw, line 87: Please add docstrings for "hook"

demisto_sdk/commands/pre_commit/hooks/hook.py
```diff
@@ -80,7 +80,8 @@ def filter_files_matching_hook_config(
             and (not exclude_pattern or not re.search(exclude_pattern, str(file)))
         }  # only exclude if defined
 
-    def _get_property(self, name, default=None):
+    @staticmethod
+    def get_property(hook: dict, mode: str, name: str, default=None):
```
mmhw, line 84: Still need the default `arg`?

demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -129,15 +129,15 @@ def _get_repos(pre_commit_config: dict) -> dict:
...
+        for i, repo in enumerate(pre_commit_config["repos"]):
+            for j, hook in enumerate(repo["hooks"]):
+                if Hook.get_property(hook, self.mode, "skip"):
+                    del pre_commit_config["repos"][i]["hooks"][j]
+                    continue
                 hooks[hook["id"]] = {
                     "repo": repo,
                     "hook": hook,
                 }
-                # if the hook has a skip key, we add it to the SKIPPED_HOOKS set
```
mmhw, line 138: What about a case of `skip: true` without mode, will it still work?

mmhw (commented):
Good job!
Pay attention to my comments.

Don't forget to add `CHANGELOG`.

demisto_sdk/commands/pre_commit/hooks/hook.py
ilaner: It makes sense to have a default value if the property is not found. Sometimes we would like to set it to `True`, for instance when we will want to filter out hooks with the `community: false` property

PR #3828, opened 2023-11-28
Replace ec2-user with gcp-user (demisto-sdk)

samuelFain:
## Related Issues
relates: [CIAC-8955](https://jira-dc.paloaltonetworks.com/browse/CIAC-8955)

## Description
This PR introduces the following changes:
- Update file paths in demisto-sdk which used `ec2-user` to `gcp-user`.
- Grant correct permissions to commit and push changes in the mock files repo (more information in Jira ticket).


demisto_sdk/commands/test_content/mock_server.py
```diff
@@ -251,6 +251,9 @@ def push_mock_files(self):
                 f"Failed pushing mock files with error: {exc}", real_time=True
             )
         finally:
+            self.logging_module.debug(
+                "Successfully pushed new/updated mock files to mock git repo.", real_time=True
```
JasBeilin, line 255: Do we actually know it was successful if it is in the finally section?

JasBeilin (commented):
Looks good, just one comment.

demisto_sdk/commands/test_content/mock_server.py
samuelFain: No, we don't. While there are logs for when an Exception is caught, we can't reliably confirm the success of the push, which might be misleading. Removing this log.

JasBeilin (approved):
Nice!

PR #3832, opened 2023-11-29
Support custom value for marketplace in integrations and script

ilaner:
This supports setting values for keys for specific marketplace.
For instance, we will be able to set `defaultValue` to `true`, but override this value in XPANSE with `defaultValue:xpanse: false`.

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-8959


demisto_sdk/commands/common/schemas/classifier.yml
```diff
@@ -18,44 +18,85 @@ mapping:
...
+    type: str
+  regex;(modified:(xsoar)|(marketplacev2)|(xpanse)|(xsoar_on_prem)|(xsoar_saas)):
+    type: str
+  regex;(defaultIncidentType:(xsoar)|(marketplacev2)|(xpanse)|(xsoar_on_prem)|(xsoar_saas)):
+    type: str
+  regex;(custom:(xsoar)|(marketplacev2)|(xpanse)|(xsoar_on_prem)|(xsoar_saas)):
+    type: bool
+  regex;(definitionId:(xsoar)|(marketplacev2)|(xpanse)|(xsoar_on_prem)|(xsoar_saas)):
+    type: str
+  regex;(marketplaces:(xsoar)|(marketplacev2)|(xpanse)|(xsoar_on_prem)|(xsoar_saas)):
```
yuvalbenshalom, line 98: is there a point to add this to marketplaces?

demisto_sdk/commands/content_graph/objects/content_item.py
```diff
@@ -221,9 +221,7 @@ def prepare_for_upload(
             raise FileNotFoundError(f"Could not find file {self.path}")
         data = self.data
         logger.debug(f"preparing {self.path}")
-        return MarketplaceSuffixPreparer.prepare(
-            data, current_marketplace, self.marketplaces
```
yuvalbenshalom, line 225: how is this change relevant?

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
```diff
@@ -26,37 +23,51 @@ def prepare(
...
-                    if isinstance(key, str) and key.casefold().endswith(suffix):
-                        clean_key = key[:-suffix_len]  # without suffix
+                    if isinstance(value, (list, dict)):
+                        fix_recursively(value)
+                        continue
+                    if SEPARATOR not in key:
+                        continue
+                    for suffix in suffixes:
+                        # iterate each suffix to see if it's relevant for the key.
+                        # the order of the suffixes matter!
```
yuvalbenshalom, line 49: lets go over the cases, I like to understand the order importance.

demisto_sdk/commands/prepare_content/tests/marketplace_suffix_preparer_test.py
```diff
@@ -0,0 +1,119 @@
...
+        "ef:xsoar_on_prem": "xsoar_on_prem",
+        "gh": "test4",
+        "gh:xsoar_saas": "xsoar_saas",
+        "ij": "test5",
+        "ij:marketplacev2": "marketplacev2",
+    },
+}
+
+
+def test_remove_xsoar():
```
yuvalbenshalom, line 33: please add docstrings explaining the cases.

demisto_sdk/scripts/add_marketplace_to_schemas.py
```diff
@@ -0,0 +1,52 @@
+from copy import deepcopy
+from pathlib import Path
```
yuvalbenshalom, line 2: let's talk about what ever this should be in the repo, and if so, it shall include UTs.

yuvalbenshalom (commented):
looks ok, but I like us to discuss if we are really interested in adding all posibblties to all schemas, as it is complex and 99% of it will probably never be used.

demisto_sdk/commands/content_graph/objects/content_item.py
ilaner: This argument wasn't used and is not relevant.
We prepare the content item for a specific marketplace, and we need to provide only the marketplace we work on, and not all the marketplaces this content item supports.

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
ilaner: OK.
In general: if running on XSOAR_SAAS and we have the following keys:
<key>:xsoar_saas
<key>:xsoar
<key>

We should priorotize the xsoar_saas key over the xsoar. However, if there is only xsoar, we should take it and not the regular key.

yuvalbenshalom (approved):

ilaner:
Test plan:
1. Added unit-tests for the suffix preparer, to run on custom data (not content-like)
2. Fixed unit-tests for content data to handle the new syntax and not _x2.
3. run prepare-content with all markteplaces, and show that the value in the key was modified in the result.
4. Existing schemas still work: validate -a passes
5. Added tests for the new schema syntax, make sure that <supported_key>:<marketplace> passes validation

PR #3834, opened 2023-11-30
 auto-detect file models by path

GuyAfik:
## Description
* Allow to find the file-models by auto detection according to suffix/names/path/content-type.


ilaner (approved):
looks good!

PR #3836, opened 2023-12-03
[Pre-commit] swith to docker hook instead of run-unit-tests

ilaner:
related: https://jira-dc.paloaltonetworks.com/browse/CIAC-8987


ilaner:
already done

PR #3837, opened 2023-12-03
fix exclude in pre-commit

ilaner:
Fix the exclude dictionary on python2 items


demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -59,6 +59,7 @@
     "run-unit-tests",
     "validate",
     "format",
+    "pylint-in-docker",
```
JudahSchwartz, line 62: lets open a task to bring this to the config file

JudahSchwartz (approved):

PR #3838, opened 2023-12-03
[Pre-Commit] fix update additional dependencies to use a deep copy

ilaner:
When comparing between the original pre-commit and the updated pre-commit, we should compare a deep copy of the original one.

## Related Issues
fixes:


mmhw (approved):
Good catch!

PR #3846, opened 2023-12-04
[clients] - run_cli_command method improvments

GuyAfik:
## Description
* Detect automatically for xsoar-6 the playground ID if `investigation_id` isn't provided when running the cli-command.
* Return all entries coming to the war-room from the `run_cli_command` method


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
 ## Unreleased
+* Added support to detect automatically the playground-id when running cli commands in xsoar-6.
```
JudahSchwartz, line 3: what command is this reffering to? Is there a denistosdk command to run something from the xsoar cli? Otherwise lets specify what exactly this affects

JudahSchwartz (approved):

CHANGELOG.md
GuyAfik: Do you think it can be clearer in some way than it is now?

PR #3865, opened 2023-12-13
fix test collection in generate-unit-tests

ilaner:
Move the import of `klara` modules inside the test functions so it will not be loaded in test collection


GuyAfik (approved):
Good job! 🐛

PR #3866, opened 2023-12-14
[Pre-Commit] Improve Readability of Collected Files

MichaelYochpaz:
## Description
Screenshot before:
Screenshot 2023-12-14 at 10 24 17

Screenshot after:
Screenshot 2023-12-14 at 10 26 38


MichaelYochpaz:
@mmhw Are these log lines not duplicates? They seem to list the same files.
I noticed the `This is the files that pre-commit received, but in fact it will run on files returned from group_by_python_version` comment. Does that mean that the list could be different in some cases?

CHANGELOG.md
```diff
@@ -8,7 +8,8 @@
 * Fixed an issue where *validate* command failed with release notes files for assets modeling rules folder.
 * Added support for `skip` property in **pre-commit** hooks.
 * **generate-unit-tests** command will require installation with `pip install demisto-sdk[generate-unit-tests]`.
-* added the *IN150* and *IN161* errors to *allowed ignore errors* list.
+* Added the *IN150* and *IN161* errors to *allowed ignore errors* list.
+* Updated the output of the **pre-commit** command to improve the readability of collected files
```
GuyAfik, line 12:
```suggestion
```
don't think RN is needed here

GuyAfik (approved):
LGTM, nice

mmhw:
> @mmhw Are these log lines not duplicates? They seem to list the same files. I noticed the `This is the files that pre-commit received, but in fact it will run on files returned from group_by_python_version` comment. Does that mean that the list could be different in some cases?

In case the pre-commit only receives a yml file of a script, it will actually run on its python file as well, so there will be a difference between the two lists.

ilaner (approved):
nice

mmhw:
@ilaner I just hope that when we run it with the `--all` argument, it won't make the log too big, maybe in that case it's better not to show the files at all?

MichaelYochpaz:
> @ilaner I just hope that when we run it with the `--all` argument, it won't make the log too big, maybe in that case it's better not to show the files at all?

That's exactly what @ilaner told me on Slack :)
We've updated it to print the log only when not using the `--all` flag.

demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -474,17 +474,23 @@ def prepare_and_run(
...
-            python_version,
-            changed_files_by_version,
-        ) in self.python_version_to_files.items():
-            changed_files_string = ", ".join(
-                sorted(str(file) for file in changed_files_by_version)
-            )
+        if self.all_files:
             logger.info(
-                f"Running pre-commit with Python {python_version} on {changed_files_string}"
+                f"Running pre-commit with on all files"
```
MichaelYochpaz, line 479:
```suggestion
                f"Running pre-commit on all files"
```

PR #3872, opened 2023-12-17
install specific version of neo4j in ci

ilaner:
Install specific version of neo4j which consists the APOC version

fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9209


GuyAfik (approved):
Good job

PR #3878, opened 2023-12-19
[Pre-Commit] Continue Even if Fetch Failed

MichaelYochpaz:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9231

## Description
Fix an issue (visible mainly on pre-commit) where if fetching from git fails due to connection problems, the whole command fails, even though it should be able to continue (just won't run fully up-to-date)

Screenshot 2023-12-19 at 17 59 36


GuyAfik:
@MichaelYochpaz does it fail because of connection errors? if yes lets use the `retry` decorator

demisto_sdk/commands/common/git_util.py
```diff
@@ -20,6 +20,8 @@
     DEMISTO_GIT_UPSTREAM,
     PACKS_FOLDER,
 )
+from demisto_sdk.commands.common.logger import logger
```
GuyAfik, line 23: that will probably cause circular import

MichaelYochpaz:
> @MichaelYochpaz does it fail because of connection errors? if yes lets use the `retry` decorator (you can find it in `tools.py`)
This PR is for a case where there is no connection to GitHub for some reason (like firewall issues), to allow the `pre-commit` flow to continue even without fetching.

@GuyAfik Adding a retry mechanism for a different case (where it's an issue that might be fixed on a retry), is a good idea, but handles a different case.

Current code handles a case where there is no connection at all (like firewall issues, where retrying won't help).
But we can do retries first, and then if they all failed, skip the fetch. Will probably be better than the current code with no retries.

However, it's going to be problematic as the `retry` wrapper is in the `tools` module, which import the `git_util` module that contains the `fetch` method. So this will cause circular imports issues.
The updated code would look like:
```python
class GitUtil:

    @retry(times=3, exceptions=(ConnectionError, Timeout))
    @lru_cache
    def fetch(self):
        self.repo.remote().fetch()
...
```

We can create a new module (that doesn't have any internal imports, like `constants` and `logger`) and move the `retry` wrapper there if you think that's a good idea.

.changelog/3878.yml
```diff
@@ -0,0 +1,6 @@
+changes:
+- description: Fixed an issue where the **pre-commit** command would fail if there are connection issues with the git repository.
+  type: fix
+- description: Fixed an issue where if running `git fetch` would fail on `git_util._get_all_changed_files`, an error would be raised.
+  type: internal
```
GuyAfik, line 5:
```suggestion
```

not needed

GuyAfik (approved):

PR #3882, opened 2023-12-20
[clients] - log out the reason of failed-playbook in case of error

GuyAfik:
## Related Issues
fixes:

## Description
Add to the `RunTimeException` a log that indicates what's the reason that playbook failed if it reached a failed state.


JudahSchwartz (approved):
logging 🧙

PR #3886, opened 2023-12-20
[clients] - auto detect using new api fields

GuyAfik:
## Description
* Get the api clients by **productMode** and **deploymentMode** (see [this](https://jira-dc.paloaltonetworks.com/browse/CRTX-100719) issue)
* Fall-back to previous implementation in server version where **productMode** and **deploymentMode** are not implemented in old server version
* Fixed an issue where the `get_client_from_server_type` thought xsoar-on-prem is xsoar-saas in cases where the from_version of xsoar-on-prem is > 8.x.x


ilaner (approved):
looks good.

PR #3891, opened 2023-12-22
Mock tests using sleep

GuyAfik:
## Related Issues
fixes:

### DIFF
--- a/demisto_sdk/commands/common/tests/pack_unique_files_test.py
+++ b/demisto_sdk/commands/common/tests/pack_unique_files_test.py
@@ -1063,7 +1063,7 @@ class TestPackUniqueFilesValidator:
             not in errors
         )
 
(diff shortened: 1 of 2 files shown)


JudahSchwartz (commented):
to me this doesnt even look like whoever wrote these tests made a mistake. This isnt something we can expect devs to catch
A better solution wiould be to automock sleep always. Perhaps we can consider a global autouse pytest fixture

JudahSchwartz (approved):

GuyAfik:
> to me this doesnt even look like whoever wrote these tests made a mistake. This isnt something we can expect devs to catch A better solution wiould be to automock sleep always. Perhaps we can consider a global autouse pytest fixture

As already being said, mocking `sleep` depends on which module actually imports it, mocking the `time.sleep` isn't a good solution, we'll find a way to handle it.

PR #3892, opened 2023-12-23
[clients] - improve error entries messages

GuyAfik:
## Related Issues
fixes:

## Description
* align the error messages to fit both for a command executed in a playground and a command executed in a playbook's task, use the same code for both cases.

### DIFF
--- a/demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
+++ b/demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
@@ -4,7 +4,7 @@ import socket
 import time
 import urllib.parse
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Set, Tuple, Union
 from urllib.parse import urlparse
 
 import dateparser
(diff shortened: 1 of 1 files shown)


demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
```diff
@@ -980,6 +982,37 @@ def run_cli_command(
...
+
+        Returns:
+            Formatted error entries
+        """
+        error_entries: Set[str] = set()
+
+        for entry in entries:
+            if entry.type == self._ENTRY_TYPE_ERROR and entry.parent_content:
+                # Checks for passwords and replaces them with "******"
+                parent_content = re.sub(
```
JudahSchwartz, line 1000: what are you checking here? Why is the entry printing passwords? Is this the only case we need to check for?

GuyAfik: we talked about it previously [here](https://github.com/demisto/demisto-sdk/pull/3732#discussion_r1382542636)

JudahSchwartz: are there other forms of authentication or other secrets that might be printed?

GuyAfik: nope

JudahSchwartz (approved):

PR #3897, opened 2023-12-25
[Pre-Commit] Execute fixable hooks before rest of the hooks

ilaner:
This adds support for `fixable` property in pre-commit-config, so we will be able to add `fixable` on the property, and the fixable hooks will run before the rest of the hooks

fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9269


JudahSchwartz (commented):
think we should solve this issue by making a run_order argument or something that better indicates what we're trying to do.
meaning if i have 3 things that say 
`run_order: 1`
then those run at the same time and we dont care of order, then 
`run_order: 2` would be the next batch etc.

`fixable: true` is a bit misleading because it sounds like without having this it wont fix things, which isnt true

we can probably come up with a better name than run_order

order_index maybe?

PR #3903, opened 2023-12-26
Add validation for non stash in demisto-sdk update script

MosheEichler:
## Related Issues
fixes:

## Description
run `git stash pop` only if we stash something in the begging of the script.


mmhw (approved):
Nice!

PR #3916, opened 2024-01-01
dont generate coverage xml

ilaner:
`coverage.xml` file is generated in `coverage analyze` hook. No need to generate it in the `merge-pytest-reports` anymore.


GuyAfik (approved):
good job

PR #3932, opened 2024-01-10
Refactoring TestSuite to support graph operations

ilappe:
## Related Issues
fixes:https://jira-dc.paloaltonetworks.com/browse/CIAC-7115

## Description
- added a base class for all the TestSuite objects to support graph operations
- refactor the following test to use the new TestSuite abilities:
  - `GR_validators_test`
  - `create_content_graph_test`


TestSuite/test_suite_base.py
```diff
@@ -0,0 +1,31 @@
...
+
+    @property
+    def object(self):
+        return BaseContent.from_path(self.obj_path)
+
+    def get_graph_object(self, interface: ContentGraphInterface):
+        return interface.from_path(self.obj_path)
+
+    @abstractmethod
+    def set_data(self, key_path_to_val: dict):
```
ilaner, line 20: why can't we implement it here?

TestSuite/test_suite_base.py
```diff
@@ -0,0 +1,31 @@
...
+class TestSuiteBase:
+    def __init__(self, path: Path):
+        self.obj_path = path
+
+    @property
+    def object(self):
+        return BaseContent.from_path(self.obj_path)
+
+    def get_graph_object(self, interface: ContentGraphInterface):
+        return interface.from_path(self.obj_path)
```
ilaner, line 17: please document the difference. Document that this is with `relationships_data` populated.

conftest.py
```diff
@@ -87,6 +90,38 @@ def repo(request: FixtureRequest, tmp_path_factory: TempPathFactory) -> Repo:
...
+    import demisto_sdk.commands.content_graph.objects.base_content as bc
+
+    repo = get_repo(request, tmp_path_factory)
+
+    bc.CONTENT_PATH = Path(repo.path)
+    neo4j_path = bc.CONTENT_PATH.parent.parent / "neo4j"
+
+    mock.patch.object(ContentGraphInterface, "repo_path", bc.CONTENT_PATH)
+    mock.patch.object(neo4j_service, "REPO_PATH", neo4j_path)
+    mock.patch(
```
ilaner, line 108: I don't think we need this mock in conftest. this should be part of the test

demisto_sdk/commands/content_graph/tests/get_relationships_test.py
```diff
@@ -15,14 +22,13 @@
...
-    repository,  # noqa: F401
-    setup_method,  # noqa: F401
-)
+from demisto_sdk.commands.content_graph.tests import create_content_graph_test
+
+
+@pytest.fixture(autouse=True)
+def setup_method(mocker, graph_repo):
+    """Auto-used fixture for setup before every test run"""
+    create_content_graph_test.setup_method(mocker, graph_repo)
```
ilaner, line 31: why do we need it?

ilaner (commented):
Good job. Please see comments

TestSuite/test_suite_base.py
ilappe: Because it dependent on the file structure YML or JSON so implemented in those classes

ilaner: I think a better solution is to have a property if this content item is JSON or YAML, and set data accordingly, instead implementing the same method in all classes

ilappe: as we discussed - due to this implemented only in the base classes (YML & JSON), the code will remain as is

ilaner (approved):
Good job!

PR #3958, opened 2024-01-17
IntegrationScript - update docker_image to DockerImage object

GuyAfik:
## Description
* This PR changes the `dockerimage` field from `Optional[str]` to `DockerImage` object to allow doing extra operations on a docker-image such as getting its latest tag / metadata / creation-date, python-version, digest and more.
* The `DockerImage` class inherits the `str` class, hence the `DockerImage` is considered a `primitive` type.

that would allow us to do operations directly on docker-images objects
```
script = BaseContent.from_path(path)
latest_tag = script.docker_image.latest_tag
creation_date = script.docker_image.creation_date
current_tag = script.docker_image.tag
python_version = script.docker_image.python_version
```


demisto_sdk/commands/content_graph/objects/integration_script.py
```diff
@@ -26,7 +26,7 @@
 class IntegrationScript(ContentItem):
     type: str
     subtype: Optional[str]
-    docker_image: Optional[str]
+    docker_image: DockerImage = DockerImage("")
```
ilaner, line 29: why not Optional[DockerImage]?

demisto_sdk/commands/content_graph/parsers/integration_script.py
```diff
@@ -31,9 +32,13 @@ def field_mapping(self):
         super().field_mapping.update({"object_id": "commonfields.id"})
         return super().field_mapping
 
-    @property
-    def docker_image(self) -> str:
-        return get_value(self.yml_data, self.field_mapping.get("docker_image", ""), "")
+    @cached_property
+    def docker_image(self) -> DockerImage:
+        docker_image = (
```
ilaner, line 37: What if it is missing?
We have two options:
1. it is a python script which runs on alpine.
2. It is javascript which runs natively.

Maybe it worth distinguish them?

demisto_sdk/commands/validate/validators/DO_validators/DO100_docker_image_tag_is_not_latest.py
```diff
@@ -36,20 +37,22 @@ def is_valid(self, content_items: Iterable[ContentTypes]) -> List[ValidationResu
...
     ) -> FixResult:
-        docker_image = content_item.docker_image_object
+        docker_image = content_item.docker_image
         try:
             latest_numeric_tag = docker_image.latest_tag
             message = self.fix_message.format(
-                content_item.docker_image, f"{docker_image.name}:{latest_numeric_tag}"
+                docker_image, f"{docker_image.name}:{latest_numeric_tag}"
+            )
+            content_item.docker_image = DockerImage(
```
ilaner, line 53: You checked that it works?

ilaner (commented):
Very good job!

demisto_sdk/commands/content_graph/objects/integration_script.py
GuyAfik: we could, but it can be annoying with mypy, we would need to do:
```
if script.docker_image: (to make sure its not `None`)
  ....
```

demisto_sdk/commands/validate/validators/DO_validators/DO100_docker_image_tag_is_not_latest.py
GuyAfik: yep works just fine

demisto_sdk/commands/content_graph/parsers/integration_script.py
GuyAfik: what's wrong if the DockerImage will be an empty string and not `None`? how would it make any difference?

GuyAfik: You are right basically, but this was done because of https://github.com/demisto/demisto-sdk/pull/3958#discussion_r1468853116

ilaner (approved):
really good job

GuyAfik:
Creating the graph from this PR keeps the docker-image field to behave the same, it is populated as a string in the graph.
image

PR #3969, opened 2024-01-22
validate changelog job - skip when running CI on master

GuyAfik:
## Description

Works on commit: https://github.com/demisto/demisto-sdk/pull/3969/commits/44d58220ea487793ab49a8a0e5a8f96860b97985

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -16,7 +16,7 @@ concurrency:
 jobs:
   validate-changelog:
     runs-on: ubuntu-latest
-    if: github.ref_name != 'master'
+    if: github.ref_name != 'skip_changelog_execution_master'
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/on-push.yml
```diff
@@ -16,6 +16,7 @@ concurrency:
 jobs:
   validate-changelog:
     runs-on: ubuntu-latest
+    if: github.event.pull_request.base.ref == 'master'
```
dorschw, line 19: so this only limits us to pull requests to master?

GuyAfik: it would be better to use `ref_name`

from github docs:
```
The short ref name of the branch or tag that triggered the workflow run. This value matches the branch or tag name shown on GitHub.
```

dorschw (approved):

PR #3970, opened 2024-01-22
Fix upload to handle approaching value in explicit way- e.g ${}

RosenbergYehuda:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-31260

## Description

- When a user uses the upload command to upload a playbook that contains a task with a field name "CustomFields", if the value of that field has been retrieved from the context data using the explicit method - **${}**, the command will fail and raise an error. In this PR that will be fixed.
- Found a bug in the _get_fields_by_script_argument function regarding handling JSON input and fixed it.

### DIFF
--- a/.changelog/3970.yml
+++ b/.changelog/3970.yml
@@ -1,0 +1,4 @@
+changes:
+- description: Fixed an issue in **upload**  where customFields with explicitly defined values (e.g., ${}) were failing.
+  type: fix
+pr_number: 3970
--- a/demisto_sdk/commands/common/tests/update_id_set_test.py
+++ b/demisto_sdk/commands/common/tests/update_id_set_test.py
@@ -2880,16 +2880,16 @@ class TestGenericFunctions:
         },
     }
     @staticmethod
-    def test_get_fields_by_script_argument__explicit_method(task = custom_fields_with_explicit_method):
+    def test_get_fields_by_script_argument__explicit_method(custom_fields_with_explicit_method):
         """
         Given
             - A playbook task with custom fields that retrieves its values in an explicit way (using the ${} method instead of "get" method)
         When
-            -  'get_fields_by_script_argument' is called
+            -  Searching for dependent incident fields in the task script arguments
         Then
-            -  The function should return an empty set with no errors
+            -  The function should return an empty set with no errors, since a the custom field - ${} sould be ignored
         """
-        result = get_fields_by_script_argument(task)
+        result = get_fields_by_script_argument(custom_fields_with_explicit_method)
         assert result ==set()
         
 
@@ -2900,16 +2900,16 @@ class TestGenericFunctions:
         },
     }
     @staticmethod
-    def test_get_fields_by_script_argument__json(task = custom_fields_as_a_json):
+    def test_get_fields_by_script_argument__json(custom_fields_as_a_json):
         """
         Given
             - A playbook task with custom fields value is a json
         When
-            -  'get_fields_by_script_argument' is called
+            -  Searching for dependent incident fields in the task script arguments
         Then
-            -  The function should return a set with the json keys that are not in BUILT_IN_FIELDS
+            -  The function should return all dependent incident custom fields in the task
         """
-        result = get_fields_by_script_argument(task)
+        result = get_fields_by_script_argument(custom_fields_as_a_json)
         assert result =={'root', 'test_key'}
 
 
--- a/demisto_sdk/commands/common/tests/update_id_set_test.py
+++ b/demisto_sdk/commands/common/tests/update_id_set_test.py
@@ -2871,6 +2871,26 @@ class TestGenericFunctions:
 
         result = get_fields_by_script_argument(task)
         assert "field_name" in result
+    
+    
+    custom_fields_with_explicit_method ={
+        "id": "ID",
+        "scriptarguments": {
+            "customFields": {'simple': '${keys}'}
+        },
+    }
+    @staticmethod
+    def test(task = custom_fields_with_explicit_method):
+        """
+        Given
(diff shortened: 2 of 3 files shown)


demisto_sdk/commands/common/tests/update_id_set_test.py
```diff
@@ -2871,6 +2871,46 @@ def test_get_fields_by_script_argument(task):
...
+    
+    
+    custom_fields_with_explicit_method ={
+        "id": "ID",
+        "scriptarguments": {
+            "customFields": {'simple': '${keys}'}
+        },
+    }
+    @staticmethod
+    def test_get_fields_by_script_argument__explicit_method(task = custom_fields_with_explicit_method):
```
GuyAfik, line 2883: not a good idea to put a default value of mutable variable.

demisto_sdk/commands/common/tests/update_id_set_test.py
```diff
@@ -2871,6 +2871,46 @@ def test_get_fields_by_script_argument(task):
...
+        
+
+    custom_fields_as_a_json ={
+        "id": "ID",
+        "scriptarguments": {
+            "customFields": {'complex': {'root': 'keys', 'test_key': 'tets_value','id': 'ID'}}
+        },
+    }
+    @staticmethod
+    def test_get_fields_by_script_argument__json(task = custom_fields_as_a_json):
```
GuyAfik, line 2903: same as above

demisto_sdk/commands/common/update_id_set.py
```diff
@@ -679,13 +679,19 @@ def get_fields_by_script_argument(task):
...
                 if isinstance(custom_field_value, str):
-                    custom_fields_list = json.loads(custom_field_value)
-                    if not isinstance(custom_fields_list, list):
-                        custom_fields_list = [custom_fields_list]
-                    for custom_field in custom_fields_list:
-                        for field_name in custom_field.keys():
-                            if field_name not in BUILT_IN_FIELDS:
-                                dependent_incident_fields.add(field_name)
+                    if custom_field_value.startswith("$"):
+                        logger.info("You're using an unrecommended method - ${} - to retrieve values from the context data.")
```
GuyAfik, line 683:
```suggestion
                        logger.warning("You're using an unrecommended method - ${} - to retrieve values from the context data.")
```
I would make it warning

demisto_sdk/commands/common/tests/update_id_set_test.py
```diff
@@ -2871,6 +2871,46 @@ def test_get_fields_by_script_argument(task):
...
+        """
+        Given
+            - A playbook task with custom fields value is a json
+        When
+            -  'get_fields_by_script_argument' is called
+        Then
+            -  The function should return a set with the json keys that are not in BUILT_IN_FIELDS
```
GuyAfik, line 2910: That doesn't say much, what is `BUILT_IN_FIELDS`?
give a brief of what the expected output should be here in a "high level", rather than "code level"

GuyAfik (commented):
@RosenbergYehuda nice, did you test it?

RosenbergYehuda:
Yes, this was tested and works.

demisto_sdk/commands/common/update_id_set.py
RosenbergYehuda: Fixed.

demisto_sdk/commands/common/tests/update_id_set_test.py
RosenbergYehuda: fixed

demisto_sdk/commands/common/tests/update_id_set_test.py
RosenbergYehuda: fixed

demisto_sdk/commands/common/tests/update_id_set_test.py
```diff
@@ -2871,6 +2871,46 @@ def test_get_fields_by_script_argument(task):
...
+        """
+        Given
+            - A playbook task with custom fields that retrieves its values in an explicit way (using the ${} method instead of "get" method)
+        When
+            -  Searching for dependent incident fields in the task script arguments
+        Then
+            -  The function should return an empty set with no errors, since a the custom field - ${} sould be ignored
+        """
+        result = get_fields_by_script_argument(custom_fields_with_explicit_method)
+        assert result ==set()
```
GuyAfik, line 2893:
```suggestion
    @staticmethod
    def test_get_fields_by_script_argument__explicit_method():
        """
        Given
            - A playbook task with custom fields that retrieves its values in an explicit way (using the ${} method instead of "get" method)
        When
            -  Searching for dependent incident fields in the task script arguments
        Then
            -  The function should return an empty set with no errors, since a the custom field - ${} sould be ignored
        """
            custom_fields_with_explicit_method ={
        "id": "ID",
        "scriptarguments": {
            "customFields": {'simple': '${keys}'}
        },
    }
        result = get_fields_by_script_argument(custom_fields_with_explicit_method)
        assert result ==set()
```

do the same for the other test

RosenbergYehuda: Done

.changelog/3970.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Fixed an issue in **upload**  where customFields with explicitly defined values (e.g., ${}) were failing.
```
GuyAfik, line 2:
```suggestion
- description: Fixed an issue in **upload** where customFields with explicitly defined values (e.g., ${}) caused the command to fail.
```

is that correct?

GuyAfik (approved):

ilaner (approved):
Good job!

PR #3985, opened 2024-01-28
[clients] - Refactor classes not to use pydantic

GuyAfik:
## Description
* Check if xsoar server is healthy for any type of server (xsoar/xsoar-saas/xsiam) - raise exception if server isn't healthy
* Refactor the clients objects not to use pydantic.
* [Successful SDK nightly](https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/pipelines/799043)


demisto_sdk/commands/common/clients/__init__.py
```diff
@@ -61,41 +64,54 @@ def get_client_from_marketplace(
...
+            api_key=_api_key,
+            auth_id=_auth_id,
+            verify_ssl=_verify_ssl,
         )
     return get_client_from_config(config)
 
 
 @lru_cache
 def get_client_from_server_type(
-    base_url: Optional[str] = os.getenv(DEMISTO_BASE_URL, ""),
```
JudahSchwartz, line 93: why move this into the function body?

demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
```diff
@@ -48,6 +46,26 @@ class XsoarClient(BaseModel):
     class Config:
         arbitrary_types_allowed = True
 
+    @classmethod
+    def is_xsoar_healthy(cls, client: DefaultApi) -> bool:
```
JudahSchwartz, line 50: is_server_healthy and lets use inheritance

demisto_sdk/commands/common/clients/xsoar_saas/xsoar_saas_api_client.py
```diff
@@ -34,6 +39,35 @@ def is_xsoar_saas(
...
+                response.status_code == requests.codes.ok
+                and server_health_status == "available"
+            )
+        except JSONDecodeError as e:
+            logger.debug(
+                f"Could not validate if {server_config} is healthy, error:\n{e}"
+            )
+            return False
+
     @validator("session", always=True)
```
JudahSchwartz, line 71: the use of validators in this class to init the internals is a bit abuse of pydantic. This object isnt meant to be serializable 
We should refactor this to be a normal class that the init method initializes it's parts.

demisto_sdk/commands/common/clients/__init__.py
```diff
@@ -112,14 +128,31 @@ def get_client_from_server_type(
...
-        password=password,
+        base_url=_base_url,
+        api_key=_api_key,
+        auth_id=_auth_id,
+        verify_ssl=_verify_ssl,
+        username=_username,
+        password=_password,
     )
 
+    if not XsoarClient.is_xsoar_healthy(_client):
```
JudahSchwartz, line 151: we should have this check happen based on a init param
check_server_health
with a default True

JudahSchwartz (commented):
Looks good, refactor requested

demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
GuyAfik: called it `is_healthy`

demisto_sdk/commands/common/clients/__init__.py
GuyAfik: because when we define default values, python automatically reads them if the module that is being imported in some sort of way, that can cause un-expected issues, for example we might read env vars before reading `.env` file, that is just more secure way of doing it.

demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
GuyAfik: done

demisto_sdk/commands/common/clients/__init__.py
GuyAfik: done

demisto_sdk/commands/common/clients/xsoar_saas/xsoar_saas_api_client.py
GuyAfik: done, the clients will not be based on pydantic anymore

demisto_sdk/commands/common/clients/__init__.py
```diff
@@ -112,75 +128,69 @@ def get_client_from_server_type(
...
-                password=password,
-                verify_ssl=verify_ssl,
-            ),
-            client=_client,
-            about=about,
-        )
-
-    raise RuntimeError(f"Could not determine the correct api-client for {base_url}")
+        except InvalidServerType as error:
+            logger.debug(f"{error=}")
```
JudahSchwartz, line 179: should be a warn with an explanation that continueing anyway as xsoar (and why)

demisto_sdk/commands/common/clients/xsoar_saas/xsoar_saas_api_client.py
```diff
@@ -19,48 +26,92 @@ class XsoarSaasClient(XsoarClient):
...
-        cls,
-        server_version: str,
-        product_mode: Optional[str] = None,
-        deployment_mode: Optional[str] = None,
+    def __init__(
+        self,
+        config: XsoarClientConfig,
+        client: Optional[DefaultApi] = None,
+        raise_if_server_not_healthy: bool = True,
+        should_validate_server_type: bool = False,
```
JudahSchwartz, line 34:
```suggestion
        validate_server_type: bool = False,
```

.changelog/3985.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Added support to check the health status when initialing a client for xsoar-on-prem/xsoar-saas.
```
JudahSchwartz, line 2: note here that by default will throw exceptions if unhealthy now

JudahSchwartz (approved):
nice change! 🧑‍⚕️

demisto_sdk/commands/common/clients/xsoar_saas/xsoar_saas_api_client.py
GuyAfik: why? should indicates its a boolean IMO

PR #3986, opened 2024-01-28
[CI] - Add default timeouts to cancel hanging jobs

GuyAfik:
## Description
* Do not run `validate-changelog` when running in master branch as its redundant, works in commit https://github.com/demisto/demisto-sdk/pull/3986/commits/1a64c346d5b151db604a64db860111cbe0634fd6
* Add default-timeouts of 60 minutes for hanging jobs/steps, that means that If the steps / flows haven't finished in 1 hour, they will be automatically cancelled.
* Upgrade node to 20 from 16 as 16 is deprecated


ilaner (approved):
nice

PR #3971, opened 2024-01-23
fix validate failing on list data.json

merit-maita:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-32391

## Description
Fixed a bug in validate command, where it failed with the data json list in Lists folder
fixed a bug with validating pack lists

PR #3997, opened 2024-01-30
fixed a bug in validate lists files

merit-maita:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-32391

## Description
Fixed a bug in validate command with _data.json file types in list directory. (Follow up to #3971)


demisto_sdk/commands/validate/old_validate_manager.py
```diff
@@ -747,7 +747,7 @@ def is_skipped_file(self, file_path: str) -> bool:
                 path.name == "CommonServerPython.py"
                 and path.parent.parent.name != "Base"
             )
-            or (path.parent.name == LISTS_DIR and path.name.endswith("_data.json"))
+            or (LISTS_DIR in path.parts and path.name.endswith("_data.json"))
```
dorschw, line 750: why `LISTS_DIR in path.parts` and not `path.parent.parent.name == LISTS_DIR`? I think it's always two levels deep.

dorschw (approved):

PR #4006, opened 2024-01-31
Added fix

anas-yousef:
## Related Issues
fixes:

## Description
Fixed an issue where validate would state that it failed, even though it shouldn't, and then it would display an empty list of content items.

### DIFF
--- a/demisto_sdk/commands/common/hook_validations/readme.py
+++ b/demisto_sdk/commands/common/hook_validations/readme.py
@@ -335,6 +335,7 @@ class ReadMeValidator(BaseValidator):
             return False
         return True
 
+    @error_codes("RM112")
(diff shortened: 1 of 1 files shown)


.changelog/4006.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Fixed an issue where validate would state that it failed, even though it shouldn't, and then it would display an empty list of content items.
```
YuvHayun, line 2:
```suggestion
- description: Fixed an issue where validate would state that it failed, even though it shouldn't, and then it would display an empty list of content items when running on specific validations.
```

YuvHayun (changes requested):

YuvHayun (approved):

PR #4010, opened 2024-02-01
Move Neo4j Cache Directory To  Project Directory

MichaelYochpaz:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9617

## Description
Move the Neo4j directory to a constant directory within project's directory.


demisto_sdk/commands/common/constants.py
```diff
@@ -11,7 +11,10 @@
 
 
 PROJECT_DATA_DIR = Path.home() / ".demisto-sdk"
+CACHE_DIR = PROJECT_DATA_DIR / "cache"
 LOGS_DIR = PROJECT_DATA_DIR / "logs"
+NEO4J_DIR = CACHE_DIR / "neo4j"
```
ilaner, line 16: That's not exactly the cache. Lets put it directly on `demisto-sdk/neo4j`

demisto_sdk/commands/content_graph/interface/neo4j/neo4j_graph.py
```diff
@@ -108,12 +108,10 @@ class NoModelException(Exception):
 
 
 class Neo4jContentGraphInterface(ContentGraphInterface):
-    # this is used to save cache of packs and integrations which queried
-    _import_handler = Neo4jImportHandler()
-
```
MichaelYochpaz, line 113: This caused `Neo4jImportHandler()` to run whenever this module is imported, even if an instance of `Neo4jContentGraphInterface` was never initialized, creating the `neo4j` directory even on unrelated commands.

In this PR I updated `Neo4jImportHandler` to be a singleton, and then initialized `_import_handler` in `__init__` to resolve this issue.

demisto_sdk/commands/common/singleton.py
```diff
@@ -1,6 +1,16 @@
...
 
 
+class SingletonMeta(type):
+    _instances = {}
+
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            cls._instances[cls] = super(SingletonMeta, cls).__call__(*args, **kwargs)
+
+        return cls._instances[cls]
```
GuyAfik, line 11: this saves multiple instances based on their args? why do we need a list of instances and not only a single one?

demisto_sdk/commands/common/constants.py
```diff
@@ -11,7 +11,10 @@
 
 
 PROJECT_DATA_DIR = Path.home() / ".demisto-sdk"
+CACHE_DIR = PROJECT_DATA_DIR / "cache"
 LOGS_DIR = PROJECT_DATA_DIR / "logs"
+NEO4J_DIR = PROJECT_DATA_DIR / "neo4j"
```
GuyAfik, line 16: +1 on moving it to the official `.demisto-sdk` folder :)

GuyAfik (commented):
@MichaelYochpaz kudos on moving it finally to the correct folder!

demisto_sdk/commands/common/singleton.py
MichaelYochpaz: Since this is a metaclass that can be used for multiple different classes across the SDK.

For example, if we have:
```python
class A(metaclass=SingletonMeta):
    pass

class B(metaclass=SingletonMeta):
    pass
```

After both have been created once, the `_instances` dict will contain `{A: <A class instance>, B: <B class instance>`.
Then when you try to create a new `A` / `B` instance, it will fetch the existing instance from the dict instead of creating a new one.

This is exact code is quite common, if you'll look up "Python Singleton implementation".

GuyAfik:
LGTM, please wait for @ilaner cr

demisto_sdk/commands/content_graph/interface/neo4j/import_utils.py
```diff
@@ -4,12 +4,13 @@
...
 
 from demisto_sdk.commands.common.logger import logger
+from demisto_sdk.commands.common.singleton import SingletonMeta
 from demisto_sdk.commands.content_graph.neo4j_service import get_neo4j_import_path
 
 GRAPHML_FILE_SUFFIX = ".graphml"
 
 
-class Neo4jImportHandler:
+class Neo4jImportHandler(metaclass=SingletonMeta):
```
ilaner, line 13: nice job with the meta

demisto_sdk/commands/content_graph/neo4j_service.py
```diff
@@ -6,16 +6,14 @@
...
 from demisto_sdk.commands.common.docker_helper import init_global_docker_client
 from demisto_sdk.commands.common.logger import logger
 from demisto_sdk.commands.content_graph.common import (
     NEO4J_DATABASE_HTTP,
-    NEO4J_FOLDER,
     NEO4J_PASSWORD,
 )
 
-REPO_PATH = CONTENT_PATH.absolute()
 NEO4J_VERSION = "5.13.0"
```
ilaner, line 17: i know it is not related but let's move this to constants and have an env library for the version to override it

ilaner (approved):
Awesome!

GuyAfik (approved):
Great change!

PR #4011, opened 2024-02-01
Load .env file in all commands

ilaner:
Move the load_dotenv function to the beginning of the `__main__` function.

fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9638


GuyAfik (approved):
Very nice catch!

GuyAfik:
@ilaner add small changelog for it

PR #4019, opened 2024-02-05
Catch docker errors when creating docker container

MosheEichler:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9664


demisto_sdk/commands/common/docker_helper.py
```diff
@@ -224,6 +224,17 @@ def copy_files_container(
...
+    def remove_container(self, name, **kwargs):
+        container: docker.models.containers.Container = (
+            init_global_docker_client().containers.get(name)
+        )
+        container.remove()
+
+    @retry(
+        times=3,
+        exceptions=(requests.exceptions.ConnectionError, requests.exceptions.Timeout),
+        recover=remove_container,
```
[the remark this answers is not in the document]
MosheEichler: @GuyAfik 

> if we failed to create the container due to connection errors, why would we need to remove it at all?

To avoid errors like this, the container probably created and for some (Timeout) exception it fails but the container name already exist and causes conflict errors.

> Also adding a recover function is kind of an abuse to the retry decorator IMO.

So what do you suggested?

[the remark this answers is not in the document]
GuyAfik: i think you should recover inside the `create_container` method rather than the `retry` decorator.

If you have ConnectionError exception, no way a container will be created, as to timeout, it might be possible that docker can create the container but won't provide any response about it. 

why do you need to use the `retry` decorator if you recover in case of an exception?
Don't you need to recover here only when you get 409?

.changelog/4019.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Added a recover function in the retry wrapper (in case of failures).
```
GuyAfik, line 2: update change-log

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -235,13 +239,21 @@ def create_container(
...
+
+        try:
+            container: docker.models.containers.Container = (
+                docker_client.containers.create(
+                    image=image, command=command, environment=environment, **kwargs
+                )
             )
-        )
+        except DockerException as e:
+            container.remove()
```
GuyAfik, line 251: i dont think it makes sense, the `container` variable would not be defined if an exception would be raised.
I suggest to get the container from a custom name that we used / from the image it was created with.

any-way during the exception you will need to initialize the existing container model.

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -235,13 +239,21 @@ def create_container(
...
+            raise e
```
GuyAfik, line 252: why raising? don't we want to remove the container and re-create it?

GuyAfik (changes requested):

demisto_sdk/commands/common/docker_helper.py
MosheEichler: We are already recreating the container with the wrapper, here we just want to remove the container if it has already been created and for some reason it fails

GuyAfik: but the retry wrapper doesn't catch `DockerException` so retry will not be re-triggered here.

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -235,13 +239,22 @@ def create_container(
...
+        try:
+            container: docker.models.containers.Container = (
+                docker_client.containers.create(
+                    image=image, command=command, environment=environment, **kwargs
+                )
             )
-        )
+        except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as e:
+            if container_name := kwargs.get("container_name"):
+                docker_client.containers.get(container_name).remove(force=True)
```
GuyAfik, line 252: in case we didn't create the container with a specific name, we should try to remove the container that was created with the image, meaning we should get the container model by the image and not only by the name.

GuyAfik (changes requested):
@MosheEichler I don't understand your implementation here, lets have a call on Sunday

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -224,6 +224,10 @@ def copy_files_container(
                 with open(tar_file_path.name, "rb") as byte_file:
                     container.put_archive("/", byte_file.read())
 
+    @retry(
+        times=3,
+        exceptions=(requests.exceptions.ConnectionError, requests.exceptions.Timeout),
```
GuyAfik, line 229: catch here only DockerException

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -235,13 +239,22 @@ def create_container(
...
+
+        try:
+            container: docker.models.containers.Container = (
+                docker_client.containers.create(
+                    image=image, command=command, environment=environment, **kwargs
+                )
             )
-        )
+        except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as e:
+            if container_name := kwargs.get("container_name"):
```
GuyAfik, line 251:
```suggestion
            if container_name := kwargs.get("name"):
```

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -235,13 +243,26 @@ def create_container(
...
+                    image=image, command=command, environment=environment, **kwargs
+                )
             )
-        )
+        except (
+            requests.exceptions.ConnectionError,
+            requests.exceptions.Timeout,
+            DockerException,
+        ) as e:
+            if container := docker_client.containers.get(kwargs.get("name")):
```
GuyAfik, line 259: make sure that this line won't raise an exception if the container is not found

GuyAfik (approved):
@MosheEichler looks great, just add a test.

demisto_sdk/commands/common/tests/docker_helper_test.py
```diff
@@ -86,3 +87,39 @@ def test_cache_of_get_python_version_from_image():
     docker_helper.get_python_version(image)
     cache_info = docker_helper.get_python_version.cache_info()
     assert cache_info.hits == cache_info_before.hits + 1
+
+
+@pytest.mark.parametrize('image_name, container_name, exception, exception_text', [
```
MosheEichler, line 92: @GuyAfik I added UT with parametrize please approve

GuyAfik: add given when then

GuyAfik (approved):
looks good

PR #4026, opened 2024-02-08
[File Management] - do not use pydantic anymore

GuyAfik:
## Description
* Do not use pydantic in the files objects.
* Validate that file exist within a git path in the specific branch before processing it.


ilaner (approved):
nice job :)

PR #4027, opened 2024-02-08
Add A Deprecation Notice To The `update-content-graph` Command

MichaelYochpaz:
## Description
Add a deprecation note to the `update-content-graph` command, which is replaced by the `graph update` command.
Screenshot 2024-02-08 at 12 06 43


MichaelYochpaz:
> @MichaelYochpaz its not also correct for `create-content-graph` command?

You're right, didn't know `create-content-graph` existed.
I've updated the PR to deprecate it as well.

GuyAfik (approved):
🥇 
Very nice

PR #4029, opened 2024-02-08
Fix issues with .env file support

ilaner:
fixes:
1. loads the env in the module scope, as the clients are loaded before the `main` function starts.
2. Removes PYTHONPATH and MYPYPATH from the .env since the VSCode Docker extension now loads the dotenv file into the docker debugging, and it contradicts the file system in the container (and we don't need it since we configure it in VSCode settings)


GuyAfik (approved):
Add test and change log, did you check that it works now with the clients?

PR #4035, opened 2024-02-11
[CI] - GitHub actions master slack notifier

GuyAfik:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9298

## Description
* Add slack-notifier for github-actions to report about falling jobs when merging to master if there are any.
* Test that it works as expected (https://github.com/demisto/demisto-sdk/actions/runs/7970304881/job/21758511785)


JudahSchwartz (changes requested):
https://github.com/marketplace/actions/slack-notify-build
can we use this instead? especially since there isnt rally much custom behavior here

GuyAfik:
> https://github.com/marketplace/actions/slack-notify-build can we use this instead? especially since there isnt really much custom behavior here

Not really, here's why:
* in the future I want to extend the behaviour to parse and report failed tests as well from a junit-file
* I want to be able to customize and parse the message that is being sent on my own
* Using built-in slack-notifier gives a very limited abilities

JudahSchwartz (approved):

PR #4037, opened 2024-02-11
Check if DEMISTO_SDK_GRAPH_FORCE_CREATE is set in update-content-graph

MosheEichler:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9707

## Description
Added a check if `DEMISTO_SDK_GRAPH_FORCE_CREATE` environment variable is set to true in update content graph.

### DIFF
--- a/.changelog/4037.yml
+++ b/.changelog/4037.yml
@@ -1,0 +1,4 @@
+changes:
+- description: Update 'update_content_graph' to create the graph from scratch if DEMISTO_SDK_GRAPH_FORCE_CREATE env is set to true.
+  type: internal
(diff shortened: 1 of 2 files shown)


ilaner (changes requested):
Good job. 
Please remove the check in the typer command, so it will be done in one place only:
https://github.com/demisto/demisto-sdk/blob/5ab14809d6f04d40ccb0cfa089d9b2465b4a5d0f/demisto_sdk/commands/content_graph/commands/update.py#L217

ilaner (approved):
nice job!

PR #4044, opened 2024-02-14
[setup-env] - fix an issue with the client

GuyAfik:
## Description
* Fixed an issue where **setup-env** crashed when trying to upload an instance of an integration into xsoar.


ilaner (approved):
nice

PR #4045, opened 2024-02-14
[File Management] - fix an issue with git_util in tests 

GuyAfik:
## Description
Because the `git_util` is a class member for `File` object it gets read during tests only once, and that causes bugs as the same `git_util` object is shared between all the tests which can work on different repos.

that happens **only** in tests because there are multiple repositories created for each test using `TestSuite`.
When running in production we only have a single repository.


ilaner (approved):
nice

PR #4050, opened 2024-02-18
[pre-commit] - add validate-deleted-files script hook

GuyAfik:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9729

## Description
See how it works on https://github.com/demisto/content/pull/33004


demisto_sdk/scripts/validate_deleted_files.py
```diff
@@ -0,0 +1,106 @@
...
+        str(file_path)
+        for file_path in deleted_files_in_protected_dirs
+        if not is_file_allowed_to_be_deleted_by_file_type(file_path)
+    ]
+
+
+main = typer.Typer(pretty_exceptions_enable=False)
+
+
+@main.command()
```
ilaner, line 79: You dont need the decorator and the context i think. You can do this:
```
def main():
    print("Hello World")


if __name__ == "__main__":
    typer.run(main)
```

ilaner (approved):
nice job! see minor comment about using typer

demisto_sdk/scripts/validate_deleted_files.py
GuyAfik: 1. If changing it, it is not possible to test it with `runner.invoke` as it expects it to have the decorator
2. without defining the decorator, it doesn't seem to be possible to add a custom argument such as `--protected-dir`.

PR #4065, opened 2024-02-20
use uv in setup-env to speed up virtualenv creation

ilaner:


pyproject.toml
```diff
@@ -116,6 +116,7 @@ klara = { version = "^0.6.3", optional = true }
 
 pre-commit = "^3.5.0"
 lxml = "^5.0.0"
+uv = "^0.1.5"
```
dorschw, line 119:
```suggestion
uv = "^0.1.6"
```
🤡

demisto_sdk/commands/setup_env/setup_environment.py
```diff
@@ -491,23 +492,56 @@ def install_virtualenv(
...
-                check=True,
-            )
-            logger.info(f"Installed {req}")
-        except subprocess.CalledProcessError:
-            logger.warning(f"Could not install {req}, skipping...")
+    env = os.environ.copy()
+    env.update({"VIRTUAL_ENV": str(venv_path)})
+    requirements = [req for req in requirements if req]
+    try:
+        subprocess.run(
```
dorschw, line 499: for readability, let's split to methods: `install_all_using_uv`, `install_single_using_uv`, `install_all_using_pip`

demisto_sdk/commands/setup_env/setup_environment.py
```diff
@@ -460,6 +461,76 @@ def configure_vscode(
...
+            + requirements,
+            check=True,
+            env=env,
+        )
+        return True
+    except subprocess.CalledProcessError:
+        return False
+
+
+def install_one_by_one(requirements: List[str], env: dict, venv_path: Path) -> None:
```
dorschw, line 491: would it be clearer if we changed it to `install_single` and loop outside?

dorschw (commented):
Cool!

demisto_sdk/commands/setup_env/setup_environment.py
```diff
@@ -491,23 +562,15 @@ def install_virtualenv(
...
-                    "install",
-                    req,
-                ],
-                check=True,
-            )
-            logger.info(f"Installed {req}")
-        except subprocess.CalledProcessError:
-            logger.warning(f"Could not install {req}, skipping...")
+    env = os.environ.copy()
+    env.update({"VIRTUAL_ENV": str(venv_path)})
```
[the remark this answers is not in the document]
ilaner: we can have the same env for both `uv` and `pip`. I don't see a reason to seperate them

pyproject.toml
dorschw:
```suggestion
uv = "^0.1.24"
```

dorschw (approved):

PR #4066, opened 2024-02-20
[clients] - allow checking if server is healthy

GuyAfik:
## Description
* Allow checking if server is healthy
* Add PollTimeout exception when polling fails to get incident / playbook state.

### DIFF
--- a/demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
+++ b/demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
@@ -1283,7 +1283,7 @@ class XsoarClient:
 
         while elapsed_time < timeout:
             playbook_state_raw_response = self.get_playbook_state(incident_id)
-            logger.debug(f"playbook state: {playbook_state_raw_response}")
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
```diff
@@ -735,7 +737,11 @@ def poll_incident_state(
                 time.sleep(interval)
                 elapsed_time = int(time.time() - start_time)
 
-        raise RuntimeError(f"status of incident {incident_name} is {incident_status}")
+        raise PollTimeout(
```
JudahSchwartz, line 740: why not just raise the connection error? Think this will be more helpful for people trying to debug the real issue. this should be a log

GuyAfik: Connection Error isn't the case here, it just means the incident hasn't reached the required state.

JudahSchwartz (approved):

ilaner (approved):
nice!

PR #4081, opened 2024-02-25
[Files] - fixes and improvements

GuyAfik:
## Description
* Error improvements for the `files` package
* Fixes an issue where if failing to load a text based file when reading file directly from memory and not from path, the SDK crashes.


demisto_sdk/commands/common/files/structured_file.py
```diff
@@ -30,7 +30,13 @@ def as_default(cls, **kwargs):
         return instance
 
     def load(self, file_content: bytes) -> Any:
-        return self.handler.load(StringIO(super().load(file_content)))
+        file_content_as_text = super().load(file_content)
```
ilaner, line 33: what happens if this fails?

demisto_sdk/commands/common/files/ini_file.py
```diff
@@ -16,7 +17,14 @@ def is_model_type_by_path(cls, path: Path) -> bool:
 
     def load(self, file_content: bytes) -> ConfigParser:
         config_parser = ConfigParser(allow_no_value=True)
-        config_parser.read_string(super().load(file_content))
+        file_content_as_text = super().load(file_content)
```
ilaner, line 20: what happens if this fails?

ilaner (approved):
nice job, see comments

demisto_sdk/commands/common/files/ini_file.py
GuyAfik: will raise `FileLoadError` which is what we want, it means we couldn't load the object into the requested object.

PR #4088, opened 2024-02-26
[SDK] - do not fail if repo isn't git

GuyAfik:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9790

## Description
Fixes issue where SDK would crash when running on repositories which are not git.

### DIFF
--- a/demisto_sdk/commands/common/files/file.py
+++ b/demisto_sdk/commands/common/files/file.py
@@ -205,12 +205,10 @@ class File(ABC):
             cls.read_from_local_path.cache_clear()
 
         if not path.is_absolute():
+            logger.debug(f"path {path} is not absolute path")
             try:
                 git_util = GitUtil.from_content_path()
                 working_dir = git_util.repo.working_dir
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/files/file.py
```diff
@@ -33,14 +34,22 @@
...
-        if current_path != Path(cls._git_util.repo.working_dir):
-            cls._git_util = GitUtil.from_content_path(current_path)
-        return cls._git_util
+    def git_util(cls) -> Union[GitUtil, None]:
+        if cls._git_util:
+            current_path = Path.cwd()
+            if current_path != Path(cls._git_util.repo.working_dir):
+                cls._git_util = GitUtil.from_content_path(current_path)
+            return cls._git_util
+        return None
```
ilaner, line 52: if it is None try to create a new GitUtil, maybe something was changed

ilaner (approved):

PR #4094, opened 2024-02-28
[setup-env] - get demisto-mock from content github if not exist locally

GuyAfik:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9830

## Description
* Fixed an issue where setup-env would fail when trying to copy demistomock.py locally.


demisto_sdk/commands/setup_env/setup_environment.py
```diff
@@ -85,13 +90,31 @@ def get_docker_python_path(docker_prefix: str) -> List[str]:
             docker_python_path.append(
                 f"{docker_prefix}/{path.relative_to(CONTENT_PATH.absolute())}"
             )
+
+    if not COMMON_SERVER_PYTHON_PATH.exists():
+        try:
+            common_server_python_contnet = TextFile.read_from_github_api(
```
dorschw, line 96: Does this support the offline mode env var?

GuyAfik: nope, need to add, good catch

GuyAfik: will be done in https://github.com/demisto/demisto-sdk/pull/4098

GuyAfik: fixed in https://github.com/demisto/demisto-sdk/pull/4098

demisto_sdk/commands/setup_env/setup_environment.py
```diff
@@ -85,13 +91,35 @@ def get_docker_python_path(docker_prefix: str) -> List[str]:
             docker_python_path.append(
                 f"{docker_prefix}/{path.relative_to(CONTENT_PATH.absolute())}"
             )
+
     if (
-        f"{docker_prefix}/Packs/Base/Scripts/CommonServerPython"
+        f"{docker_prefix}/Packs/Base/Scripts/CommonServerPython"  # CommonServerPython does not exist locally
```
ilaner, line 96: why it means it don't exist locally?

GuyAfik: if line 92 produces ValueError silently, then the `CommonServerPython` is likely does not exist.

ilaner (approved):

PR #4119, opened 2024-03-07
change to months

JudahSchwartz:
primary focus on this validation should no longer be that the image youre using is the latest, it should be that your image isnt outdated. Outdated is > 3 months 
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9920

### DIFF
--- a/.changelog/4119.yml
+++ b/.changelog/4119.yml
@@ -1,4 +1,4 @@
 changes:
-- description: Updated the **validate** command to not fail when an image is younger than 3 months instead of 3 days.
+- description: Updated the *DO106* error code in the **validate** command to not fail when a docker image is younger than 3 months instead of 3 days.
   type: feature
 pr_number: 4119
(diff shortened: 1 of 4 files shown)


GuyAfik (approved):
@JudahSchwartz looks good, please only add a changelog

.changelog/4119.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Updated the **validate** command to not fail when an image is younger than 3 months instead of 3 days.
```
GuyAfik, line 2:
```suggestion
- description: Updated the *DO106* error code in the **validate** command to not fail when a docker image is younger than 3 months instead of 3 days.
```

.changelog/4119.yml
```diff
@@ -0,0 +1,4 @@
...
+  type: breaking
```
GuyAfik, line 3:
```suggestion
  type: feature
```

GuyAfik (approved):

PR #4130, opened 2024-03-12
[pre-commit] - split hooks with multi-processing

GuyAfik:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9269

## Description
**Before:**

* We ran all the hooks with multi-processing at the same time except the hooks which require docker execution before running.
* We created a separate file for each docker hook / split hook.

**With the new implementation:**
* All the hooks will be on the **main** file (only one file will be created)
* Hooks which are split will run with multi-processing
* The execution of the hooks will be in the same order where they are defined in the template
* There is no need for the **needs** parameter anymore as we keep the hooks execution ordered.
* Each `prepare-hook` implementation requires now to return all the created hook IDs for the original hooks.
* Added support for the *parallel* key to determine whether a split hook should run in parallel or sequentially.


demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -186,40 +186,63 @@ def run(
...
         (
-            docker_hooks,
-            no_docker_hooks,
-        ) = pre_commit_context._get_docker_and_no_docker_hooks(local_repo)
-        local_repo["hooks"] = no_docker_hooks
+            split_hooks,
+            non_split_hooks,
+        ) = pre_commit_context._get_split_and_non_split_hooks(local_repo)
+        logger.info(f'Split hooks: {[hook["id"] for hook in split_hooks]}')
+        logger.info(f'Non-split hooks: {[hook["id"] for hook in non_split_hooks]}')
```
GuyAfik, line 193:
```suggestion
        logger.debug(f'Split hooks: {[hook["id"] for hook in split_hooks]}')
        logger.debug(f'Non-split hooks: {[hook["id"] for hook in non_split_hooks]}')
```

will be changed before merging, now its used for debugging issues

demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -183,57 +198,54 @@ def run(
...
-            # run hooks that needs docker after all the docker hooks finished
-            pre_commit_context._update_hooks_needs_docker(full_hooks_need_docker)
-            path = PRECOMMIT_CONFIG_MAIN_PATH.with_name(
-                f"{PRECOMMIT_CONFIG_MAIN_PATH.stem}-needs.yaml"
-            )
-            write_dict(path, pre_commit_context.precommit_template)
-            process_needs_docker = PreCommitRunner._run_pre_commit_process(
-                path, precommit_env, verbose=verbose
+
+        stdout = (
```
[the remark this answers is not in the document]
GuyAfik: good idea, will fix

demisto_sdk/commands/pre_commit/README.md
```diff
@@ -54,6 +54,14 @@ In this example, both hooks will be skipped.
   needs: ["sample-hook"]
 ```
 
+## parallel key
+The parallel indicates whether a hook should run in parallel, by default hooks such as mypy, ruff and docker produce multiple hooks which run in parallel.
```
JudahSchwartz, line 58:
```suggestion
The parallel key indicates whether a hook should run in parallel, by default hooks such as mypy, ruff and docker produce multiple hooks which run in parallel.
```

JudahSchwartz (approved):

PR #4135, opened 2024-03-14
[validate-deleted-files] - update typer main signature

GuyAfik:


demisto_sdk/scripts/validate_deleted_files.py
```diff
@@ -79,16 +80,9 @@ def get_forbidden_deleted_files(protected_dirs: Set[str]) -> List[str]:
 
 @main.command()
 def validate_forbidden_deleted_files(
-    ctx: typer.Context,
-    protected_dirs: List[str] = typer.Option(
-        [],
-        "--protected-dir",
-        help="a protected dir to to disallow deleting files from it or from it sub-directories",
```
JudahSchwartz, line 86: where is this documentation defined now?

JudahSchwartz (approved):

demisto_sdk/scripts/validate_deleted_files.py
GuyAfik: using it in the help section

PR #4154, opened 2024-03-20
Graph - reuse parse content dto, skip empty packs

ilaner:
This fixes the following:

* This will reuse the `ContentDTO.from_path`, so if the content was parsed in the sdk run, it will not parse
* This will skip empty packs

fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9717


demisto_sdk/commands/content_graph/parsers/repository.py
```diff
@@ -42,14 +45,25 @@ def parse(
...
         except Exception:
             logger.error(traceback.format_exc())
             raise
 
+    @staticmethod
+    def parse_pack(pack_path: Path) -> Optional[PackParser]:
+        try:
+            return PackParser(pack_path)
+        except NotAContentItemException:
+            logger.warning(f"Pack {pack_path.name} is not a valid pack. Skipping")
```
GuyAfik, line 64:
```suggestion
            logger.error(f"Pack {pack_path.name} is not a valid pack. Skipping")
```

maybe its better to log an error in case we can't parse a pack?

demisto_sdk/commands/content_graph/commands/update.py
```diff
@@ -72,7 +72,10 @@ def update_content_graph(
         dependencies (bool): Whether to create the dependencies.
         output_path (Path): The path to export the graph zip to.
     """
-    if os.getenv("DEMISTO_SDK_GRAPH_FORCE_CREATE"):
+    logger.debug(
+        f'DEMISTO_SDK_GRAPH_FORCE_CREATE = {os.getenv("DEMISTO_SDK_GRAPH_FORCE_CREATE")}'
+    )
+    if os.getenv("DEMISTO_SDK_GRAPH_FORCE_CREATE") == "true":
```
GuyAfik, line 78: put inside a variable and reuse it also in the line 76

demisto_sdk/commands/content_graph/objects/repository.py
```diff
@@ -17,25 +18,41 @@
...
+        if not packs_to_parse
+        else len(packs_to_parse),
+        unit="packs",
+        desc="Parsing packs",
+        position=0,
+        leave=True,
+    ) as progress_bar:
+        repo_parser.parse(progress_bar=progress_bar)
+    content_dto = ContentDTO.from_orm(repo_parser)
+    return content_dto
```
GuyAfik, line 41:
```suggestion
    return ContentDTO.from_orm(repo_parser)
```

demisto_sdk/commands/content_graph/objects/repository.py
```diff
@@ -17,25 +18,41 @@
 USE_MULTIPROCESSING = False  # toggle this for better debugging
 
 
+@lru_cache
+def from_path(path: Path = CONTENT_PATH, packs_to_parse: Optional[Tuple[str]] = None):
```
GuyAfik, line 22: why isn't it a class method of `ContentDTO`?

GuyAfik (approved):

demisto_sdk/commands/content_graph/objects/repository.py
ilaner: I documented, for some reason I get cache miss whin it's in the class.

PR #4166, opened 2024-03-24
[setup env] fix pack upload

ilaner:
We need to reparse the pack again since the pack doesn't contain all the content items to upload.
in addition fixed an issue where debugging locally failed

fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10024?filter=-1


GuyAfik (approved):
🥇

PR #4170, opened 2024-03-24
XT100 new path vaildation

anas-yousef:
## Related Issues
fixes:

## Description
XT100 new path vaildation

### DIFF
--- a/demisto_sdk/scripts/validate_content_path.py
+++ b/demisto_sdk/scripts/validate_content_path.py
@@ -176,6 +176,10 @@ class InvalidCommandExampleFile(InvalidPathException):
     message = "This file's name must be command_examples"
 
 
+class InvalidXDRCTemplatesFileName(InvalidPathException):
+    message = "Files in the XDRC templates directory must be titled exactly as the pack, e.g. `myPack.yml`"
+
+
 class ExemptedPath(Exception, ABC):
     message: ClassVar[str]
 
@@ -260,11 +264,15 @@ def _validate(path: Path) -> None:
         ):
             raise InvalidLayoutFileName
 
-    if depth == 2 and first_level_folder in {
-        ContentType.INTEGRATION.as_folder,
-        ContentType.SCRIPT.as_folder,
-    }:
-        _validate_integration_script_file(path, parts_after_packs)
+    if depth == 2:
+        if first_level_folder in {
+            ContentType.INTEGRATION.as_folder,
+            ContentType.SCRIPT.as_folder,
+        }:
+            _validate_integration_script_file(path, parts_after_packs)
+        elif first_level_folder in {ContentType.XDRC_TEMPLATE.as_folder}:
+            if path.stem != path.parent.name:
+                raise InvalidXDRCTemplatesFileName()
 
 
 def _validate_integration_script_file(path: Path, parts_after_packs: Sequence[str]):
--- a/demisto_sdk/tests/validate_content_path_test.py
+++ b/demisto_sdk/tests/validate_content_path_test.py
@@ -9,6 +9,7 @@ from demisto_sdk.commands.common.constants import (
     PACKS_FOLDER,
     PLAYBOOKS_DIR,
     SCRIPTS_DIR,
+    XDRC_TEMPLATE_DIR,
 )
 from demisto_sdk.scripts.validate_content_path import (
     DEPTH_ONE_FOLDERS,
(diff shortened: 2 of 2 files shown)


demisto_sdk/tests/validate_content_path_test.py
```diff
@@ -31,6 +33,31 @@
...
+
+
+@pytest.mark.parametrize(
+    "templates_folder, template_name",
+    (
+        ("template_foo", "template_foo_test.json"),
+        ("template_foo", "template_foo_test.yml"),
+    ),
+)
+def test_xdrc_template_file_invalid(templates_folder: str, template_name: str):
```
dorschw, line 54: Let's test a capital letter, trailing underscore, just to make it a tiny bit clearer what this test is for (took me a minute to figure out the `_test` was the punch here)

demisto_sdk/tests/validate_content_path_test.py
```diff
@@ -31,6 +33,31 @@
...
 
+@pytest.mark.parametrize(
+    "templates_folder, template_name",
+    (
+        ("template_foo", "template_foo.json"),
+        ("template_foo", "template_foo.yml"),
+    ),
+)
+def test_xdrc_template_file_valid(templates_folder: str, template_name: str):
+    _validate(DUMMY_PACK_PATH / XDRC_TEMPLATE_DIR / templates_folder / template_name)
```
dorschw, line 44: Both tests can be simplified, instead of passing `template_foo` four times, just set it as a variable 

```suggestion
@pytest.mark.parametrize(
    "suffix",
    ( ".json", ".yml"),
)
def test_xdrc_template_file_valid(templates_folder: str, template_name: str):
    folder = "foo"
    _validate(DUMMY_PACK_PATH / XDRC_TEMPLATE_DIR / folder / f"{folder}.{suffix}")
```

demisto_sdk/tests/validate_content_path_test.py
dorschw, line 54: simplify similar to the test above

demisto_sdk/scripts/validate_content_path.py
```diff
@@ -260,11 +264,15 @@ def _validate(path: Path) -> None:
...
-    }:
-        _validate_integration_script_file(path, parts_after_packs)
+    if depth == 2:
+        if first_level_folder in {
+            ContentType.INTEGRATION.as_folder,
+            ContentType.SCRIPT.as_folder,
+        }:
+            _validate_integration_script_file(path, parts_after_packs)
+        elif first_level_folder in {ContentType.XDRC_TEMPLATE.as_folder}:
+            if path.stem != path.parent.name:
```
dorschw, line 274: can we check the suffix here too? while we're there

anas-yousef: I believe all the files under the **XDRCTemplates** folder should be either json or YAML. Should we check if the suffix is either one?

dorschw: > I believe all the files under the **XDRCTemplates** folder should be either json or YAML. Should we check if the suffix is either one?

yeah, like we do for integrations.

demisto_sdk/tests/validate_content_path_test.py
```diff
@@ -33,6 +35,33 @@
...
+    (
+        ("MyXDRCTemplate_test", "json"),
+        ("MyXDRCTemplate_test", "yml"),
+        ("MyXDRCTemplate", "py"),
+    ),
+)
+def test_xdrc_template_file_invalid(file: str, suffix: str):
+    """
+    The first two tests will fail because the file name is incorrect
+    The last one will fail because the file extension is incorrect (should be .json or .yml)
```
dorschw, line 58: please add this information to the test cases using `pytest.param` and `id='bad name, good suffix'` etc

demisto_sdk/tests/validate_content_path_test.py
```diff
@@ -33,6 +35,29 @@
...
+def test_xdrc_template_file_valid(suffix: str):
+    folder = "foo"
+    _validate(DUMMY_PACK_PATH / XDRC_TEMPLATE_DIR / folder / f"{folder}.{suffix}")
+
+
+@pytest.mark.parametrize(
+    "file, suffix",
+    (
+        pytest.param("MyXDRCTemplate_test", "json", id="bad name, good suffix"),
+        pytest.param("MyXDRCTemplate_test", "yml", id="bad name, good suffix"),
```
dorschw, line 51:
```suggestion
        pytest.param("MyXDRCTemplate_test", "json", id="bad name, good suffix - json"),
        pytest.param("MyXDRCTemplate_test", "yml", id="bad name, good suffix - yml"),
```

dorschw (approved):

PR #4171, opened 2024-03-24
Improve path validation CI

dorschw:
1. only run if the validation/its test is modified
2. fail `validate-all` if there are failured
3. use `fetch-depth=1` for content

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -335,7 +335,9 @@ jobs:
         id: validate-content-path-changed
         uses: tj-actions/changed-files@v43
         with:
-          files: demisto-sdk/demisto_sdk/scripts/validate_content_path.py
+          files: |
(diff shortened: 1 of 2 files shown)


.github/workflows/on-push.yml
```diff
@@ -331,19 +331,30 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Check whether validate_content_path changed
+        id: files-changed
+        uses: tj-actions/changed-files@v43
```
GuyAfik, line 336:
```suggestion
        uses: tj-actions/changed-files@v4
```

43?

GuyAfik (approved):
There is no way to skip the entire job if those files had not been changed rather than skipping each step separately?

.github/workflows/on-push.yml
dorschw: [43](https://github.com/tj-actions/changed-files/releases/tag/v43.0.1)

PR #4175, opened 2024-03-25
[Pre-Commit] Handle API Modules

ilaner:
When Api Module is changed, it will trigger all the used integrations by the module to lint.
In addition, it will add the API modules to the used integration files, so they will be linted by the environment of the integration that uses the module.

fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-7447


demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -347,19 +356,39 @@ def group_by_language(
...
     ):
         with multiprocessing.Pool(processes=cpu_count()) as pool:
-            integrations_scripts.extend(
-                pool.map(BaseContent.from_path, integration_script_paths)
-            )
+            chunk = pool.map(BaseContent.from_path, integration_script_paths)
+            for i in chunk:
+                if not i or not isinstance(i, IntegrationScript):
+                    continue
+                integrations_scripts.add(i)
```
GuyAfik, line 368:
```suggestion
            for content_item in content_items:
                if not content_item or not isinstance(content_item, IntegrationScript):
                    continue
                # content-item is a script/integration
                integrations_scripts.add(content_item)
```

demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -347,19 +356,39 @@ def group_by_language(
...
-    integrations_scripts = []
+    integrations_scripts: Set[IntegrationScript] = set()
     for integration_script_paths in more_itertools.chunked_even(
         integrations_scripts_mapping.keys(), INTEGRATIONS_BATCH
     ):
         with multiprocessing.Pool(processes=cpu_count()) as pool:
-            integrations_scripts.extend(
-                pool.map(BaseContent.from_path, integration_script_paths)
-            )
+            chunk = pool.map(BaseContent.from_path, integration_script_paths)
```
GuyAfik, line 364:
```suggestion
            content_items = pool.map(BaseContent.from_path, integration_script_paths)
```

demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -347,19 +356,39 @@ def group_by_language(
...
+            continue
+    if api_modules:
+        update_content_graph(graph)
+        api_modules: List[Script] = graph.search(
+            object_id=[api_module.object_id for api_module in api_modules]
+        )  # type: ignore[no-redef]
+        for api_module in api_modules:
+            for imported_by in api_module.imported_by:
+                # we need to add the api module for each integration that uses it, so it will execute the api module check
+                assert isinstance(imported_by, Integration)
```
GuyAfik, line 384:
```suggestion
               if not isinstance(imported_by, Integration):
                   raise RunTimeError("....")
```
or

```suggestion
                assert isinstance(imported_by, Integration), f'add-error....'
```

Would be more informative then just an assert

demisto_sdk/commands/pre_commit/hooks/docker.py
```diff
@@ -342,7 +342,7 @@ def generate_hooks(
         docker_extra_args = self._get_property("pass_docker_extra_args", "")
         new_hook[
             "entry"
-        ] = f'--entrypoint {new_hook.get("entry")} {docker_extra_args} {get_environment_flag(env)} {"--quiet" if quiet else ""} {dev_image}'
+        ] = f'--entrypoint {new_hook.get("entry")} {docker_extra_args} --user demisto {get_environment_flag(env)} {"--quiet" if quiet else ""} {dev_image}'
```
GuyAfik, line 345:
```suggestion
        ] = f'--entrypoint {new_hook.get("entry")} {docker_extra_args} --user {USER_DEMISTO} {get_environment_flag(env)} {"--quiet" if quiet else ""} {dev_image}'
```

make it constant

demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -321,8 +328,10 @@ def group_by_language(
     Returns:
         Dict[str, set]: The files grouped by their python version, and a set of excluded paths
     """
+    graph = ContentGraphInterface()
```
GuyAfik, line 331: define it only in line 376, no? is there a reason to init the object before?

GuyAfik (commented):
Amazing job!

If possible lets add a test for it

demisto_sdk/commands/pre_commit/pre_commit_command.py
ilaner: this is for type checking, i changed the return value for `imported_by` function as it has to be Script no matter what

GuyAfik: but won't it crash the code if type isn't as expected? do we want it to crash?

ilaner (commented):
> but won't it crash the code if type isn't as expected? do we want it to crash? 

it has to be a `Script`. If it is not a script there is a bug in the graph, and we want it to crash

ilaner (commented):
added a unit test

GuyAfik (approved):
Good job

GuyAfik:
add changelog

PR #4179, opened 2024-03-26
fix IN101 rationale

dorschw:


Utils/github_workflow_scripts/slack_notifier_master/slack_notifier.py
```diff
@@ -191,5 +191,34 @@ def slack_notifier(
         logger.info("There are not any failed jobs to report to slack")
 
 
+@main.command("validate-doc-gen")
+def validate_doc_gen(
```
YuvHayun, line 195: should it be here?

YuvHayun (approved):

Utils/github_workflow_scripts/slack_notifier_master/slack_notifier.py
dorschw: removed

PR #4183, opened 2024-03-26
[slack-notifier] - avoid sending duplicated tests

GuyAfik:
## Description
* Fixed an issue where slack-notifier sent duplicated tests in case the same tests failed on master for every python version.

If the same test failed for each python version, it will be reported into slack notifier just once.


ilaner (approved):
nice

PR #4192, opened 2024-03-28
[files] - Refactor read_from_github method

GuyAfik:
## Description
Fixed an issue where in the new validate we try to get deprecated dockers list from `dockerfiles` repo, but instead the code tries to access the content repo because of a bad usage of `GitContentConfig` class.

Refactor `read_from_github_api` method to build the URL explicitly and not to use `GitContentConfig` class.


ilaner (approved):
nice job!

PR #4209, opened 2024-04-07
Support Python 3.12

dorschw:


demisto_sdk/commands/common/docker/dockerhub_client.py
```diff
@@ -107,13 +107,9 @@ def get_token(
             )
             if (
                 _error.response is not None
-                and (
-                    _error.response.status_code
-                    in (requests.codes.unauthorized, requests.codes.too_many_requests)
```
dorschw, line 112: not sure why this was done, revert it unless you do understand

PR #4219, opened 2024-04-09
Fix setup env test collection

ilaner:
fixes the following:
* This will create a seperate .env for the docker run. We have to use it because the path are different between the host and container
* Change to use `debugpy` in luanch since `python` debugging is deprecated.

fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10277


GuyAfik (approved):
Good Job!

PR #4235, opened 2024-04-16
exclude Trigger from query

thefrieddan1:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10304

## Description
Remove the Trigger content item type from the GR101 from_version validation in old validator


ilaner (approved):
approved

PR #4250, opened 2024-04-30
remove `poetry-export`, replace `poetry-lock` hook with `poetry-check --lock`

dorschw:
- following https://github.com/demisto/demisto-sdk/pull/4225, `poetry-export` is no longer necessary.
- `poetry-check` hook with with `--lock` has been deprecated. Replacing with `poetry-lock --check`

related (similar): https://github.com/demisto/content/pull/34092


ilaner (approved):
nice

PR #4273, opened 2024-05-12
add pb validations

merit-maita:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10535


demisto_sdk/commands/validate/validators/PB_validators/PB101_is_playbook_has_unreachable_condition.py
```diff
@@ -0,0 +1,88 @@
...
+
+                # ADD all replyOptions to unhandled_reply_options (UPPER)
+                unhandled_reply_options = set(
+                    map(str.upper, task.get("message", {}).get("replyOptions", []))
+                )
+
+                # Rename the keys in dictionary to upper case
+                next_tasks_upper = {k.upper(): v for k, v in next_tasks.items()}
+
+                mapper_dict = {
```
thefrieddan1, line 67: Seems like we can define this once outside the nested for loops. as a CONST

demisto_sdk/commands/validate/validators/PB_validators/PB101_is_playbook_has_unreachable_condition.py
```diff
@@ -0,0 +1,88 @@
...
+    def is_valid(self, content_items: Iterable[ContentTypes]) -> List[ValidationResult]:
+        """Checks whether a builtin conditional task branches are handled properly
+        Args:
+            - content_items (Iterable[ContentTypes]): The content items to check.
+        Return:
+            - List[ValidationResult]. List of ValidationResults objects.
+        """
+        results: List[ValidationResult] = []
+        for content_item in content_items:
+            invalid_tasks = self.invalid_tasks(content_item)
```
thefrieddan1, line 35: There are allot of nested for loops. is_valid has 2 from which one is calling invalid_tasks which have 3 more nested loops.
the result append is not dependent on invalid_tasks.
append all invalid_tasks in a single for loop.
Then append all items from invalid_tasks to resutls not in a nested for loop.

demisto_sdk/commands/validate/validators/PB_validators/PB101_is_playbook_has_unreachable_condition.py
```diff
@@ -0,0 +1,88 @@
...
+
+                mapper_dict = {
+                    "YES": ["YES", "TRUE POSITIVE"],
+                    "TRUE POSITIVE": ["YES", "TRUE POSITIVE"],
+                    "NO": ["NO", "FALSE POSITIVE"],
+                    "FALSE POSITIVE": ["NO", "FALSE POSITIVE"],
+                }
+
+                # Remove all nexttasks from unhandled_reply_options (UPPER)
+                for next_task_branch, next_task_id in next_tasks_upper.items():
```
thefrieddan1, line 75: same here with nested loops gather all next_tasks_upper and iterate over it without a nested loop.

demisto_sdk/commands/validate/validators/PB_validators/PB123_is_conditional_task_has_unhandled_reply_options.py
```diff
@@ -0,0 +1,97 @@
...
+        Args:
+            - content_items (Iterable[ContentTypes]): The content items to check.
+        Return:
+            - List[ValidationResult]. List of ValidationResults objects.
+        """
+        results: List[ValidationResult] = []
+        for content_item in content_items:
+            invalid_tasks = self.unhandled_conditions(content_item)
+
+            for task_id, unhandled_options in invalid_tasks.items():
```
thefrieddan1, line 36: break the nested for

demisto_sdk/commands/validate/validators/PB_validators/PB123_is_conditional_task_has_unhandled_reply_options.py
```diff
@@ -0,0 +1,97 @@
...
+
+                # ADD all replyOptions to unhandled_reply_options (UPPER)
+                unhandled_reply_options = set(
+                    map(str.upper, task.get("message", {}).get("replyOptions", []))
+                )
+
+                # Rename the keys in dictionary to upper case
+                next_tasks_upper = {k.upper(): v for k, v in next_tasks.items()}
+
+                mapper_dict = {
```
thefrieddan1, line 68: move to CONST

demisto_sdk/commands/validate/validators/PB_validators/PB123_is_conditional_task_has_unhandled_reply_options.py
```diff
@@ -0,0 +1,97 @@
...
+                    "YES": ["YES", "TRUE POSITIVE"],
+                    "TRUE POSITIVE": ["YES", "TRUE POSITIVE"],
+                    "NO": ["NO", "FALSE POSITIVE"],
+                    "FALSE POSITIVE": ["NO", "FALSE POSITIVE"],
+                }
+
+                # Remove all nexttasks from unhandled_reply_options (UPPER)
+                for next_task_branch, next_task_id in next_tasks_upper.items():
```
thefrieddan1, line 76: same here break the nested for

thefrieddan1 (changes requested):
Good job.
Please see comments.
If for loops can run sequentially and are not dependent on each other it is better to break them to separate runs instead of nested for loops

demisto_sdk/commands/validate/validators/PB_validators/PB123_is_conditional_task_has_unhandled_reply_options.py
thefrieddan1: Seems like there is a dependency down the road which prevents breaking the nested for loop

demisto_sdk/commands/validate/validators/PB_validators/PB123_is_conditional_task_has_unhandled_reply_options.py
thefrieddan1: Seems like there is a dependency down the road which prevents breaking the nested for loop

demisto_sdk/commands/validate/validators/PB_validators/PB101_is_playbook_has_unreachable_condition.py
thefrieddan1: Seems like there is a dependency down the road which prevents breaking the nested for loop

demisto_sdk/commands/validate/validators/PB_validators/PB101_is_playbook_has_unreachable_condition.py
thefrieddan1: Seems like there is a dependency down the road which prevents breaking the nested for loop

thefrieddan1 (approved):

PR #4287, opened 2024-05-20
Added fix for auto update docker image field

anas-yousef:
## Related Issues
fixes:https://jira-dc.paloaltonetworks.com/browse/CIAC-10899

## Description
To take into consideration if the YAML contains a boolean or string value


demisto_sdk/commands/content_graph/parsers/integration_script.py
```diff
@@ -91,4 +91,4 @@ def subtype(self):
     def auto_update_docker_image(self):
         return (
             get_value(self.yml_data, "autoUpdateDockerImage", "")
-        ).lower() != "false"
+        ).lower() not in ("false", False)
```
JudahSchwartz, line 94: lets add a test that a yml provided with no input at all defaults to `auto_update_docker_image = True`

JudahSchwartz (approved):

PR #4292, opened 2024-05-22
Validation ds102

moishce:
## Related Issues
https://jira-dc.paloaltonetworks.com/browse/CIAC-10562
## Description
Check if beta disclaimer exists in yml detailed description


moishce:
@ShahafBenYakir, This validation did not run at all in the old validate
And we haven't added so far in any beta integration a disclaimer in the yml file, do we need it?

moishce:
@ShahafBenYakir I'm deleting this validation because we don't require the beta disclaimer to be in the yml file,
because it already appears in the display name that it is a beta
and the description file already has the disclaimer

PR #4294, opened 2024-05-23
Fix latest docker image logic

DinaMeylakh:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10596

## Description
Fix the calculation of the latest docker image. Instead of looking at the whole tag as a version, we will now decide on the latest version via the last digits of the image. Those represent the build number where the image was created and should always increase.


.changelog/4294.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Fixed the logic for retrieving the latest Docker image tag in both the old validate and the new.
```
YuvHayun, line 2: Fixed an issue where the latest docker image validator didn't retrieve the latest docker image version as a fix suggestion.

YuvHayun (approved):
Looking great!

ShahafBenYakir (approved):

PR #4309, opened 2024-05-28
[demisto-sdk] Add skipped paths to old_validate_manager

samuelFain:
## Related Issues
relates: [CIAC-10490](https://jira-dc.paloaltonetworks.com/browse/CIAC-10490)

## Description
Add SKIPPED_PATHS constant, which is used to exclude specific paths when using `--include-untracked` flag when running `demisto-sdk validate`.


dorschw (changes requested):
offline discussion

PR #4314, opened 2024-05-30
[pre-commit] Ignore docker_native_image_config.json in simple repo

mmhw:
## Description
Ignore `docker_native_image_config.json` in simple repo.
To avoid the [following error](https://github.com/demisto/demisto-sdk/actions/runs/9299023762/job/25592089151?pr=4314#step:9:72):
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/demisto-sdk/demisto-sdk/content/Tests/docker_native_image_config.json'
```


.github/workflows/on-push.yml
```diff
@@ -283,6 +283,8 @@ jobs:
           source $(poetry env info --path)/bin/activate
           cd content
           rm -rf .pre-commit-config_template.yaml
+          rm -f Tests/docker_native_image_config.json
+          touch .private-repo-settings
```
JudahSchwartz, line 287: why is this file needed?

JudahSchwartz (changes requested):
doesnt seem like the right solution to me. We shouldnt distinguish based on internal vs external repo. If an external repo wants to run on the native image they should be allowed to. Lets instead lazy load this tag, that we should only need it if native image is declated in the precommit file

.github/workflows/on-push.yml
mmhw: Because I thought to use the `is_external_repository` function and it checks against this file.

mmhw:
> doesnt seem like the right solution to me. We shouldnt distinguish based on internal vs external repo. If an external repo wants to run on the native image they should be allowed to. Lets instead lazy load this tag, that we should only need it if native image is declated in the precommit file

@JudahSchwartz I have made the change, if you could please look again.

JudahSchwartz (approved):
amazing

PR #4328, opened 2024-06-06
fix download --force --run-format merging problems

DinaMeylakh:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-9685

## Description
When running, we first call a get_dict_from_file function to do the merging with the local file. When running format after we use the get_dict_from_file function a second time. By default, it uses the cache and therefore returned the old (pre-merge) file again and ran format on it instead of the new file and saved it again. This resulted in the merger results not being saved. Passing clear_cache=True resolved this problem.


DinaMeylakh:
TDD: failing UT in commit [ba210aa](https://github.com/demisto/demisto-sdk/pull/4328/commits/ba210aa4c526ba989c038160ffcb096d7880fdad) in this [build job](https://github.com/demisto/demisto-sdk/actions/runs/9402653760/job/25897271551?pr=4328)

ShahafBenYakir (approved):

PR #4342, opened 2024-06-13
Add PA114 to new validate

DinaMeylakh:
## Description
Added PA114 pack metadata version bump validation.
Added mechanism to check pack metadata files even if only relevant content was changed and the pack metadata files did not.

### Design Considerations
Unlike other validations, PA114 needs to collect the pack metadata along with its remote master version even if it was not modified and bump according to other, non-pack metadata file changes.

As of so, it has been decided that:
1. We always collect the pack metadata of the modified content items, along with its remote master version.
2. We trigger the PA114 check on all the collected content items.
3. The logic of when the validation runs is kept inside the validation file rather than in the collection of the files step.

This was done to:
* Keep the logic of when the validation runs inside the specific validation file rather than in the collection of the files step.
* Keep the infrastructure unified and simple: Since not a lot of other validations need to cross reference files, it has been decided no major change/enhancement should be made to the initializer/validate_manager files. If this changes in the future, such an enhancement should be considered.


demisto_sdk/commands/validate/initializer.py
```diff
@@ -504,7 +508,7 @@ def git_paths_to_basecontent_set(
                 if obj:
                     obj.git_status = git_status
                     # Check if the file exists
-                    if git_status in (GitStatuses.MODIFIED, GitStatuses.RENAMED):
+                    if git_status in (GitStatuses.MODIFIED, GitStatuses.RENAMED, None):
```
YuvHayun, line 511: Let's do that for None only in cases where the file is a pack_metadata.json file.

demisto_sdk/commands/validate/initializer.py
```diff
@@ -571,13 +575,23 @@ def get_items_status(
                     statuses_dict[path] = git_status
             elif PACKS_PACK_META_FILE_NAME in path_str:
                 statuses_dict[path] = git_status
-            elif self.is_pack_item(path_str):
+            if (
+                self.is_pack_item(path_str)
+                or find_type(path_str) not in SKIP_RELEASE_NOTES_FOR_TYPES
+            ):
```
YuvHayun, line 581: Maybe we should add this condition at the end of the for loop after the else statement?
The idea of this else statement is to catch items that doesn't have multiple files (like mappers, incident fields, etc...) so I think we will miss them in the current implementation.

demisto_sdk/commands/validate/initializer.py
```diff
@@ -571,13 +575,23 @@ def get_items_status(
...
                     statuses_dict[metadata_path] = None
             else:
                 statuses_dict[path] = git_status
 
+        all_collected_files = set(statuses_dict.keys())
+        git_only_files = set(file_by_status_dict.keys())
+        related_files = all_collected_files.difference(git_only_files)
+        if related_files:
+            logger.info("Running on related files:")
+            logger.info(f"{[str(path) for path in related_files]}")
```
YuvHayun, line 593: If I'm not wrong - the related files, in this case, is all the files that are not being parsed into graph objects, is this log really helpful?

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
```diff
@@ -0,0 +1,51 @@
...
+        "make sure you are merged from master and "
+        'update the "currentVersion" field in the '
+        "pack_metadata.json or in case release notes are required run:\n"
+        "`demisto-sdk update-release-notes -i Packs/{pack} -u "
+        "(major|minor|revision|documentation)` to "
+        "generate them according to the new standard."
+    )
+    related_field = "currentVersion, name"
+    is_auto_fixable = False
+    # Validate manager should populate the external args for the validation.
```
YuvHayun, line 33:
```suggestion
```
Remove.

demisto_sdk/commands/validate/tests/validators_test.py
```diff
@@ -763,3 +770,41 @@ def test_get_unfiltered_changed_files_from_git_in_external_pr_use_case(
...
+    """
+    Given: pack with newly added integration.
+    When: Initializing ValidateManager using git
+    Then: Ensure PackMetadataVersionShouldBeRaisedValidator is initialized and the external args are properly passed.
+    """
+    pack = create_pack_object(["currentVersion"], ["1.1.1"])
+    integration = create_integration_object()
+    pack.content_items.integration.extend(integration)
+    validation_results = ResultWriter()
+    config_reader = ConfigReader(category_to_run="use_git")
```
YuvHayun, line 785: you can use the new argument **specific_validations** here.
validation_config shouldn't be updated in any case.

demisto_sdk/commands/validate/tests/PA_validators_test.py
```diff
@@ -1615,3 +1618,31 @@ def test_PackFilesValidator_fix(file_attribute: str):
...
+    pack = create_pack_object(["currentVersion"], [current_version])
+    integration = create_integration_object()
+    pack.content_items.integration.extend(integration)
+
+    class MockOldMetadata:
+        current_version = old_version
+
+    pack.old_base_content_object = MockOldMetadata()
+    version_bump_validator = PackMetadataVersionShouldBeRaisedValidator()
+    assert len(version_bump_validator.is_valid([pack])) == expected_invalid
```
YuvHayun, line 1648: Let's also assert the messages.

YuvHayun (changes requested):
Nice!
see my notes.

demisto_sdk/commands/validate/initializer.py
DinaMeylakh: We are still validating them. I also believe they are being parsed into graph objects, just not their previous versions. 
Even if I don't change the pack metadata, I'm still running validate on it & I would like to know about it from the logs. Don't you think?

demisto_sdk/__main__.py
```diff
@@ -920,6 +919,7 @@ def validate(ctx, config, file_paths: str, **kwargs):
                 prev_ver=kwargs["prev_ver"],
                 file_path=file_path,
                 all_files=kwargs.get("validate_all"),
+                skip_pack_dependencies=kwargs.get("skip_pack_dependencies"),
```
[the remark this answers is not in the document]
DinaMeylakh: this is outdated, from before we talked. Already removed :)

demisto_sdk/commands/validate/initializer.py
DinaMeylakh: not sure I follow..
The implementation of is_pack_item returns True if the file is one of the following:
`PACKS_PACK_IGNORE_FILE_NAME,
RELEASE_NOTES_DIR,
PACKS_WHITELIST_FILE_NAME,
PACKS_README_FILE_NAME,
AUTHOR_IMAGE_FILE_NAME,
PACKS_CONTRIBUTORS_FILE_NAME,
DOC_FILES_DIR,`

The else refers to the following conditions:
if f"/{INTEGRATIONS_DIR}/" in path_str or f"/{SCRIPTS_DIR}/" in path_str: ...
elif f"/{PLAYBOOKS_DIR }/" in path_str: ...
elif MODELING_RULES_DIR in path_str or PARSING_RULES_DIR in path_str:...
elif PACKS_PACK_META_FILE_NAME in path_str:...

Correct me if I'm wrong but if any of these conditions hold is_pack_item will return False anyway, so no harm in making it an if statement instead of an elif

DinaMeylakh: good catch

demisto_sdk/commands/validate/initializer.py
YuvHayun: Not sure this log is necessary but we can leave it there if you think it does.

demisto_sdk/commands/validate/initializer.py
```diff
@@ -525,7 +531,10 @@ def git_paths_to_basecontent_set(
                 if obj:
                     obj.git_status = git_status
                     # Check if the file exists
-                    if git_status in (GitStatuses.MODIFIED, GitStatuses.RENAMED):
+                    if (
+                        git_status in (GitStatuses.MODIFIED, GitStatuses.RENAMED)
+                        or find_type_by_path(file_path) == FileType.METADATA
```
YuvHayun, line 536:
```suggestion
                        or find_type_by_path(file_path) == FileType.METADATA and not gitstatus
```

git status can also be added files which may lead to issues with attempting to create old file object without an existing one.

YuvHayun (approved):
Looks good, minor comment.
Approved from my side.

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
```diff
@@ -0,0 +1,49 @@
...
+        "make sure you are merged from master and "
+        'update the "currentVersion" field in the '
+        "pack_metadata.json or in case release notes are required run:\n"
+        "`demisto-sdk update-release-notes -i Packs/{pack} -u "
+        "(major|minor|revision|documentation)` to "
+        "generate them according to the new standard."
+    )
+    related_field = "currentVersion, name"
+
+    def is_valid(self, content_items: Iterable[ContentTypes]) -> List[ValidationResult]:
```
ShahafBenYakir, line 33: I don't see any reference to the type of change here at all seems like we always require bump based on this logic

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
```diff
@@ -0,0 +1,184 @@
...
+        return False, is_metadata
+
+    def obtain_invalid_content_items(
+        self, content_items: Iterable[ContentTypes]
+    ) -> typing.List[ValidationResult]:
+        validation_results = []
+        content_packs = {}
+        content_packs_ids_to_bump = []
+        # Go over all the content items
+        for content_item in content_items:
```
YuvHayun, line 155: can we do it in a way that group items by pack and then check once for each pack? the content items are hashables so you should be able to do it with a dictionary
Pack Object: Changed pack items set

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
```diff
@@ -0,0 +1,184 @@
...
+            and content_item.content_type.value == Pack.content_type.value  # type: ignore[union-attr]
+        ):
+            content_item_type = FileType.METADATA
+            is_metadata = True
+
+        if content_item_type not in SKIP_RELEASE_NOTES_FOR_TYPES:
+            if content_item_type != FileType.METADATA:
+                return True, False
+            else:
+                is_metadata = True
```
YuvHayun, line 140: Why do we need to check if is_metadata again?

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
```diff
@@ -0,0 +1,184 @@
...
+        """Return if the version should be bumped as a result of the input content item.
+        Args:
+            content_item (ContentTypes): The content item for which to decide.
+
+        Return:
+            ((Boolean) True if should bump due to change False otherwise,
+             (Boolean) True if the content item is the pack_metadata False otherwise)
+        """
+        content_item_type = find_type(str(content_item.path))  # type: ignore[union-attr]
+        is_metadata = False
```
YuvHayun, line 127: can be determined once by the pack object

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
```diff
@@ -0,0 +1,184 @@
...
+                    if old_dict.get(field) != current_dict.get(field):
+                        return True, True
+        return False, is_metadata
+
+    def obtain_invalid_content_items(
+        self, content_items: Iterable[ContentTypes]
+    ) -> typing.List[ValidationResult]:
+        validation_results = []
+        content_packs = {}
+        content_packs_ids_to_bump = []
```
YuvHayun, line 153: why not using a set?

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
```diff
@@ -0,0 +1,184 @@
...
+                content_packs[content_item.pack_id] = content_item  # type: ignore[union-attr]
+
+        # Go over all the pack ids that need to be bumped.
+        for pack_id in content_packs_ids_to_bump:
+            # Access them via the dict that was created earlier
+            pack = content_packs[pack_id]
+            # Check if their old version >= current version
+            old_version = pack.old_base_content_object.current_version  # type: ignore[union-attr]
+            current_version = pack.current_version  # type: ignore[union-attr]
+            if current_version and Version(old_version) >= Version(current_version):
```
YuvHayun, line 173: maybe we can do this check first? might be quicker than iterating on all items especially after we group by pack

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
```diff
@@ -0,0 +1,184 @@
...
+        "make sure you are merged from master and "
+        'update the "currentVersion" field in the '
+        "pack_metadata.json or in case release notes are required run:\n"
+        "`demisto-sdk update-release-notes -g` - for automatically generation of release notes and version\n"
+        "`demisto-sdk update-release-notes -i Packs/{pack} -u "
+        "(major|minor|revision|documentation)` for a specific pack and version."
+    )
+
+    @staticmethod
+    def should_bump_is_metadata(content_item: ContentTypes):
```
YuvHayun, line 117: Keep in mind that if the content_item was collected it doesn't mean we need to bump.
For example - integration image change doesn't require version bump but it'll collect the integration object.

demisto_sdk/commands/validate/tests/validators_test.py
```diff
@@ -816,3 +823,43 @@ def test_get_unfiltered_changed_files_from_git_in_external_pr_use_case(
...
+    validate_manager = ValidateManager(
+        validation_results=validation_results,
+        config_reader=config_reader,
+        initializer=initializer,
+    )
+
+    version_bump_validator = None
+    for validator in validate_manager.validators:
+        if isinstance(validator, PackMetadataVersionShouldBeRaisedValidator):
+            version_bump_validator = validator
```
YuvHayun, line 862: why do we need this if we are using the explicitly_selected?

YuvHayun (changes requested):
Great work! see my comments.

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
DinaMeylakh: In this case will the content_item.path point to the integration.yml or to the integration image?

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
DinaMeylakh: for metadata, content_item_type can either return None or FileType.METADATA, if it's None and it is the metadata (content_item.content_type.value == Pack.content_type.value) then update the content_item_type accordingly. 

we need this in the case find_type returned FileType.METADATA instead of None.

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
DinaMeylakh: what do you mean?

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
DinaMeylakh: wouldn't that just result in an extra for loop to sort them? why would we want this?

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
DinaMeylakh: We don't know if we should bump yet before / we are not sure we have collected the metadata content item yet. Not sure I fully understand your comment.

demisto_sdk/commands/validate/tests/validators_test.py
DinaMeylakh: This is to check that even though a pack_metadata was not changed, the validation is collected and will run.

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
```diff
@@ -0,0 +1,196 @@
...
+
+        if content_item.git_status is None and content_item.content_type.value in [
+            Integration.content_type.value,
+            Script.content_type.value,
+        ]:
+            # If the file collected is an Integration or Script and they were not modified directly,
+            # check that their code files and description files were not modified as well.
+            related_files_unchanged = [
+                content_item.code_file.git_status is None,  # type: ignore[union-attr]
+                content_item.description_file.git_status is None,  # type: ignore[union-attr]
```
YuvHayun, line 135: won't fail on script description?

YuvHayun (approved):
Amazing work!
Approved from my side.

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
DinaMeylakh: good catch

.github/workflows/on-push.yml
```diff
@@ -284,7 +284,7 @@ jobs:
           cd content
           rm -rf .pre-commit-config_template.yaml
           rm -f Tests/docker_native_image_config.json
-          demisto-sdk pre-commit --validate --show-diff-on-failure --verbose
+          demisto-sdk pre-commit --no-validate --show-diff-on-failure --verbose
```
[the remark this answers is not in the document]
ShahafBenYakir: why we changed --validate  to --no-validate

demisto_sdk/commands/validate/initializer.py
```diff
@@ -578,13 +585,20 @@ def get_items_status(
...
+            metadata_path = self.obtain_metadata_path(path)
+            if metadata_path not in statuses_dict:
+                statuses_dict[metadata_path] = None
+
+        all_collected_files = set(statuses_dict.keys())
+        git_only_files = set(file_by_status_dict.keys())
+        related_files = all_collected_files.difference(git_only_files)
+        if related_files:
+            logger.info("Running on related files:")
+            logger.info(f"{[str(path) for path in related_files]}")
```
ShahafBenYakir, line 600: This is only for log printing right? let's move ti a dedicated function.

demisto_sdk/commands/validate/initializer.py
```diff
@@ -578,13 +585,20 @@ def get_items_status(
                     statuses_dict[path] = git_status
             elif PACKS_PACK_META_FILE_NAME in path_str:
                 statuses_dict[path] = git_status
-            elif self.is_pack_item(path_str):
-                metadata_path = self.obtain_metadata_path(path)
-                if metadata_path not in statuses_dict:
-                    statuses_dict[metadata_path] = None
-            else:
+            elif not self.is_pack_item(path_str):
```
ShahafBenYakir, line 588: Since we have several nested if/elif here I would appreciate if you could add a comment explaining each of the, for clarity

ShahafBenYakir (approved):
Good job

.github/workflows/on-push.yml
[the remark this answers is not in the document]
DinaMeylakh: previously this was used to check that pre-commit will run after some changes in the code. This made PA114 fail successfully because the pack metadata version was not updated. Since I started it seems someone fixed the issue and validate now only runs a specific validation to see that it is triggered (and PA114 is skipped).

Changed it back to the way it works now on master.

PR #4361, opened 2024-06-17
RP103

ilaredo:
## Description
Moved RP103 to new validate


.changelog/4349.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+  - description: Added RP103 to the new validation format.
```
YuvHayun:
```suggestion
  - description: Added RP103 to the new validation format. The validation <what it does>.
```

.changelog/4361.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+  - description: Added RP103 to the new validation format.
```
YuvHayun, line 2:
```suggestion
  - description: Added RP103 to the new validation format. The validation <what id does>.
```

YuvHayun (approved):
Nice, minor comments.

demisto_sdk/commands/validate/tests/RP_validatiors_test.py
```diff
@@ -55,3 +58,49 @@ def test_ExpirationFieldIsNumericValidator_is_valid(
             for result, expected_msg in zip(results, expected_msgs)
         ]
     )
+
+
+@pytest.mark.parametrize(
```
dantavori, line 63: as you might have heard - avoid using pytest.mark.parametrize
instead, split into different test methods, each one with more specific naming and docstring based on the test case.
Another option: just call the is_valid() multiple times in the same method. Still, I'd split into "valid" and "invalid" test methods.

Generally, this decision was made collectively in the team a few months ago, and the goal is for better readability and maintenance for the tests.
there are some cases where it is ok to use parametrize, e.g. - in very very simple tests with up to 2 arguments, that shouldn't be modified with more test-cases over the time. When you think this is the case, consult with Eyal/me if you want to use it.

demisto_sdk/commands/common/constants.py
```diff
@@ -821,6 +821,8 @@ class FileType(StrEnum):
 
 LAYOUT_FILE_NAME__REGEX = r"layout.*\.json"
 
+VALID_INDICATOR_TYPE = "^[A-Za-z0-9_& ]*$"
```
dantavori, line 824:
```suggestion
VALID_INDICATOR_TYPE_REGEX = "^[A-Za-z0-9_& ]*$"
```

demisto_sdk/commands/validate/validators/RP_validators/RP103_is_valid_indicator_type_id.py
```diff
@@ -0,0 +1,41 @@
...
+ContentTypes = IndicatorType
+
+
+# Valid indicator type can include letters, numbers whitespaces, ampersands and underscores.
+
+
+class IsValidIndicatorTypeId(BaseValidator[ContentTypes]):
+    error_code = "RP103"
+    description = "Validate that the 'id' field of indicator type has valid value."
+    error_message = "id field contain invalid value."
```
dantavori, line 23:
```suggestion
    error_message = "The `id` field must consist of alphanumeric characters (A-Z, a-z, 0-9), underscores (_), and ampersands (&) only."
```

dantavori (commented):
nice! see notes

dantavori (approved):

PR #4354, opened 2024-06-17
Moved removed_integration_parameters to new validations

jlevypaloalto:
## Related Issues
fixes: CIAC-7519

## Description
Moves the `removed_integration_parameters` validation to the new validation

PR #4362, opened 2024-06-17
Fix no_removed_integration_parameters unit-tests

jlevypaloalto:
## Related Issues
fixes: #4354


JudahSchwartz (approved):
check that this is the only test affected

PR #4368, opened 2024-06-18
Mr101 validation

Shellyber:
## Related Issues
relates: https://jira-dc.paloaltonetworks.com/browse/CIAC-7814

## Description
- Combined the validations MR101 + MR102 that is checking that the `schema` and `rules` keys exist and empty in the modeling rules yml
- Updated the mapping_fields for modeling rules


demisto_sdk/commands/validate/validators/MR_validators/MR101_validate_empty_keys.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def is_valid(self, content_items: Iterable[ContentTypes]) -> List[ValidationResult]:
+        return [
+            ValidationResult(
+                validator=self,
+                message=self.error_message,
+                content_object=content_item,
+            )
+            for content_item in content_items
+            if content_item.schema_key != "" or content_item.rules_key != ""
```
ShahafBenYakir, line 34:
```suggestion
            if content_item.schema != "" or content_item.rules != ""
```

demisto_sdk/commands/validate/validators/MR_validators/MR101_validate_empty_keys.py
```diff
@@ -0,0 +1,35 @@
...
+
+ContentTypes = ModelingRule
+
+
+class ValidateEmptyKeysValidator(BaseValidator[ContentTypes]):
+    error_code = "MR101"
+    description = (
+        "Validate that the modeling rules keys - rules and schema exist and are empty"
+    )
+    rationale = "This standardization is to handle modeling rules correctly."
```
ShahafBenYakir, line 19: Rationale should explain why this validation is needed, i.e. why we want to keep the schema and modeling rules key empty. You can sync with the siem team or check the original ticket requesting the v validation to better understand this.

demisto_sdk/commands/validate/validators/MR_validators/MR101_validate_empty_keys.py
```diff
@@ -0,0 +1,35 @@
...
+    ValidationResult,
+)
+
+ContentTypes = ModelingRule
+
+
+class ValidateEmptyKeysValidator(BaseValidator[ContentTypes]):
+    error_code = "MR101"
+    description = (
+        "Validate that the modeling rules keys - rules and schema exist and are empty"
```
ShahafBenYakir, line 17:
```suggestion
        "Validate that the modeling rules keys - "rules" and "schema" exist and are empty"
```

ShahafBenYakir (approved):
Good job

demisto_sdk/commands/validate/validators/MR_validators/MR101_validate_empty_keys.py
ShahafBenYakir: This should work due to the alias

Shellyber: This is not correct. 
In the modeling rules, I added it as a schema_key and rules_key. And so is the property. 
I can't use `schema` because Pydetic already uses it

Shellyber: The alias lets you get to the field itself, that's why it works.
But it does not know what rules are as part of the object

PR #4380, opened 2024-06-23
Added support for run-specific-validations in new validate

YuvHayun:
## Related Issues
fixes: [CIAC-10893](https://jira-dc.paloaltonetworks.com/browse/CIAC-10893)

## Description
Added support for run-specific-validations in new validate.

### DIFF
--- a/.changelog/4380.yml
+++ b/.changelog/4380.yml
@@ -1,0 +1,4 @@
+changes:
+- description: Added support to **run-specific-validations** flag to the new validate format.
+  type: feature
+pr_number: 4380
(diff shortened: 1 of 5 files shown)


.changelog/4380.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Added support to **run-specific-validations** flag to the new validate format.
```
JudahSchwartz, line 2: can we also add a flag to --ignore-specific-validations like we discussed?

JudahSchwartz (approved):
amazing!

.changelog/4380.yml
YuvHayun: What would be the point in adding such flag?
I don't think it's necessary

PR #4375, opened 2024-06-20
added marketplace values to schemas

DinaMeylakh:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10623

## Description
Added the option to specify a marketplace to the following playbook fields: inputs, scriptarguments, message, form, left, right, tasks.
Made it so the top hierarchy always wins (and is overridden) when collecting schema marketplace specific arguments.

PR #4401, opened 2024-07-04
Add PB marketplace keys validation

DinaMeylakh:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10623

## Description
Recently we added marketplace suffixes to the playbook schema [here](https://github.com/demisto/demisto-sdk/pull/4375). To verify that mandatory keys are present, we demand that any marketplace suffix key will also have a default key, otherwise upload will remove the key at unified yml construction. 
This validation assures playbooks with marketplace suffix keys that don't have default keys are considered invalid.


demisto_sdk/commands/validate/validators/PB_validators/PB127_marketplace_keys_have_default_value.py
```diff
@@ -0,0 +1,87 @@
...
+                value = datum[key]
+                if isinstance(value, (list, dict)):
+                    # If the value is a list or a dictionary recursively check.
+                    bad_inner_keys = self.check_recursively(
+                        value, path=f"{path}.{key}", bad_keys=bad_keys, fix=fix
+                    )
+                    bad_keys.extend(
+                        bad_key for bad_key in bad_inner_keys if bad_key not in bad_keys
+                    )
+                if isinstance(key, str) and SEPARATOR in str(key):
```
ShahafBenYakir, line 74: For readability (makes the reader understand it could be only one of the two options and not both
```suggestion
                elif isinstance(key, str) and SEPARATOR in str(key):
```

demisto_sdk/commands/validate/validators/PB_validators/PB127_marketplace_keys_have_default_value.py
```diff
@@ -0,0 +1,87 @@
...
+                    for suffix in self.marketplace_suffixes:
+                        if key.casefold().endswith(suffix):
+                            # Construct the default key without the suffix and check if it exists in the datum.
+                            clean_key = key[: -(len(suffix) + len(SEPARATOR))]
```
ShahafBenYakir, line 78: Why not use split?
```suggestion
                            clean_key = key.split(SEPARATOR)[0]
```

demisto_sdk/commands/validate/validators/PB_validators/PB127_marketplace_keys_have_default_value.py
```diff
@@ -0,0 +1,87 @@
...
+                        message=self.error_message.format(
+                            marketplace_keys_with_no_default
+                        ),
+                        content_object=content_item,
+                    )
+                )
+
+        return validation_results
+
+    def check_recursively(self, datum, path="root", bad_keys=[], fix=False) -> list:
```
ShahafBenYakir, line 50: I see that you added a "fix" param but it was not used.

demisto_sdk/commands/validate/validators/PB_validators/PB127_marketplace_keys_have_default_value.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class MarketplaceKeysHaveDefaultValidator(BaseValidator[ContentTypes]):
+    error_code = "PB127"
+    description = "Ensure that any yml keys that are marketplace only have a default non-marketplace counterpart."
+    rationale = "To ensure the required yml keys, we have to have a default value for any marketplace only keys."
+    error_message = (
+        "The following playbook yml keys only do not have a default option: {}. Please remove these keys or add "
+        "another default option to each key."
+    )
```
ShahafBenYakir, line 25:
```suggestion
    error_message = (
        "The following playbook yml keys only do not have a default option: {}. Please remove these keys or make sure to include the key without a marketplace suffix"
    )
```

demisto_sdk/commands/validate/validators/PB_validators/PB127_marketplace_keys_have_default_value.py
```diff
@@ -0,0 +1,87 @@
...
+    fix_message = "Added default value to the following playbook keys: {0}."
```
ShahafBenYakir, line 26: I see we did not implemented any fix function, let's do it.

ShahafBenYakir (commented):
Nice work

demisto_sdk/commands/validate/validators/PB_validators/PB127_marketplace_keys_have_default_value.py
DinaMeylakh: it can be both. For example:

```
inputs:
   key: some_key
   value:xsoar:
       simple:xsoar_saas: '5'
       simple: '6'
  value:
      simple: '7'
```

demisto_sdk/commands/validate/validators/PB_validators/PB127_marketplace_keys_have_default_value.py
```diff
@@ -0,0 +1,87 @@
...
+        # If the datum is a list, iterate over its values and check recursively.
+        if isinstance(datum, list):
+            for index, item in enumerate(datum):
+                bad_inner_keys = self.check_recursively(
+                    item, path=f"{path}.[{index}]", bad_keys=bad_keys, fix=fix
+                )
+                bad_keys.extend(
+                    bad_key for bad_key in bad_inner_keys if bad_key not in bad_keys
+                )
+            return bad_keys
```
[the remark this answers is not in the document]
DinaMeylakh: good catch

ShahafBenYakir (commented):
🚀

demisto_sdk/commands/validate/validators/PB_validators/PB127_marketplace_keys_have_default_value.py
```diff
@@ -44,44 +48,106 @@ def is_valid(self, content_items: Iterable[ContentTypes]) -> List[ValidationResu
...
+        if (
+            last_key.startswith("[")
+            and last_key.endswith("]")
+            and isinstance(data_node, list)
+        ):
+            last_key = int(key[1:-1])
+
+        return data_node, last_key
+
+    def fix(self, content_item: ContentTypes) -> FixResult:
```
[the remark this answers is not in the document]
DinaMeylakh: After is_valid is run, the dictionary `bad_paths_dict` is updated with all the paths that do not contain a default.
The `fix` method then goes over all the bad paths, and looks for a marketplace suffix. This is done in an orderly fashion set by the order in MarketplaceVersions. 
After it finds one, it updates the content item data to contain a default with this value and continues to the next bad path. 
It then returns a fixed content item data. 

The docstring says:
"Add default value to keys with marketplace suffix. Add the first one encountered."
& there are comments throughout the function. Do you think there is something I could add to make it clearer?

ShahafBenYakir (approved):

PR #4412, opened 2024-07-09
Ciac 10644 add timestamp

talzich:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10644

## Description
Adding timestamp for console logs, not only when running with CI env variable.


demisto_sdk/commands/common/logger.py
```diff
@@ -117,12 +117,8 @@ def environment_variable_to_int(variable_name: str, default_value: int) -> int:
 
 FILE_LOG_RECORD_FORMAT = "[%(asctime)s] - [%(threadName)s] - [%(levelname)s] - %(filename)s:%(lineno)d - %(message)s"
 
-if environment_variable_to_bool("CI"):
-    CONSOLE_LOG_RECORD_FORMAT = "[%(asctime)s] [%(levelname)s] %(message)s"
-    CONSOLE_LOG_RECORD_FORMAT_SHORT = "[%(asctime)s] [%(levelname)s] "
-else:
-    CONSOLE_LOG_RECORD_FORMAT = "[%(levelname)s] %(message)s"
```
JudahSchwartz, line 124: this means that the time will be printed by every log. Is that what we were going for ? I thought we just wanted 1 print to the console with the time

talzich: From what I could understand from @dorschw, every sdk command and every print should have a timestamp. See his comment on this ticket https://jira-dc.paloaltonetworks.com/browse/CIAC-10644.
Dor, could you approve or correct me if I'm wrong?

JudahSchwartz (approved):

PR #4430, opened 2024-07-16
Add support attribute to all content items

RosenbergYehuda:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10624

## Description
Adding a "support" attribute to all content items:
Now, content items can have their own support level, independent of the pack support level.
If no specific support level is provided for a content item, it will inherit the support level from its pack.


demisto_sdk/commands/content_graph/objects/content_item.py
```diff
@@ -75,13 +76,64 @@ def match(_dict: dict, path: Path) -> bool:
...
+    def in_pack(self) -> Optional["Pack"]:
+        """
+        This returns the Pack which the content item is in.
+
+        Returns:
+            Pack: Pack model.
+        """
+        if pack := ContentItem.get_pack(self.pack, self.relationships_data, self.path):
+            self.pack = pack
+        return pack  # type: ignore[return-value]
```
dantavori, line 101: you can do:
```suggestion
        if not self.pack:
            self.pack = ContentItem.get_pack(self.pack, self.relationships_data, self.path)
        return pack  # type: ignore[return-value]
```
so self.in_pack will not always run ContentItem.get_pack()

demisto_sdk/commands/content_graph/objects/content_item.py
```diff
@@ -75,13 +76,64 @@ def match(_dict: dict, path: Path) -> bool:
...
+
+    @staticmethod
+    def get_pack(
+        pack: Any,
```
dantavori, line 105: not critical but we don't really need the `pack` argument if we call get_pack only when it's None anyway

demisto_sdk/commands/validate/validators/base_validator.py
```diff
@@ -138,7 +138,9 @@ def should_run(
                     self.related_file_type,
                 ),
                 not is_support_level_support_validation(
-                    self.error_code, support_level_dict, content_item.support_level
+                    self.error_code,
+                    support_level_dict,
+                    content_item.support,  # type: ignore
```
JasBeilin, line 143: why ignoring type?

demisto_sdk/commands/content_graph/tests/parsers_and_models_test.py
JasBeilin, line 1: Lets have separate test to check for pack with x support level, and integration with None, x and y cases.

demisto_sdk/tests/integration_tests/unify_integration_test.py
```diff
@@ -333,8 +333,9 @@ def test_layout_unify(self, repo, mocker, monkeypatch):
 
             assert result.exit_code == 0
             assert not result.exception
-
-            assert logger_warning.call_count == 0
+            #!!!!!!
+            # TODO: Fix this test to b 0
+            assert logger_warning.call_count == 1
```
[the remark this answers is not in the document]
RosenbergYehuda: was fixed

demisto_sdk/commands/validate/validators/base_validator.py
RosenbergYehuda: since we delletted these lines in BaseContent:
    @property
    def support_level(self) -> str:
        raise NotImplementedError

    and I think its better to ignore than to addd a property that reises and never in use

demisto_sdk/commands/content_graph/objects/content_item.py
```diff
@@ -75,13 +76,60 @@ def match(_dict: dict, path: Path) -> bool:
     def pack_id(self) -> str:
         return self.in_pack.pack_id if self.in_pack else ""
 
+    @validator("pack", always=True)
+    def validate_pack(cls, v: Any, values) -> Optional["Pack"]:
+        # Validate that we have the pack containing the content item.
+        # The pack is either provided directly or needs to be located.
+
+        pack = cls.get_pack(values.get("relationships_data"), values.get("path"))
+        if v and not isinstance(v, fields.FieldInfo):
```
dantavori, line 85: when will it not be FieldInfo?

demisto_sdk/commands/content_graph/objects/content_item.py
```diff
@@ -75,13 +76,60 @@ def match(_dict: dict, path: Path) -> bool:
...
+            return v
+        return pack
```
dantavori, line 87:
```suggestion
        if v and not isinstance(v, fields.FieldInfo):
            return v
        return cls.get_pack(values.get("relationships_data"), values.get("path"))

```

demisto_sdk/commands/content_graph/objects/content_item.py
```diff
@@ -75,13 +76,60 @@ def match(_dict: dict, path: Path) -> bool:
...
+        if in_pack := relationships_data[RelationshipType.IN_PACK]:
+            pack = next(iter(in_pack)).content_item_to  # type: ignore[return-value]
+        if not pack:
+            if pack_name := get_pack_name(path):
+                pack = BaseContent.from_path(
+                    CONTENT_PATH / PACKS_FOLDER / pack_name, metadata_only=True
+                )  # type: ignore[assignment]
+        return pack  # type: ignore[return-value]
+
+    @validator("support", always=True)
```
dantavori, line 122: let's move up below the other validators, just for better organization of the code

demisto_sdk/commands/validate/validators/base_validator.py
dantavori: isn't it because `content_item.support` is `Optional[str]` and not `str`?
Can we make sure it's `str`, as we expect it to be in reality after the validator? It requires Pack's `support` to be `str` as well

dantavori (commented):
nice, some minor notes

demisto_sdk/commands/content_graph/objects/content_item.py
RosenbergYehuda: Great question, It was copied from what Ilan wrote here:
https://github.com/demisto/demisto-sdk/blob/master/demisto_sdk/commands/content_graph/objects/content_item.py#L132

I usually avoid changing things that seem to be there on purpose, even if I don't get why. But I tried removing it and the build was ok with it. What do you think?

dantavori: I think we should check if it's necessary and remove if not
it's just confusing if it's not relevant and next time people do something similar they will add it also for no reason

demisto_sdk/commands/validate/validators/base_validator.py
dantavori: there's always a way to make mypy happy. two things we can try:
1. not sure it will work but I have a feeling that yes:
```
if isinstance(integration_script, IntegrationScript):
        file = integration_script.path.parent / f"{integration_script.path.stem}.py"
        if not file.exists():
            return results
    xsoar_linter_env = build_xsoar_linter_env_var(integration_script)
    env.update(xsoar_linter_env)
    command = build_xsoar_linter_command(integration_script.support)
else:
   return results
```

2. if the one above doesn't work:
```
from typing import cast
...
integration_script = cast(IntegrationScript, integration_script)
```

dantavori (approved):
approving but see my 2 comments

demisto_sdk/commands/validate/validators/base_validator.py
dantavori: we're getting:
```
"ContentTypes" has no attribute "support"  [attr-defined]
```
mypy is asking you legitimately, is there any chance `content_item` will not have this property?
if there is a chance, make sure it doesn't reach this code block!
if not - why is it "BaseContent" and not one of the actual type that really has this field?

in both cases, this can be solved by putting the code in the `if isinstance(content_item, ContentItem)` block, or `content_item = cast(ContentItem, content_item)` (I prefer the first one)

PR #4435, opened 2024-07-18
Replace `Black` with `ruff-format`

dorschw:
- update ruff hook to 0.5.1 
- remove black hook
- remove black dependency
- use `ruff-format` instead of `Black`
- apply slight differences between the two formatters


samuelFain (approved):
Nice!

PR #4449, opened 2024-07-23
moved IN147 to BC114

YuvHayun:
## Related Issues
fixes:


.changelog/4449.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Moved IN147 validation to the new validate format as BC114. The validation ensures no fields were removed/modified from the integration yml for a predefined list of fields..
```
JudahSchwartz, line 2:
```suggestion
- description: Moved IN147 validation to the new validate format as code BC114. The validation ensures no fields were removed/modified from the integration yml for a predefined list of fields..
```

demisto_sdk/commands/validate/validators/BC_validators/BC114_is_changed_or_removed_fields.py
```diff
@@ -0,0 +1,82 @@
...
+                        ]
+                    )
+                ),
+                content_object=content_item,
+            )
+            for content_item in content_items
+            if (old_obj := content_item.old_base_content_object)
+            and (old_obj_data := old_obj.data)  # type: ignore[attr-defined]
+            and (
+                altered_fields := self.obtain_removed_fields(
```
JudahSchwartz, line 48: is the issue altering? or only removing. Seems like we shouldnt be able to go from a feed to not a feed, but we should be able to add these features. 
asking because you have `removed or modified` in the description

demisto_sdk/commands/validate/validators/BC_validators/BC114_is_changed_or_removed_fields.py
```diff
@@ -0,0 +1,82 @@
...
+        for field in INTEGRATION_FIELDS_NOT_ALLOWED_TO_CHANGE:
+            if old_field := old_data.get(field):
+                if (current_field := current_data.get(field)) in [
+                    True,
+                    False,
+                    "true",
+                    "false",
+                ]:
+                    if current_field != old_field:
+                        modified_fields.append(field)
```
JudahSchwartz, line 77: not sure this makes sense to not be able to add features to this integration. I think we should only be checking if it was changed to false

demisto_sdk/commands/validate/validators/BC_validators/BC114_is_changed_or_removed_fields.py
YuvHayun: Yea, modified it to work only on removing, good point!

PR #4464, opened 2024-07-30
[TestContext] Fix the Investigation URL in XSOAR SAAS

mmhw:
## Description
[TestContext] Fix the Investigation URL in XSOAR SAAS


omerKarkKatz (approved):
Looks good!
Please check all the cases before merging XSIAM/XSOAR/XSOAR8

PR #4495, opened 2024-08-15
Fixing the way we try to remove old containers

RosenbergYehuda:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-11519

## Description
fixing the way we try to remove old containers by adding "all=True" to when we search for the old container in order to remove him.


JasBeilin (approved):
NICE

PR #4512, opened 2024-08-21
Fix RM114

BEAdi:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-11204

## Description
Extend RM114 to better find whether an image in README exists, using `click.Path`  in order to find more cases of missing pictures (jpg file type, and pictures that have unneeded chars in their path).

The fixed UT failing on the old code: https://github.com/demisto/demisto-sdk/actions/runs/10496298391/job/29076644173?pr=4512 - fails to raise that 3 more cases of images doesn't exist.


demisto_sdk/commands/common/constants.py
```diff
@@ -1633,6 +1633,7 @@ class PB_Status:
 FEATURE_BRANCHES = ["v4.5.0"]
 VERSION_REGEX = r"(\d{1,2}\.){2}\d{1,2}$"
 DOC_FILE_IMAGE_REGEX = r"[\.\./]*doc_files/[a-zA-Z0-9_-]+\.png"
+DOC_FILE_FULL_IMAGE_REGEX = r"!\[.*?\]\(((?!http).*?doc_files.*?)\)"
```
BEAdi, line 1636: `(?!http)` in order to catch only files in the repo and `doc_files` in order to catch only files that are under doc_files folder. RM108 should validate all images are located there - [related ticket](https://jira-dc.paloaltonetworks.com/browse/CIAC-10606).

jbabazadeh (approved):
great!

PR #4549, opened 2024-09-12
[GR106] reputation_tests

shmuel44:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-11761

## Description
This validation 
```
Checks that every test playbook is linked to at least one content item.
 (the content item has a 'tests:' key with the id of the test playbook)
 ```
But the the collect test for reputation playbooks checks the r`eputation_tests` field in conf.json and not the `tests` key in the yml so we need to ignore reputation test

### DIFF
--- a/.changelog/4549.yml
+++ b/.changelog/4549.yml
@@ -1,0 +1,4 @@
+changes:
+- description: Fixed an issue where the GR106 validation rule incorrectly flagged reputation tests as unused test playbooks.
+  type: fix
+pr_number: 4549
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: .changelog/4549.yml [hunk 1]
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Fixed an issue where the GR106 validation rule incorrectly flagged reputation tests as unused test playbooks.
```
dorschw, line 2:
```suggestion
- description: Fixed an issue where the GR106 validation would fail on test playbooks that are marked as reputation tests.
```

dorschw (approved):
Excellent

PR #4646, opened 2024-11-07
Hide the convert command

Shellyber:
## Related Issues
relates: https://jira-dc.paloaltonetworks.com/browse/CIAC-12125

## Description
- Hide the convert command so it won't be visible to customers any more
- Add a changelog that will let the customer know that the following commands will be deprecated and won't be usable anymore:
  - Convert
  - create-id-set
  - create-content-artifacts
  - lint ?
  - extract-code
- Delete some of the deprecated commands tests


samuelFain (commented):
Nice!
Maybe we should add "Deprecated" to `convert` command's description as well? (like in `lint` command):
```none
Deprecated. Convert the content of the pack/directory in the given input to be compatible with the version given by version command.
```
Just so users will also see the deprecation warning when using the `--help` flag.

samuelFain (approved):

PR #4651, opened 2024-11-10
Manual Release 1.32.4

RotemAmit:
### Breaking
* The following commands will be deprecated and removed entirely in the next SDK release in two weeks: convert, create-content-artifacts, create-id-set, extract-code and lint.
* Moved CJ102 to the new validate format. The validation's error number is now BA103 and it ensure that the tests section is either a non-empty list or "No tests".

### Fix
* Fixed an issue where getting a malformed file raised an non-indicative error.
* Fixed an issue where the **validate** and **create-id-set** commands would fail to process correctly when encountering null values in the filters or transformers fields.

### Internal
* Moved RN113 to the new validate format. The validation ensures that all the first level headers in the release note are valid content types.
* Moved RN115 to the new validate format. The validation ensure that the release notes first level headers are valid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@
 * Moved RN113 to the new validate format. The validation ensures that all the first level headers in the release note are valid content types.
 * Moved RN115 to the new validate format. The validation ensure that the release notes first level headers are valid.
 
-
 ## 1.32.3
 * Fixed an issue where integrations and scripts validations didn't run in builds. [#4609](https://github.com/demisto/demisto-sdk/pull/4609)
 * Added BA127 new validation. The validation checks that the level of depth of the context output path in the yml is lower or equal to 5 in XSOAR supported content items. [#4490](https://github.com/demisto/demisto-sdk/pull/4490)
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,18 @@
...
+
+### Fix
+* Fixed an issue where getting a malformed file raised an non-indicative error.
+* Fixed an issue where the **validate** and **create-id-set** commands would fail to process correctly when encountering null values in the filters or transformers fields.
+
+### Internal
+* Moved RN113 to the new validate format. The validation ensures that all the first level headers in the release note are valid content types.
+* Moved RN115 to the new validate format. The validation ensure that the release notes first level headers are valid.
+
+
```
samuelFain, line 15: Removing extra line break to match format of previous releases. 
```suggestion

```

samuelFain (approved):
Nice! Approved.
Please don't merge until the correct step in the release (after the release to Pypi is completed).
@RotemAmit Since we are doing a manual release - see if you can add the release changelog to the PR description to match previous release PRs (the current PR description will be copied and used in the Pypi release, so we want it to be consistent).

PR #4657, opened 2024-11-11
Fix logger setup

barryyosi-panw:
## Related Issues
fixes: [CIAC-12159](https://jira-dc.paloaltonetworks.com/browse/CIAC-12159)

## Description

When using 'demisto-sdk' as a library within content scripts/integrations, the new logging setup/initialization conflicts with existing logger and causes the container to crash when initialized statically.

Additionally, logs does not show up on default python's logging which is used in content scripts/integration.

## Solution:
 - Excluding sdk's logging initialization logic when imported as a library.
 - Propagating log entries to default logging system when flagged.

### DIFF
--- a/.changelog/4657.yml
+++ b/.changelog/4657.yml
@@ -1,4 +1,4 @@
 changes:
-- description: Fixing logger crashes when imported as library in content scripts/integrations.
+- description: Fixing logger crashes when imported as library.
   type: fix
 pr_number: 4657
--- a/.changelog/4657.yml
+++ b/.changelog/4657.yml
@@ -1,0 +1,4 @@
+changes:
+- description: Fixing logger crashes when imported as library in content scripts/integrations.
+  type: fix
+pr_number: 4657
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/common/logger.py
```diff
@@ -93,39 +98,63 @@ def logging_setup(
...
+    - path (Union[Path, str], optional): Path for file logs. If None, defaults to the calculated log directory.
+    - initial (bool): Indicates if the setup is for the initial configuration (console-only if True).
+    - propagate (bool): If True, propagates logs to Python's logging system.
+
     """
     global logger
     _setup_neo4j_logger()
 
-    logger.remove(None)  # Removes all pre-existing handlers
+    logger.remove()  # Removes all pre-existing handlers
```
dorschw, line 119: you sure passing None vs nothing still achieves what we wanted?

.changelog/4657.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Fixing logger crashes when imported as library in content scripts/integrations.
```
dorschw, line 2:
```suggestion
- description: Fixing logger crashes when imported as library.
```

dorschw (commented):
nice

demisto_sdk/commands/common/logger.py
barryyosi-panw: Default value is None than yes, and it is less confusing that way imo. https://loguru.readthedocs.io/en/stable/api/logger.html#loguru._logger.Logger.remove

thefrieddan1 (approved):

PR #4659, opened 2024-11-12
Replace incorrect marketplace

RosenbergYehuda:
## Related Issues
relates: https://jira-dc.paloaltonetworks.com/browse/CIAC-10198

## Description
Modifying the SDK `prepare_content` command to replace all occurrences of an incorrect marketplace reference in the content.
This means that if the `prepare_content` command is called with the marketplace flag set to *xpanse* or *xsiam*, and the code contains the phrase "Cortex XSOAR" or "Cortex XSOAR" followed by a specific version number, it will be replaced with "Cortex".


demisto_sdk/commands/content_graph/tests/common_test.py
```diff
@@ -32,3 +38,161 @@ def test_to_neo4j_pattern():
...
+    expected = "This is a Cortex example."
+    result = replace_incorrect_marketplace(
+        data, MarketplaceVersions.MarketplaceV2, path="example/path"
+    )
+    assert result == expected
+
+
+def test_replace_incorrect_marketplace_string_no_replacement():
+    """
+    Test the replace_incorrect_marketplace function with a string where no replacement should occur.
```
JasBeilin, line 90: Why should it not be replaced in this case?

JasBeilin, line 90:
```suggestion
    Test the replace_incorrect_marketplace function with a string where no replacement should occur in the specific marketplace.
```

JasBeilin (approved):
Nice work.

PR #4663, opened 2024-11-12
support content path update

barryyosi-panw:
## Related Issues
fixes: [CIAC-12161](https://jira-dc.paloaltonetworks.com/browse/CIAC-12161)

## Description
Within the content contributions-flow, the `ValidateContent` script uses demisto-sdk as a library.
The script constructs an isolated, local temporary `Content` repository which its path is required when using demisto-sdk's relevant functionality.

Currently the `CONTENT_PATH` constant is resolved statically (for obvious reasons) when importing the required modules, which prevents the script from utilizing it for validation and lint of the local temporary content repo.

These changes expose an updating functionality for the global `CONTENT_PATH` constant and all of its derivations.


demisto_sdk/__init__.py
barryyosi-panw, line 1: This fixes the change introduced here https://github.com/demisto/demisto-sdk/pull/4657

PR #4665, opened 2024-11-13
fix issue when calling GitUtil directly and not via `Content.git_util()`

kobymeir:
## Description
fix issue when calling GitUtil directly and not via `Content.git_util()`


dorschw (commented):
do we want to add a `TID251` to `Pyproject.toml`, to require everyone to import from content?

PR #4666, opened 2024-11-13
add `setuptools`

dorschw:
Add a `setuptools` dependency, solving an issue where SDK commands would fail in some systems.


samuelFain (approved):
Nice!

PR #4667, opened 2024-11-14
[Valiidate] Fix GR107 

shmuel44:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-11971

## Description

- Fixed an issue where the validation of GR107 was failing on unrelated content-item when running on deprecated content items.
- Improved the error message: instead of returning multiple error messages for each deprecated item, we now consolidate all deprecated items in use into a single error message. 

- Additionally, we have removed duplicates in cases where the playbook uses the same deprecated command in two different tasks; the error will only be raised once instead of twice. 

For example:  
```
RCEs.yml: [GR107] - The item 'RCEs' is using the following deprecated items: redlock-get-rql-response
```


demisto_sdk/commands/validate/validators/GR_validators/GR107_is_deprecated_content_item_in_usage.py
```diff
@@ -99,18 +99,27 @@ def obtain_invalid_content_items_using_graph(
...
         return [
             ValidationResult(
                 validator=self,
                 message=self.error_message.format(
-                    deprecated_item_type=item.deprecated_item_type,
-                    deprecated_item=item.deprecated_item_id,
-                    using_deprecated_item=str(
-                        item_using_deprecated.path.relative_to(CONTENT_PATH)
-                    ),
+                    using_deprecated_item=path,
```
YuvHayun, line 119: Not sure the path is necessary here as it should be a part of the error message.

YuvHayun (approved):
Nice job! 1 minor comment.

.changelog/4667.yml
```diff
@@ -0,0 +1,6 @@
+changes:
+- description: Fixed an issue where validate GR107 was failing on unrelated content-item when running on deprecated content-items.
+  type: fix
+- description: The GR107 validation now fails once on each item and the message contains all deprecated item used, instead of failing multiple times
```
shmuel44, line 4:
```suggestion
- description: The GR107 validation now fails once on each item and the message contains all deprecated item used, instead of failing multiple times.
```

PR #4683, opened 2024-11-26
fix rm108 by removing conent-assets absulote-link

maimorag:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10606 

## Description
removing absolute links that contains "content-assests"


.changelog/4683.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Fixed an issue where content-assets absolute paths raised an error from RM108 validation.
```
YuvHayun, line 2:
```suggestion
- description: Fixed RM108 validation to allow absolute paths if the files are a part of content-assets.
```

demisto_sdk/commands/validate/tests/RM_validators_test.py
```diff
@@ -1126,3 +1126,40 @@ def test_invalid_short_file():
...
+
+
+def test_ImagePathIntegrationValidator_content_assets():
+    """
+        Given
+        content_items.
+        - Pack with:
+            1. invalid readme that contains absolute path.
+            2. description contains relative path that saved not under dec_files.
+    demisto_sdk/commands/validate/sdk_validation_config.toml
```
YuvHayun, line 1138:
```suggestion
    demisto_sdk/commands/validate/sdk_validation_config.toml
```
?

demisto_sdk/commands/validate/tests/RM_validators_test.py
```diff
@@ -1126,3 +1126,40 @@ def test_invalid_short_file():
...
     assert result[0].message == short_readme_error
+
+
+def test_ImagePathIntegrationValidator_content_assets():
+    """
+        Given
+        content_items.
+        - Pack with:
+            1. invalid readme that contains absolute path.
+            2. description contains relative path that saved not under dec_files.
```
YuvHayun, line 1137:
```suggestion
            1. invalid readme that contains absolute path.
            2. description contains relative path that saved not under dec_files.
```

Let's list the exact files that are given and the ones that are invalid - will be better for documenting the expected behavior.

YuvHayun (approved):
Great work!
Approved from my side.

demisto_sdk/commands/validate/tests/RM_validators_test.py
maimorag: by mistake

demisto_sdk/commands/validate/tests/RM_validators_test.py
```diff
@@ -1126,3 +1128,50 @@ def test_invalid_short_file():
...
+
+
+def test_ImagePathIntegrationValidator_content_assets():
+    """
+    Given
+    content_items.
+    - Pack with:
+        1. invalid readme that contains absolute path. For example:
+            - https://www.example.com/content-assets/example_image.jpg
+        2. invalid description contains relative path that saved not under doc_files. For example:
```
eyalpalo, line 1140:
```suggestion
        2. An invalid description contains a relative path not saved under doc_files. For example:
```
Please edit other descriptions as suggested here.

demisto_sdk/commands/validate/validators/RM_validators/RM108_is_image_path_valid.py
```diff
@@ -115,7 +122,7 @@ def verify_relative_saved_in_doc_files(self, content_item) -> str:
         invalid_links = [
             rel_img
             for rel_img in relative_images
-            if not re.match(DOC_FILE_IMAGE_REGEX, rel_img)
+            if not re.search(DOC_FILE_IMAGE_REGEX, rel_img)
```
eyalpalo, line 125: Why are we changing this?

demisto_sdk/commands/validate/validators/RM_validators/RM108_is_image_path_valid.py
```diff
@@ -72,20 +72,27 @@ def detect_absolute_image_paths(self, content_item) -> str:
...
+                # Extract the correct link and filter out unwanted assets in one step
+                (absolute_link[1] if absolute_link[0] else absolute_link[2])
                 for absolute_link in absolute_links
+                if not (
+                    "content-assets"
+                    in (absolute_link[1] if absolute_link[0] else absolute_link[2])
+                    and (
+                        absolute_link[1] if absolute_link[0] else absolute_link[2]
+                    ).endswith(".gif")
+                )
```
eyalpalo, line 86: Let's refactor this code. It is very unclear.
Also, why store the link in a tuple?

eyalpalo (changes requested):

demisto_sdk/commands/validate/validators/RM_validators/RM108_is_image_path_valid.py
maimorag: Tuples are generally faster and more memory-efficient, plus I don't need to make changes to the items.
https://www.simplilearn.com/difference-between-list-and-tuple-in-python-article#:~:text=Tuples%20are%20generally%20faster%20and,when%20iteration%20speed%20is%20crucial.

demisto_sdk/commands/validate/validators/RM_validators/RM108_is_image_path_valid.py
maimorag: The change from re.match to re.search was necessary because re.match only checks for the pattern at the start of the string, whereas re.search looks for the pattern anywhere in the string. (for example ../../doc_files / ../../../doc_files /doc)

This ensures all occurrences of the pattern are correctly matched, regardless of where they appear in the string.

PR #4684, opened 2024-11-26
CIAC-8160: generate pack readme in marketplace contributions

RotemAmit:
## Related Issues
relates: https://jira-dc.paloaltonetworks.com/browse/CIAC-8160

## Description
Added the option to add content to the pack README of a new pack that was contributed through XSOAR UI.


demisto_sdk/commands/init/contribution_converter.py
```diff
@@ -745,6 +748,10 @@ def create_pack_base_files(self):
         """
         logger.info("Creating pack base files")
         Path(self.working_dir_path, PACKS_README_FILE_NAME).touch()
+        if self.pack_readme:
```
eyalpalo, line 751: Few notes:
1. Please change the docstring to describe the usage when self.pack_readme is not None
2. I'm not sure this is the correct place logic-wise to place this assignment to the readme. Is there any other place we are writing to files created? If so, think about moving this to that place and leave this method as the creating files method.

eyalpalo (changes requested):
Nice!

eyalpalo (approved):

PR #4692, opened 2024-12-02
Exclude silent-Playbooks from metadata

moishce:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12156


demisto_sdk/commands/content_graph/objects/pack_metadata.py
```diff
@@ -188,12 +188,11 @@ def _get_content_items_and_displays_metadata(
         collected_content_items: dict = {}
         content_displays: dict = {}
         for content_item in content_items:
-            if content_item.is_test:
+            if content_item.is_test or content_item.is_silent:
```
ShahafBenYakir, line 191: Let's create a dedicated function: 
```Python
def should_ignore_item_in_metadata(content_item):
    """
    add some docstring explanation
    """
    if content_item.is_test:
        logger.debug(Skiping {content_item.name} in metadata creation: item is test playbook/script.)
    elif content_item.is_silent:
        logger.debug(Skiping {content_item.name} in metadata creation: item is silent playbook/trigger.)
    else:
        return False
    return True
```
```suggestion
            if should_ignore_item_in_metadata:
                continue
```

ShahafBenYakir (approved):
good! small suggestion

PR #4694, opened 2024-12-02
setup-env file_paths fix

Shellyber:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12290

## Description
Fixed the error for missing file_paths

### DIFF
--- a/.changelog/4694.yml
+++ b/.changelog/4694.yml
@@ -1,0 +1,4 @@
+changes:
+- description: Fixed an issue where setup-env would fail when FILE_PATHS argument was missing.
+  type: internal
+pr_number: 4694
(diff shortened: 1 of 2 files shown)


.changelog/4694.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Fixed an issue where setup-env would fail when FILE_PATHS argument was missing.
+  type: internal
```
ShahafBenYakir, line 3: Internal?

ShahafBenYakir (approved):

.changelog/4694.yml
Shellyber: Yes, since it wasn't release and it was caught internally, I don;t think it should go out as a fix.

PR #4710, opened 2024-12-13
fix rn106 and pa114

dantavori:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12384

## Description
- Fixed PA114 and RN106 failures for new packs
- Fixed create_pack_object and its callers to use `currentVersion` instead of `version`


YuvHayun (approved):
Nice!

PR #4718, opened 2024-12-17
remove py3.9

dantavori:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12263


demisto_sdk/__main__.py
```diff
@@ -256,10 +256,11 @@ def main(
     load_dotenv(CONTENT_PATH / ".env", override=True)
     if platform.python_version_tuple()[:2] == ("3", "9"):
         message = typer.style(
-            "Warning: Demisto-SDK will soon stop supporting Python 3.9. Please update your python environment.",
+            "Demisto-SDK supports Python 3.10 and above. Please update your python environment.",
             fg=typer.colors.RED,
         )
         typer.echo(message)
+        raise typer.Exit()
```
dorschw, line 263: We can delete this whole block, python shouldn't allow (unless forced) installation on 3.9 anyway

pyproject.toml
```diff
@@ -102,7 +101,7 @@ pre-commit = "^3.5.0"
 lxml = "^5.0.0"
 uv = "^0.1.5"
 toml = "^0.10.2"
-werkzeug = "^2.2.0" # Added as mitmproxy8 -> flask<2.1
+werkzeug = "^3.0.3"
```
dorschw, line 104: looking at the comment we had before, is `werkzeug` still needed as a dependency? Or can we do without it?

PR #4720, opened 2024-12-19
Exclude silent items from release notes validation

moishce:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12157


dantavori (approved):
let's add a test case in test_IsMissingReleaseNoteEntries

PR #4736, opened 2024-12-24
Fix incident to alert issue /CRTX-145079

RosenbergYehuda:
## Related Issues
relates: https://jira-dc.paloaltonetworks.com/browse/CRTX-145079

## Description
Simplified the `replace_incorrect_marketplace` function to modify the input data directly, addressing the issue where "incident to alert" was not applied. This occurred because a copy of the data was inadvertently created, leading to unexpected behavior in downstream processes.

For the record:  
Here are two versions of the Phishing pack that went through the **prepare content** command, one with and one without my `replace_marketplace_references` function.  
https://drive.google.com/drive/folders/10m6dv0MSBTUpdAFZ1vhUIKnb_D7CVZBs

I compared them using [this tool](https://marketplace.visualstudio.com/items?itemName=L13RARY.l13-diff)

The only change is the expected update to "Cortex XSOAR."  
The "incident to alert" functionality works the same in both versions.
Screenshot 2024-12-25 at 14 49 45

Screenshot 2024-12-25 at 14 48 58

### DIFF
--- a/.changelog/4736.yml
+++ b/.changelog/4736.yml
@@ -1,0 +1,3 @@
+changes:
+- description: Fixed an issue where the "incident to alert" conversion was not applied to the metadata file during the execution of the **prepare-content** command.
+pr_number: 4736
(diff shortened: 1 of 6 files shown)


yaakovpraisler (approved):
Looks good to me!

JasBeilin (approved):
Nice.

PR #4740, opened 2024-12-29
PB130: Added file name check and triggers

moishce:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12153


demisto_sdk/commands/validate/tests/PB_validators_test.py
```diff
@@ -1354,87 +1354,170 @@ def test_MarketplaceKeysHaveDefaultValidator(
...
+        case 3: is_silent = True, trigger_name contain and file_name do not contain silent prefix.
+        case 4: is_silent = False, and trigger_name/file_name contain silent prefix.
+        case 5: is_silent = False, trigger_name contain and file_name do not contain silent prefix.
+        case 6: is_silent = True, and trigger_name/file_name do not contain silent prefix.
+
+    When:
+    - calling IsSilentPlaybookValidator.obtain_invalid_content_items.
+
+    Then:
+    - Checks that it only fails in the right cases.
```
ShahafBenYakir, line 1509: also here let's add what's expected in each case.

ShahafBenYakir (commented):
Nice.
I see that we added the entire validation for both playbooks and triggers inside the PB validations. Is this right? shouldn't we separate?

ShahafBenYakir (approved):

PR #4741, opened 2024-12-29
Rn111 update fail when no git change

JudahSchwartz:
## Related Issues
relates: https://jira-dc.paloaltonetworks.com/browse/CIAC-8051

## Description
Made rn111 fail when you update the docker image when you shouldnt
added fix() method to this validation to override whatever you have in the rn with whats in the yml

intended to be used in the autobump pack flow

Did not add any cacheing/flush mechanism to the rn as i think it would complicate things and theres no indication its needed yet


demisto_sdk/commands/validate/validators/RN_validators/RN111_is_docker_entry_match_yml.py
```diff
@@ -4,60 +4,127 @@
...
         ]
 
-    def get_docker_image_entry(self, rn: str, content_item_name: str) -> str:
-        rn_items = rn.split("##### ")
-        docker = "No docker entry found"
+    def fix(
+        self,
+        content_item: ContentTypes,
+    ) -> FixResult:
+        should_be_rn_entry = release_notes_shouldbe_entry(content_item)
```
YuvHayun, line 87: why not keeping the results from the validation part in a class var and using it here rather than running it again? see IN102 as reference

YuvHayun (approved):
Amazing work!
Minor comments.

demisto_sdk/commands/validate/validators/RN_validators/RN111_is_docker_entry_match_yml.py
JudahSchwartz: i think its easier to just add @cached on this method, wdyt

PR #4743, opened 2024-12-29
Removed Convert Command

Shellyber:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12125

## Description
Removed the deprecated and non-relevant `Convert` command.


samuelFain (commented):
Looks good.
We should definitely add a [changelog](https://github.com/demisto/demisto-sdk/tree/master/.changelog) for this.
I think we should run at least demisto-sdk nightly with this branch to be on the safe side.

Shellyber:
Nightly build - https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/pipelines/1893349

samuelFain (approved):

PR #4751, opened 2025-01-01
update-release-notes --text special chars

inbalapt1:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12202

### DIFF
--- a/demisto_sdk/commands/update_release_notes/update_rn.py
+++ b/demisto_sdk/commands/update_release_notes/update_rn.py
@@ -746,7 +746,7 @@ class UpdateRN:
         :return
         The release notes description
         """
-
+        text = bytes(text, 'utf-8').decode('unicode_escape')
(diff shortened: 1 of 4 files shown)


demisto_sdk/commands/update_release_notes/update_rn.py
```diff
@@ -746,7 +746,7 @@ def build_rn_desc(
         :return
         The release notes description
         """
-
+        text = bytes(text, "utf-8").decode("unicode_escape")
```
omerKarkKatz, line 749: do we need to convert it to bytes will this work?
text.encode("utf-8").decode("unicode_escape")

omerKarkKatz (commented):
Looks good, please see the comment and add a UT.

omerKarkKatz (approved):

PR #4756, opened 2025-01-05
limit poetry version

barryyosi-panw:
## Related Issues
fixes: [CIAC-12544](https://jira-dc.paloaltonetworks.com/browse/CIAC-12544)

## Description

Limiting poetry version in `setup-environment` step - untill v2.0.0 support is added.


mmhw (approved):
Good catch!

PR #4766, opened 2025-01-14
Enhanced BC107  validation

merit-maita:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12402

## Description
Improved implementation of validation BC107 to accept adding toversion field to content item as long as we add a new content item to replace it.


.changelog/4766.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Fixed implementation of `BC107` validation to allow adding `toversion` field to a content item as long as a new content item is added to replace it in the next version.
```
thefrieddan1, line 2: instead of Fixed I would say enhanced or added anility to BC107

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
```diff
@@ -14,30 +16,115 @@
...
 class IsValidToversionOnModifiedValidator(BaseValidator[ContentTypes]):
     error_code = "BC107"
-    description = (
-        "Check that the toversion property was not changed on existing Content files."
-    )
-    rationale = "Changing the `toversion` field for a content item can break backward compatibility."
-    error_message = "Changing the maximal supported version field `toversion` is not allowed. Please undo, or request a force merge."
+    description = "Check that the toversion property is valid."
+    rationale = "Changing the `toversion` field for a content item should include adding a new item to replace it."
+    error_message = "{0}"
```
YuvHayun, line 21: Let's not remove the message.
We should remember that the main purpose of the validation is to ensure no changes has been done to the toversion field.
However, the new issue is to allow these changes in very specific cases.
Therefore the error should be something like this:
"Changing the maximal supported version field `toversion` is not allowed, unless you're adding new content_item with the same id <content_item_id> and their from/to version fulfills the following: <valid_cases>

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
```diff
@@ -14,30 +16,115 @@
 
 class IsValidToversionOnModifiedValidator(BaseValidator[ContentTypes]):
     error_code = "BC107"
-    description = (
-        "Check that the toversion property was not changed on existing Content files."
-    )
-    rationale = "Changing the `toversion` field for a content item can break backward compatibility."
-    error_message = "Changing the maximal supported version field `toversion` is not allowed. Please undo, or request a force merge."
+    description = "Check that the toversion property is valid."
+    rationale = "Changing the `toversion` field for a content item should include adding a new item to replace it."
```
YuvHayun, line 20: revert the change to rationale field

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
```diff
@@ -14,30 +16,115 @@
 
 class IsValidToversionOnModifiedValidator(BaseValidator[ContentTypes]):
     error_code = "BC107"
-    description = (
-        "Check that the toversion property was not changed on existing Content files."
-    )
-    rationale = "Changing the `toversion` field for a content item can break backward compatibility."
-    error_message = "Changing the maximal supported version field `toversion` is not allowed. Please undo, or request a force merge."
+    description = "Check that the toversion property is valid."
```
YuvHayun, line 19:
```suggestion
    description = Check that the toversion property was not changed on existing Content files. unless there's a new file with the same id and continues from_version."
```

demisto_sdk/commands/validate/tests/BC_validators_test.py
```diff
@@ -1026,34 +1031,93 @@ def test_IsContextPathChangedValidator_remove_command():
...
+                }
+            },
+            [],
+            id="Case 2: Modeling Rule - toversion changed to 6.9.0 and the new fromversion is 6.10.0 which is a valid case",
         ),
     ],
 )
 def test_IsValidToversionOnModifiedValidator_obtain_invalid_content_items(
-    content_items, old_content_items
+    mocker, content_items, old_content_items, new_items, expected_errs
```
YuvHayun, line 1120: make sure to updated the given-when-then doc strings.

YuvHayun (changes requested):
Great work, see my comments.

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
```diff
@@ -14,30 +16,111 @@
 
 class IsValidToversionOnModifiedValidator(BaseValidator[ContentTypes]):
     error_code = "BC107"
-    description = (
-        "Check that the toversion property was not changed on existing Content files."
-    )
+    description = "Check that the toversion property was not changed on existing Content files. unless there's a new file with the same id and continues from_version."
     rationale = "Changing the `toversion` field for a content item can break backward compatibility."
-    error_message = "Changing the maximal supported version field `toversion` is not allowed. Please undo, or request a force merge."
+    error_message = "Changing the maximal supported version field `toversion` is not allowed. unless you're adding new content_item with the same id {0} and their from/to version fulfills the following:\n{1}"
```
thefrieddan1, line 21: remove {0}

thefrieddan1 (approved):
Nice. approved with minor changes.

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
```diff
@@ -14,30 +16,111 @@
...
+        (
+            old_to.major == new_from.major
+            and old_from.minor == new_from.minor
+            and old_to.micro + 1 == new_from.micro
+        ),  # old_from 6.0.0, new_to 6.0.1
+    ]
+    is_continuous = any(res)
+    return all(
+        [
+            old_from < new_from,  # old_from 6.0.0, new_from 8.9.0
```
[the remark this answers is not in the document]
merit-maita: yuval asked for them as an example to understand the condition

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
merit-maita: in this string it's replaced with the id, i removed the {0} we saw in the demo

demisto_sdk/commands/validate/validators/GR_validators/GR105_duplicate_content_id.py
```diff
@@ -97,17 +99,43 @@ def obtain_invalid_content_items_using_graph(
...
+                                content_object=content_item,  # type: ignore[arg-type]
+                            )
+                            for duplicate in duplicates
+                        ]
+                    )
+
+        return results
+
+
+def invalid_duplicates(items: list):
```
dantavori, line 126: please check if this is not something that we already do [here](https://github.com/demisto/demisto-sdk/blob/master/demisto_sdk/commands/content_graph/interface/neo4j/queries/validations.py#L316)

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
```diff
@@ -14,30 +16,111 @@
 
 class IsValidToversionOnModifiedValidator(BaseValidator[ContentTypes]):
     error_code = "BC107"
-    description = (
-        "Check that the toversion property was not changed on existing Content files."
-    )
+    description = "Check that the toversion property was not changed on existing Content files. unless there's a new file with the same id and continues from_version."
```
dantavori, line 19:
```suggestion
    description = "Checks that the `toversion` property was not changed on existing content, unless it is replaced by a new file with the same ID and a continuous fromversion."
```

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
```diff
@@ -14,30 +16,111 @@
...
     related_field = "toversion"
-    expected_git_statuses = [GitStatuses.MODIFIED, GitStatuses.RENAMED]
+    expected_git_statuses = [
+        GitStatuses.MODIFIED,
+        GitStatuses.RENAMED,
+        GitStatuses.ADDED,
+    ]
+    valid_cases = (
+        "The old item `fromversion` field should be less than the new item `fromversion` field\n"
+        "The old item `toversion` field should be less than the new item `fromversion` field\n"
```
dantavori, line 30: combined with BA118, the first check is redundant

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
```diff
@@ -14,30 +16,111 @@
...
+    for item in content_items:
+        if item.git_status in [GitStatuses.MODIFIED, GitStatuses.RENAMED]:
+            modified_items.append(item)
+        elif item.git_status == GitStatuses.ADDED:
+            new_items[item.object_id] = {"from": item.fromversion, "to": item.toversion}
+
+    return modified_items, new_items
+
+
+def is_valid_versions(old_from: Version, old_to: Version, new_from: Version) -> bool:
```
dantavori, line 90: please try to have more meaningful/accurate names in your methods - is_valid_versions is very vague, you have to read the code to understand your definition of what is valid or not.
Specifically here, we defined a rule to describe "continuousness of versions" which I'm not sure it's always accurate - for example, what about the jump between XSOAR 6 and 8?

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
dantavori:
```suggestion
    error_message = "Setting a `toversion` field is not allowed, unless replaced by a new content item with the same ID {0}, and their fromversion/toversion fields fulfills the following:\n{1}"
```

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
dantavori: also, where is the check discussed that "the newest item has no toversion"?

demisto_sdk/commands/validate/validators/GR_validators/GR105_duplicate_content_id.py
merit-maita: we dont, it's part of the newly added check.

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
merit-maita: it's not the same in BA118, BA118 if for the same file "toversion" and "fromversion" fields.
my first condition is for the old and new file.
 "the newest item has no toversion" is a case that's not related to my changes in specific, aka we were supposed to already have this validation  (new items has no toversion).
 im discussing it with the relevant side now. we need a new BA validation.

dantavori: if we know old.from < old.to (BA118)
and old.to < new.from (the 2nd condition above)
then old.from < new.from

demisto_sdk/commands/validate/validators/GR_validators/GR105_duplicate_content_id.py
```diff
@@ -97,17 +99,43 @@ def obtain_invalid_content_items_using_graph(
...
+
+    Returns:
+        whether the versions are valid or not.
+    """
+    items.sort(key=lambda item: Version(item.fromversion).release)
+    for i in range(len(items[:-1])):
+        old_toversion = items[i].toversion
+        new_fromversion = items[i + 1].fromversion
+        if Version(old_toversion) >= Version(new_fromversion):
+            return True
```
YuvHayun, line 141:
```suggestion
    items.sort(key=lambda item: Version(item.fromversion).release)
    for i in range(len(items) - 1):
        old_toversion = items[i].toversion
        new_fromversion = items[i + 1].fromversion
        if Version(old_toversion) >= Version(new_fromversion):
            return True
    return False

```

demisto_sdk/commands/validate/validators/GR_validators/GR105_duplicate_content_id.py
```diff
@@ -97,17 +99,43 @@ def obtain_invalid_content_items_using_graph(
...
-            )
-            for content_item, duplicates in self.graph.validate_duplicate_ids(
-                paths_of_content_items_to_validate
-            )
-            for duplicate in duplicates
-        ]
+        results = []
+
+        for content_item, duplicates in self.graph.validate_duplicate_ids(
+            paths_of_content_items_to_validate
```
YuvHayun, line 105: Let's add a test case for this validation.

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
```diff
@@ -14,30 +16,112 @@
...
+    expected_git_statuses = [
+        GitStatuses.MODIFIED,
+        GitStatuses.RENAMED,
+        GitStatuses.ADDED,
+    ]
+    valid_cases = (
+        "The old item `toversion` field should be less than the new item `fromversion` field\n"
+        "The old and the new item should be continuous, aka the old one `toversion` is one version"
+        " less than the new one `fromversion`"
+    )
```
YuvHayun, line 32: Is there a reason why we don't add is as a part of the error_message?
I don't think it is necessary to use a separate variable.

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
```diff
@@ -14,30 +16,112 @@
...
+                ),
                 content_object=content_item,
             )
-            for content_item in content_items
-            if toversion_modified(content_item)
+            for content_item in modified_items
+            if self.invalid_toversion(content_item, new_items)
         ]
 
+    def invalid_toversion(self, modified_item: ContentTypes, new_items: dict):
```
YuvHayun, line 50: doc strings

demisto_sdk/commands/validate/validators/BC_validators/BC107_is_valid_toversion_on_modified.py
```diff
@@ -14,30 +16,112 @@
...
-                message=self.error_message,
+                message=self.error_message.format(
+                    content_item.object_id, self.valid_cases
+                ),
                 content_object=content_item,
             )
-            for content_item in content_items
-            if toversion_modified(content_item)
+            for content_item in modified_items
+            if self.invalid_toversion(content_item, new_items)
```
YuvHayun, line 47:
```suggestion
            if self.invalid_toversion(content_item, new_items)
```
maybe rename function name to invalid_toversion_modification or something similar?

demisto_sdk/commands/validate/tests/BC_validators_test.py
```diff
@@ -1026,54 +1031,104 @@ def test_IsContextPathChangedValidator_remove_command():
...
     result = IsValidToversionOnModifiedValidator().obtain_invalid_content_items(
         content_items
     )
 
-    assert (
-        len(result) == 1
-        and result[0].message
-        == "Changing the maximal supported version field `toversion` is not allowed. Please undo, or request a force merge."
-    )
+    assert len(result) == errors
```
YuvHayun, line 1131: let's also assert the error message.

demisto_sdk/commands/validate/tests/BC_validators_test.py
```diff
@@ -1026,54 +1031,104 @@ def test_IsContextPathChangedValidator_remove_command():
...
+        - Case 1: The obtain_invalid_content_items function will catch the change in `toversion` and will fail the validation since they are more than one release apart.
+        - Case 2: The obtain_invalid_content_items function will catch the change in `toversion` and will fail the validation since two items with the same id ,cannot exist in the same version.
+        - Case 3: The obtain_invalid_content_items function will catch the change in `toversion` and will fail the validation since no new item to replace it was found.
+        - Case 4: Valid case.
     """
     create_old_file_pointers(content_items, old_content_items)
+    mocker.patch(
+        "demisto_sdk.commands.validate.validators.BC_validators.BC107_is_valid_toversion_on_modified.sort_content_items",
+        return_value=(content_items, new_items),
+    )
```
YuvHayun, line 1126: why do we need this mock?
Let's use the git statuses instead.

YuvHayun (changes requested):
Nice!
See my comments.

demisto_sdk/commands/validate/tests/BC_validators_test.py
```diff
@@ -1026,54 +1031,118 @@ def test_IsContextPathChangedValidator_remove_command():
...
+                    paths=["fromversion", "toversion"],
+                    values=["6.10.0", DEFAULT_CONTENT_ITEM_TO_VERSION],
+                )
+            ],
+            # {
+            #     "TestIntegration": {
+            #         "from": "6.10.0",
+            #         "to": DEFAULT_CONTENT_ITEM_TO_VERSION,
+            #     }
+            # },
```
YuvHayun, line 1056: ?

demisto_sdk/commands/validate/tests/BC_validators_test.py
```diff
@@ -1026,54 +1031,118 @@ def test_IsContextPathChangedValidator_remove_command():
...
-        and result[0].message
-        == "Changing the maximal supported version field `toversion` is not allowed. Please undo, or request a force merge."
-    )
+    assert len(result) == errors
+    if result:
+        for res in result:
+            assert (
+                "Changing the maximal supported version field `toversion` is not allowed"
+                in res.message
+            )
```
YuvHayun, line 1145: The point of the assertion is to assert the variables that are formatted into the message to ensure the errors were thrown for the right items.
Please assert the whole message including the object_id.

YuvHayun (approved):
Great work!
approved with comments.
Also, let's change the name of the PR since GR105 is not being modified here.

PR #4773, opened 2025-01-19
Add forcemergeable, igrnoable, and must be handeled errors summary at…

YuvHayun:
## Related Issues
fixes:


demisto_sdk/commands/validate/validation_results.py
```diff
@@ -43,9 +44,34 @@ def __init__(
...
+            validation_results (List[ValidationResult]): The list of ValidationResult objects to summarize.
+
+        Returns:
+            Tuple[int, Set[str]]: The exit code and set of failing error codes.
+        """
+        exit_code = 0
+        failing_error_codes: Set[str] = set()
+        for result in validation_results:
+            if result.validator.error_code in config_file_content.warning:
+                logger.warning(f"<yellow>{result.format_readable_message}</yellow>")
```
JudahSchwartz, line 65: why are these only returned to logger? If calling by python code is this not returned?

JudahSchwartz (approved):
nice!

demisto_sdk/commands/validate/validation_results.py
YuvHayun: We return the exit code but we print out the errors.
This is how we designed it in the first place.

PR #4781, opened 2025-01-22
fix rn111 displayname

JudahSchwartz:
fixed an issue where used name not displayname


sapirshuker (approved):
Approved, the get_name function from the update RN command return the display.name of the content item.

PR #4782, opened 2025-01-22
Align validate with the new PB format

jlevypaloalto:
## Related Issues
fixes:

### DIFF
--- a/demisto_sdk/commands/validate/validators/RN_validators/RN114_validate_release_notes_header.py
+++ b/demisto_sdk/commands/validate/validators/RN_validators/RN114_validate_release_notes_header.py
@@ -119,10 +119,7 @@ class ReleaseNoteHeaderValidator(BaseValidator[ContentTypes]):
                         if ("New: " not in content_type_section[0])
                         else content_type_section[0].removeprefix("New: ")
                     )
-                    if (
-                        header not in PB_RELEASE_NOTES_FORMAT
-                    ):  # remove PB custom headers
-                        headers.setdefault(content_type, []).append(header)
+                    headers.setdefault(content_type, []).append(header)
         return headers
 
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/validate/validators/RN_validators/RN114_validate_release_notes_header.py
```diff
@@ -116,9 +117,10 @@ def extract_rn_headers(self, release_note_content: str) -> Dict[str, List[str]]:
                     header = (
                         content_type_section[0].rstrip()
                         if ("New: " not in content_type_section[0])
-                        else content_type_section[0][len("New: ") :]
+                        else content_type_section[0].removeprefix("New: ")
                     )
-                    headers.setdefault(content_type, []).append(header)
+                    if header not in PB_RELEASE_NOTES_FORMAT:  # remove PB custom headers
+                        headers.setdefault(content_type, []).append(header)
```
AradCarmi: Let's move the if section outside of this function scope as the main purpose is to extract the headers and not to filter them. Please open a new function that will filter the headers and add the filters we talked about in the `validate_release_notes_headers`

demisto_sdk/commands/common/constants.py
```diff
@@ -2218,3 +2218,13 @@ class PlaybookTaskType(StrEnum):
 # Test types:
 TEST_PLAYBOOKS = "TestPlaybooks"
 TEST_MODELING_RULES = "TestModelingRules"
+
+PB_RELEASE_NOTES_FORMAT = {
```
AradCarmi, line 2222: Nice

AradCarmi (changes requested):
Great job!
Please fix my only comment and also be sure to test it. Also, add a UT for this case.

AradCarmi (approved):

PR #4655, opened 2024-11-11
Reintroduce demisto-sdk to the new docker registry proxy

samuelFain:
## Description
Reintroduce docker registry proxy support to demisto-sdk after a previous revert.
Merged the original PR https://github.com/demisto/demisto-sdk/pull/4528 to this one.
Infra MR number: 584
fixes: [CIAC-11589](https://jira-dc.paloaltonetworks.com/browse/CIAC-11589)

PR #4783, opened 2025-01-26
[Docker proxy] Add nanoseconds to microseconds datetime conversion in `dockerhub_client.py`

samuelFain:
## Description
- Fix an issue related to https://github.com/demisto/demisto-sdk/pull/4655 where ISO 8601 datetime strings with nanosecond precision (nine digits after the decimal point) failed to convert to datetime (not supported by Python’s datetime library). Implemented a mechanism to catch case and attempt converting the datetime string to microseconds (six digits after the decimal point), which is supported by the datetime library.


.changelog/4783.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+  - description: Fixed and issue in the Docker proxy mechanism where ISO 8601 timestamps with nanosecond precision were not properly parsed.
```
samuelFain, line 2: Fix typo
```suggestion
  - description: Fixed an issue in the Docker proxy mechanism where ISO 8601 timestamps with nanosecond precision were not properly parsed.
```

demisto_sdk/commands/common/docker/dockerhub_client.py
```diff
@@ -629,3 +639,26 @@ def get_gcloud_access_token() -> Optional[str]:
...
+
+    Returns:
+        str: A ISO 8601 datetime string in the format %Y-%m-%dT%H:%M:%S.%fZ with microseconds precision.
+
+    Details:
+    Python's datetime module supports up to microseconds precision (six decimal places in the seconds fraction (e.g., 0.123456 seconds).
+    While iso8601 can support up to nanoseconds precision (nine decimal places in the seconds fraction (e.g., 0.123456789 seconds)).
+    """
+
+    datetime_until_decimal_point, nanoseconds_with_z = iso8601_datetime_str.split(".")
```
kobymeir, line 659: - no unit tests.
- if you don't have a `.` in the input you will fail.
- if you have a `z` and not `Z` you will fail.
- if you don't have a `z` in the input, you add it at the output, why?

kobymeir (changes requested):

demisto_sdk/commands/common/docker/dockerhub_client.py
samuelFain: - I'll add a UT.
- I'll add checking for `.` as its possible that a timestamp won't have a `.`
- As for the remaining use cases mentioned above - all of them will raise an error in the existing, pre-change, logic that has been functioning until now. We rely on responses from the DockerHub API, which usually adhere to the ISO 8601 format. However, there’s no formal time format specified in DockerHub’s [documentation](https://docs.docker.com/reference/api/hub/latest/).

samuelFain: Also added a "no decimal point" datetime string handling.

demisto_sdk/commands/common/docker/tests/dockerhub_client_test.py
```diff
@@ -285,3 +288,23 @@ def test_do_docker_hub_get_request_single_object(
...
 
     assert dockerhub_client.do_docker_hub_get_request("/test") == {"test": "test"}
+
+
+@pytest.mark.parametrize(
+    "datetime_str, response",
+    [
+        ("2024-11-19T12:41:58.591749197Z", "2024-11-19T12:41:58.591749Z"),
+        ("2024-11-19T12:41:58.591749Z", "2024-11-19T12:41:58.591749Z"),
+        ("2024-11-19T12:41:58Z", "2024-11-19T12:41:58Z"),
```
kobymeir, line 298: example without `Z`

demisto_sdk/commands/common/docker/dockerhub_client.py
```diff
@@ -437,7 +437,16 @@ def get_docker_image_tag_creation_date(
...
+            try:
+                return datetime.strptime(creation_date, "%Y-%m-%dT%H:%M:%S.%fZ")
+            except ValueError as e:
+                logger.debug(
+                    f"Failed initial parsing attempt for '{creation_date=}', docker image {docker_image}:{tag}. Exception: {e}"
+                )
+                logger.debug("Attempting to normalize timestamp and retry parsing.")
+                return datetime.strptime(
+                    iso8601_to_datetime_str(creation_date), "%Y-%m-%dT%H:%M:%S.%fZ"
+                )
```
kobymeir, line 449:
```suggestion
                return datetime.strptime(
                    iso8601_to_datetime_str(creation_date), "%Y-%m-%dT%H:%M:%S.%fZ"
                )
```
would work just as well 🤷 it's not like someone will ever view the logs errors.

kobymeir (approved):

PR #4786, opened 2025-01-27
Allow setting dependency min version

yaakovpraisler:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12706


demisto_sdk/commands/content_graph/objects/pack_metadata.py
```diff
@@ -242,7 +242,8 @@ def _enhance_dependencies(
         return {
             r.content_item_to.object_id: {
                 "mandatory": r.mandatorily,
-                "minVersion": r.content_item_to.current_version,  # type:ignore[attr-defined]
+                # Get the minVersion either from the pack_metadata or from graph calculation
```
ShahafBenYakir, line 245:
```suggestion
                # Get the minVersion either from the pack_metadata if exists, or from graph calculation
```

demisto_sdk/commands/content_graph/tests/pack_metadata_graph_test.py
```diff
@@ -292,3 +292,42 @@ def test_pack_metadata_marketplacev2(
...
+
+
+def test_pack_metadata_dependencies(
+    git_repo: Repo, tmp_path: Path, mocker, monkeypatch
+):
+    """
+    Given:
+        - A repository with two packs, TestPack and TestDependencyPack.
+    When:
+        - In TestPack's pack_metadata is mentioned that it is dependent mandatorily on TestDependencyPack with version 1.2.3 .
```
ShahafBenYakir, line 304: The flow does not consider a case where a pack actually uses an item from TestDependencyPack. This should have created a dependency version higher than 1.2.3, but since we manually added 1.2.3, it will use it.

demisto_sdk/commands/content_graph/interface/neo4j/queries/relationships.py
```diff
@@ -153,6 +153,7 @@ def build_depends_on_relationships_query() -> str:
 // Create the relationship, and mark as "from_metadata"
 CREATE (p1)-[r:{RelationshipType.DEPENDS_ON}{{
     mandatorily: rel_data.mandatorily,
+    target_version: rel_data.target_version,
```
ShahafBenYakir, line 156: So it is part of the relationship?

demisto_sdk/commands/content_graph/tests/pack_metadata_graph_test.py
yaakovpraisler: This is not matter whether it uses a content item from the dependency pack or not for the min version, the point here is that in the pack_metadata of TestDependencyPack there is a different current_version than what is mentioned in the TestPack's dependencies minVersion, and yet the resulted metadata has the minVersion from the TestPack's pack_metadata.

demisto_sdk/commands/content_graph/interface/neo4j/queries/relationships.py
yaakovpraisler: If I understand correctly your question, then yes, and now, when we add the dependencies for the metadata we'll have the version of the target pack.

demisto_sdk/commands/content_graph/tests/pack_metadata_graph_test.py
ShahafBenYakir: where I can see the difference? cause in demisto_sdk/commands/content_graph/tests/test_data/pack_metadata.json I see version 1.2.3

yaakovpraisler: It creates a generic pack from TestUtils which sets a current_version 1.0.0.
Screenshot 2025-01-30 at 10 32 19

demisto_sdk/commands/content_graph/interface/neo4j/queries/relationships.py
ShahafBenYakir: Doesn't this contradict your concern that this is not part of relationship?

yaakovpraisler: Maybe it was a misunderstood.
It has to be as part of the relationship, but It also need to be in the relationship in the graph and not only to the Pack object, this is why I added it here when we create the relationship in the graph.

demisto_sdk/commands/content_graph/tests/pack_metadata_graph_test.py
ShahafBenYakir: After we create it let's explicitly change that pack current version to 2.0.0. It will highlight the purpose of the feature and test (which is to manually fix the dependency to 1.2.3)

ShahafBenYakir (approved):

RosenbergYehuda (approved):
Looks good to me!
Just one small semantic point:
Why is the relationship attribute called target_version instead of target_min_version, like in all other places?

yaakovpraisler:
> Just one small semantic point: Why is the relationship attribute called target_version instead of target_min_version, like in all other places?
This is not really matter, but in the graph you don't have multiple versions of a pack like in the repo/marketplace, so it is not really necessary to call it min, the minVersion is a term we're using for the dependencies which there we have multiple versions for a pack.

PR #4787, opened 2025-01-27
Graph creation - update USES relationships for content items that contains alert in the name

adi88d:
## Related Issues
fixes:
https://jira-dc.paloaltonetworks.com/browse/CIAC-7711
https://jira-dc.paloaltonetworks.com/browse/CIAC-11954

## Description
Added a new query to run in the end of `create_relationships` to update **USES** relationships that meet the following conditions:
- Source node contains `alert` in the `object_id`.
- Source node `not_in_repository=true`.
- Source node is in **marketplacev2** but not in **xsoar** marketplaces.

We updated the target node that meet the following conditions:
- Target **object_id** with `incident` instead `alert` (for example **searchAlerts** in XSIAM -> **searchIncidents** in xsoar).
- Target is in repository.
- The content type is similar to the old target.
- Target node is in **marketplacev2** and in **xsoar** marketplaces.

The `update_alert_to_incident_relationships` method replaced the target in the relationship when all of these conditions are met.


.changelog/4787.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Fixed an issue in graph creation where the content items in XSIAM replaced the word 'alert' with 'incident' on upload.
```
RosenbergYehuda, line 2: Fixed an issue where, when creating relationships between nodes, if the target is not in the repository due to a naming convention (e.g., "incident" to "alert"), the incorrect relationship is deleted and a correct one is created.

demisto_sdk/commands/content_graph/interface/neo4j/queries/relationships.py
```diff
@@ -109,6 +109,41 @@ def build_uses_relationships_query(
 RETURN count(r) AS relationships_merged"""
 
 
+def update_alert_to_incident_relationships():
+    return f"""
+// Updated USES relationships between nodes when the source contains "alert" in the id,
```
RosenbergYehuda, line 114: // Update USES relationships when the source node contains "alert" in its ID.
// This query addresses a situation where some items used in relationships have names that 
// do not match exactly with the repository names. For example, items with "incident" in 
// their names might be uploaded as "alert" in some marketplaces, which causes a false "not_in_repository" flag.
// The query ensures that the target node is replaced with the correct item in the repository,

demisto_sdk/commands/content_graph/interface/neo4j/queries/relationships.py
```diff
@@ -109,6 +109,41 @@ def build_uses_relationships_query(
...
+    AND 'xsoar' IN target_incident.marketplaces
+WITH target_incident,target, source, r
+
+
+// create a relationship with the new incident node
+CREATE (source)-[r_new:{RelationshipType.USES}{{mandatorily: r.mandatorily}}]->(target_incident)
+
+
+// delete the old target node and old relationship
+DELETE r
```
RosenbergYehuda, line 142: Isn't deleting the target enough?

RosenbergYehuda (commented):
Very Nice!
Please see my comments.
In addition, I wonder if chat gpt can improve the query to look better...

demisto_sdk/commands/content_graph/interface/neo4j/queries/relationships.py
adi88d: No, it failed and ask to delete the relationship before

adi88d:
> Very Nice! Please see my comments. In addition, I wonder if chat gpt can improve the query to look better...

what I wrote in the code its after chat gpt improvements

.changelog/4787.yml
adi88d: done

demisto_sdk/commands/content_graph/interface/neo4j/queries/relationships.py
adi88d: updated

RosenbergYehuda (approved):
Very nice!

dantavori (approved):
looks great!
approving, but let's just add a small test for GR103 in GR_validators_test:
* create pack A that includes script S (XSIAM only) with code=`demisto.executeCommand("SearchAlerts", {})`
* create pack B with a script named `SearchIncidents` (XSIAM + XSOAR).
* create a graph, run GR103, and ensure it does not return a validation error.

PR #4793, opened 2025-02-02
fixed PB130 PB131 validtions

moishce:
## Related Issues
fixes:

## Description
PB130: removed Triggers from validation.
PB131: changed to use the graph because the related items are not always available in the content item object.


.changelog/4793.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Fixed an issue where PB130 was running on unnecessary files, and PB131 was not validating correctly in certain cases.
```
YuvHayun, line 2: let's separate to two different entries

demisto_sdk/commands/validate/validators/PB_validators/PB130_is_silent_playbook.py
```diff
@@ -3,13 +3,12 @@
...
 
 from demisto_sdk.commands.content_graph.objects.playbook import Playbook
-from demisto_sdk.commands.content_graph.objects.trigger import Trigger
 from demisto_sdk.commands.validate.validators.base_validator import (
     BaseValidator,
     ValidationResult,
 )
 
-ContentTypes = Union[Playbook, Trigger]
+ContentTypes = Union[Playbook]
```
YuvHayun, line 11:
```suggestion
ContentTypes = Playbook
```

YuvHayun (approved):
Looks good, approved with minor comments.
Also, let's make sure to move the error_codes from warning to select in the sdk_validation_config file.

PR #4807, opened 2025-02-09
Add more functionalities to test playbooks usecases

merit-maita:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12474

## Description
Added some functionalities to support testing playbooks use cases.


demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
```diff
@@ -618,6 +620,42 @@ def get_integration_instance(self, instance_name: str):
...
+    def disable_integration_instance(self, instance_name):
+        return self.update_integration_instance_state(False, instance_name)
+
+    def enable_integration_instance(self, instance_name):
+        return self.update_integration_instance_state(True, instance_name)
+
+    def update_integration_instance_state(self, enable: bool, instance_name: str):
+
+        # it will throw an error if the instance does not exist
+        instance = self.get_integration_instance(instance_name)
```
thefrieddan1, line 632: consider try except if it throws an error

demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
```diff
@@ -1343,3 +1421,134 @@ def update_playbook_input(self, playbook_id: str, new_inputs: dict):
...
+            raise ValueError(f'task_states are bad. Possible values: {self.PLAYBOOK_TASKS_STATES}')
+        if not task_states:
+            task_states = self.PLAYBOOK_TASKS_STATES
+        if complete_task and not task_input:
+            raise RuntimeError("Task input argument is missing to complete tasks.")
+        completed_tasks = []
+        found_tasks = []
+        start_time = time.time()
+
+        while True:
```
thefrieddan1, line 1455: what causes this while loop to exit?

demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
```diff
@@ -1343,3 +1421,134 @@ def update_playbook_input(self, playbook_id: str, new_inputs: dict):
...
+            # Get all tasks with one state of the states in task_states list
+            tasks_by_states, status_code, _ = self._xsoar_client.generic_request(
```
thefrieddan1, line 1457: max_time is 60. is that seconds?
the loop will create multiple calls to this API 
I think we need to think about different solutions

demisto_sdk/commands/common/clients/xsoar/xsoar_api_client.py
merit-maita: its already handled in the get_integration_instance method itself

thefrieddan1 (approved):

PR #4810, opened 2025-02-10
Add test to ensure no validation run only on path_based

YuvHayun:
## Related Issues
fixes:

## Description
Add a test to ensure that all validations that run on -a also run on -g to ensure no master-breaking validations occur.


AradCarmi (approved):
Nice!

PR #4811, opened 2025-02-10
Add support run cmd to xsoar 8

israelpoli:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-46598

## Description
Add support run cmd to xsoar 8
Due to the restriction on using `playground-id` when executing commands via the API in XSOAR 8, the command has been implemented so that at the start of execution, an incident is created, and the command runs within it. At the end of execution, the process will either close, remove, or leave the incident open based on the user's choice using the new argument `--incident-post-run` (`-ipr`).  

By default, the process will remove the incident. The allowed options for this argument are: `open`, `close`, `remove`.


demisto_sdk/commands/run_cmd/runner.py
```diff
@@ -312,3 +339,97 @@ def execute_command(self, command: str):
...
+
+    def get_server_type(self) -> ServerType:
+        if not self.client.api_client.configuration.host.startswith("https://api-"):
+            return ServerType.XSOAR
+        raw_response, _, response_headers = self.client.generic_request(
+            "/about", "GET", response_type="object"
+        )
+        if "text/html" in response_headers.get("Content-Type"):
+            raise ValueError(
+                "",  # TODO
```
samuelFain, line 351: We should raise a more informative message here, and remove the #TODO comment :)

demisto_sdk/commands/run_cmd/runner.py
```diff
@@ -312,3 +339,97 @@ def execute_command(self, command: str):
...
+            content_type="application/json",
+            body=body,
+            response_type=object,
+        )
+
+        if status_code != 200:
+            logger.info(
+                f"Failed {self.incident_post_run.value}d the incident {self.incident_id}"
+            )
+        logger.info(
```
samuelFain, line 433: Can we add the raw response from the `client.generic_request` in the log in case of a non-200 response? It could help debugging in the future.

demisto_sdk/commands/run_cmd/runner.py
```diff
@@ -312,3 +339,97 @@ def execute_command(self, command: str):
...
+                "type": "Unclassified",
+                "playbook": "",
+            },
+            response_type=object,
+        )
+
+        if status_code != 201:
+            logger.critical(
+                "Failed to create a new incident for executing the command."
+            )
```
samuelFain, line 391: Can we add the raw response from the `client.generic_request` in the log in case of a non-201 response? It could help with debugging in the future.

samuelFain (changes requested):
Looks good!
I have some minor comment about docstrings and logging. Please have a look.
- I also updated the `run` command confluence page about XSOAR 8.X support. @israelpoli Feel free to add relevant development information about this effort.
- We should also check if the https://docs-cortex.paloaltonetworks.com/r/1/Demisto-SDK-Guide/run documentation needs to updated, and if so, we should reach out to @ShirleyDenkberg about it.

PR #4813, opened 2025-02-11
Add config reader validations

YuvHayun:
## Related Issues
fixes:


demisto_sdk/commands/validate/validators/base_validator.py
```diff
@@ -296,7 +297,7 @@ class FixResult(BaseResult, BaseModel):
     """This is a class for fix results."""
 
 
-class InvalidContentItemResult(BaseResult, BaseModel):
+class GeneralValidationResult(BaseResult, BaseModel):
```
JudahSchwartz, line 300: why rename

PR #4822, opened 2025-02-17
[pre-commit] Fix issue with logs

mmhw:
## Description
Fixed an issue with the pre-commit command where debug level logs were not appearing in the log file.


mmhw:
Test with job 15946518

PR #4829, opened 2025-02-23
Fix imported files git status

ilaredo:
## Related Issues
fixes: [CIAC-12482](https://jira-dc.paloaltonetworks.com/browse/CIAC-12482)

## Description
rn108 used to failed due to mismatch classifying git statuses of files.
The fix ensures that the number of fetched files matches the number of files in the contribution_files_relative_paths.txt file. If a discrepancy is found, indicating untracked files, it halt execution due to a mismatch in file counts.

### DIFF
--- a/demisto_sdk/commands/validate/initializer.py
+++ b/demisto_sdk/commands/validate/initializer.py
@@ -265,7 +265,9 @@ class Initializer:
         See CIAC-12482 for more info.
         """
         if os.getenv("CONTRIB_BRANCH"):
-            with open("contribution_files_relative_paths.txt", "r") as contribution_file:
+            with open(
+                "contribution_files_relative_paths.txt", "r"
+            ) as contribution_file:
                 contribution_files_relative_paths_count_lines = sum(
                     1 for line in contribution_file if line.strip()
                 )
--- a/demisto_sdk/commands/validate/initializer.py
+++ b/demisto_sdk/commands/validate/initializer.py
@@ -269,7 +269,6 @@ class Initializer:
             with open(
                 "contribution_files_relative_paths.txt", "r"
             ) as contribution_file:
-                logger.info(f"\n{contribution_file=}")  # TODO: remove line
                 for single_line in contribution_file:
                     clean_line: str = single_line.rstrip("\n")
                     relative_untracked_files_paths.add(Path(clean_line))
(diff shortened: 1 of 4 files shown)


demisto_sdk/commands/validate/initializer.py
```diff
@@ -252,30 +252,36 @@ def get_unfiltered_changed_files_from_git(self) -> Tuple[Set, Set, Set]:
...
-            )
-            # modified_files = modified_files.union(relative_untracked_files_paths)
-            added_files = set(added_files).union(relative_untracked_files_paths)
+                contribution_files_relative_paths_count_lines = sum(
+                    1 for line in contribution_file if line.strip()
+                )
+
+            if contribution_files_relative_paths_count_lines != (
+                len(modified_files) + len(added_files) + len(renamed_files)
+            ):
```
samuelFain, line 277: Is there a possibility that the same file could be in more than one set?
if yes, is could mess up our file count, if no, then we are good.
If this case is possible, maybe we could unify all three sets to eliminate dupes and then measure the length.

samuelFain (approved):
Looks good.
Please see my comments before merging.
Its also important to point out that this branch was tested on several real external contribution PRs and worked as expected.

demisto_sdk/commands/validate/initializer.py
```diff
@@ -252,30 +252,36 @@ def get_unfiltered_changed_files_from_git(self) -> Tuple[Set, Set, Set]:
...
-                    clean_line: str = single_line.rstrip("\n")
-                    relative_untracked_files_paths.add(Path(clean_line))
-            logger.info(
-                f"\n######## - Added untracked:\n{relative_untracked_files_paths}"
-            )
-            # modified_files = modified_files.union(relative_untracked_files_paths)
-            added_files = set(added_files).union(relative_untracked_files_paths)
+                contribution_files_relative_paths_count_lines = sum(
+                    1 for line in contribution_file if line.strip()
+                )
```
eyalpalo, line 273:
```suggestion
                contribution_files_relative_paths_count_lines = len(contribution_file.readlines())
```

Unless there are empty lines

demisto_sdk/commands/validate/initializer.py
```diff
@@ -252,30 +252,36 @@ def get_unfiltered_changed_files_from_git(self) -> Tuple[Set, Set, Set]:
...
+
+            if contribution_files_relative_paths_count_lines != (
+                len(modified_files) + len(added_files) + len(renamed_files)
+            ):
+                logger.info(
+                    "The number of fetched files does not match the number of files in the "
+                    "contribution_files_relative_paths.txt file. This indicates that there are untracked files."
+                )
+                raise ValueError(
+                    "Error: Mismatch in the number of files. Unable to proceed."
```
eyalpalo, line 283: lets give a bit more informative error.

eyalpalo (changes requested):
Nice
Please fill the description of the PR with details on the fix.

eyalpalo (approved):
Approved with 2 things to validate before merging:
1. The file has no empty lines? If there are counting all lines will have a false number.
2. does the lists of modified, added and renamed files are lists or sets? if list please convert to set

PR #4835, opened 2025-02-25
[docker] Increase container timeout

mmhw:
## Related Issues
on-call issue.
## Description
Increase container timeout.


demisto_sdk/commands/common/docker_helper.py
```diff
@@ -49,6 +49,7 @@
 )
 
 TEST_REQUIREMENTS_DIR = Path(__file__).parent.parent / "lint" / "resources"
+DOCKER_CONTAINER_TIMEOUT = 300
```
kobymeir, line 52:
```suggestion
DOCKER_CONTAINER_TIMEOUT = int(os.getenv("DOCKER_CONTAINER_TIMEOUT")) or 300
```
might not be accurate here.

kobymeir (approved):

PR #4839, opened 2025-02-27
Platform MP: update schemas used by old validator

ilappe:
## Related Issues
fixes:https://jira-dc.paloaltonetworks.com/browse/CIAC-12972

## Description
currently, [the build will fail](https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/jobs/16733835/raw#L909) on content items with the `platform` as marketpklace
because old validator used the schemas.


samuelFain (approved):
Looks good!

PR #4849, opened 2025-03-05
Optimize disabling of mirroring settings in IntegrationScriptUnifier to avoid adding new fields

kamalq97:
## Related Issues
fixes:

## Description
Optimize disabling of mirroring in IntegrationScriptUnifier to avoid adding new fields to the Integration YML


kamalq97:
Closing this PR as this optimization is not necessary for the time being

PR #4855, opened 2025-03-12
checking if parent folder exists before iterating on it

omerKarkKatz:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-13074

## Description
When deleting a folder with a file inside it in content.
When we are collecting test for the deleted file it try to iterate over the files in the parent (also deleted folder). 

Failing with file not found error due to the fact the folder does not exists any more.

### DIFF
--- a/demisto_sdk/commands/pre_commit/pre_commit_command.py
+++ b/demisto_sdk/commands/pre_commit/pre_commit_command.py
@@ -684,7 +684,9 @@ def add_related_files(file: Path) -> Set[Path]:
             else PS1_TEST_FILE_SUFFIX
         )
 
-        test_files = [
+test_files = []
+if file.parent.exists():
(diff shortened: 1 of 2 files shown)


demisto_sdk/commands/pre_commit/pre_commit_command.py
```diff
@@ -687,7 +687,7 @@ def add_related_files(file: Path) -> Set[Path]:
         test_files = [
```
barryyosi-panw, line 687: i dont think we need to check if the "file.parent" exists on every iteration.

wdyt of 
```suggestion
test_files = []
if file.parent.exists():
    test_files = [_file for _file in file.parent.iterdir() if _file.name.endswith(test_file_suffix)]
```

barryyosi-panw (approved):

PR #4874, opened 2025-03-20
fix strict objects

barryyosi-panw:
## Description
Adapt strict objects to support new platform marketplace `supportedModules` field.


ilappe (approved):
put here a link to the Nightly you triggred with this changes

PR #4882, opened 2025-03-24
Close instead of deleting a successful incident during a test playbook

MosheEichler:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-12266

## Description
Avoid deleting incidents in build test


demisto_sdk/__main__.py
```diff
@@ -94,7 +94,7 @@ def get_version_info():
 def show_version():
     """Display the SDK version and notify if updates are available."""
     current_version, last_release = get_version_info()
-    typer.echo(f"demisto-sdk version: {current_version}")
+    typer.echo(f"demisto-sdk custom remove-clean-incident version: {current_version}")
```
MosheEichler, line 97: Just for testing will revert back

samuelFain (approved):
Approved.

demisto_sdk/__main__.py
samuelFain: > [!NOTE]
> Don't forget to revert this before merging.

PR #4885, opened 2025-03-30
Fix parsing rule and modeling rule strict objects

barryyosi-panw:
## Description
Adapt strict objects for Modeling Rules and ParsingRules to support new platform marketplace `supportedModules` field.


ilappe (approved):
nice

PR #4899, opened 2025-04-07
Fix NEO4J memory limitation issue

RosenbergYehuda:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-13209

## Description
Since we introduced the graph dashboard, we've been running heavy queries on the Neo4j graph during the 'upload-content-graph-data-to-bigquery' step. This caused us to hit the memory limit allocated to the graph.
In this fix, I'm increasing the memory allocation from the original 2GB to 3GB by adding an additional 1GB.

### DIFF
--- a/.changelog/4899.yml
+++ b/.changelog/4899.yml
@@ -1,0 +1,4 @@
+changes:
+- description: Increased Content Graph memory allocation to 3GB to support heavier queries.
+  type: fix
+pr_number: 4899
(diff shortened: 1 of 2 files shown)


RosenbergYehuda:
Closing this, as we realized it only affects the local graph. This isn’t a major concern since the user can simply delete the container and restart if they hit the limit

PR #4900, opened 2025-04-08
Fixed an issue in the unify when there is a hyphen in the pack id

moishce:
## Related Issues
fixes:

### DIFF
--- a/demisto_sdk/commands/prepare_content/integration_script_unifier.py
+++ b/demisto_sdk/commands/prepare_content/integration_script_unifier.py
@@ -552,7 +552,7 @@ class IntegrationScriptUnifier(Unifier):
             existing_pack_info_debug_log_re = (
                 r"demisto\.debug\('pack name = .*?, pack version = .*?'\)"
             )
-            global_name = f"{pack_id.upper()}_PACK_VERSION"
+            global_name = f"{pack_id.upper().replace('-', '_')}_PACK_VERSION"
             pack_version_statement = (
                 f"{global_name} = '{pack_version}'"
                 f"\ndemisto.debug(f'pack id = {pack_id}, pack version = {{{global_name}}}')"


demisto_sdk/commands/prepare_content/integration_script_unifier.py
```diff
@@ -552,10 +552,11 @@ def insert_pack_version(
             existing_pack_info_debug_log_re = (
                 r"demisto\.debug\('pack name = .*?, pack version = .*?'\)"
             )
-            global_name = f"{pack_id.upper()}_PACK_VERSION"
+            global_name = "PACK_VERSION"
+            pack_id_escaped = pack_id.replace("'", "\\'")
```
ShahafBenYakir, line 556: what is this?

moishce: This is a fix that Dan Tabori added, because if there were a ' in the name it would have caused a bug, and I removed it because I changed it to id

demisto_sdk/commands/prepare_content/integration_script_unifier.py
```diff
@@ -552,10 +552,11 @@ def insert_pack_version(
             existing_pack_info_debug_log_re = (
                 r"demisto\.debug\('pack name = .*?, pack version = .*?'\)"
             )
-            global_name = f"{pack_id.upper()}_PACK_VERSION"
+            global_name = "CONSTANT_PACK_VERSION"
+            pack_id_escaped = pack_id.replace("'", "\\'")
             pack_version_statement = (
                 f"{global_name} = '{pack_version}'"
-                f"\ndemisto.debug(f'pack id = {pack_id}, pack version = {{{global_name}}}')"
+                f"\ndemisto.debug('pack id = {pack_id_escaped}, pack version = {pack_version}')"
```
ShahafBenYakir, line 559: you need to use format here. Also why are you no longer using the global name?

moishce: To avoid problems, I thought it might be better to write straight

ShahafBenYakir (approved):

kobymeir (approved):

PR #4881, opened 2025-03-24
Remove lint command

adi88d:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-10018

PR #4905, opened 2025-04-09
Create dev image: update exec mode of python_image

ilappe:
## Description
in https://github.com/demisto/demisto-sdk/pull/4881
the path `demisto_sdk/commands/lint/resources/installation_scripts/python_image.sh`
was changed to `demisto_sdk/commands/pre_commit/resources/installation_scripts/python_image.sh`
and this cause to change the mode of this file from `-rwxr-xr-x` to `-rw-r--r--`
and this cause the [pre-commit fails](https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/jobs/19519658#L691)


DeanArbel (approved):
Well done

PR #4919, opened 2025-04-27
Contribution Converter small fix

RotemAmit:
## Related Issues
fixes:

## Description
Added a try accept to format_converted_pack function since the action process_pack.py in the contribution management fails with click.exceptions.Exit: 0, after the upgrade of the demisto-sdk version. 
image


demisto_sdk/commands/init/contribution_converter.py
```diff
@@ -338,14 +338,20 @@ def format_converted_pack(self) -> None:
...
+                from_version=from_version,
+                no_validate=True,
+                update_docker=True,
+                assume_answer=True,
+                include_untracked=False,
+                interactive=False,
+            )
+        except Exception as e:
+            if str(e).strip() == "0":
+                logger.info(f"The exit code is {str(e)}")
```
eyalpalo, line 352: Is there other way of getting the exit code? What attributes does e has?

eyalpalo (changes requested):

demisto_sdk/commands/init/contribution_converter.py
RotemAmit: From my checks, e is the exit code number without any additional attributes.

eyalpalo (approved):

PR #4921, opened 2025-04-28
Fix intermittent JSON serialization issue in BaseContent.to_dict() method

shmuel44:
## Related Issues
fixes:

## Description
This PR fixes an intermittent JSON serialization issue in the BaseContent.to_dict() method that was causing validation failures in GitHub Actions. The issue occurred when trying to serialize complex objects like RNRelatedFile that don't have JSON serialization methods defined.

Solution
Changed to_dict() to use self.dict() directly instead of using self.json() followed by json.loads(), avoiding the serialization step that was failing with complex objects.


mmhw (approved):
Nice!

PR #4928, opened 2025-05-06
XSUP-40182/zip-packs-creates-empty-lists

sapirshuker:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-40182


eyalpalo (changes requested):
I'm not sure this is the correct way to implement the test... We have the TestPack which we can use to include this use case as well... Why do we need new pack?

eyalpalo (approved):

PR #4938, opened 2025-05-12
Handle empty URL path for images in README.md during `prepare-content` command

samuelFain:
## Related Issues
fixes:

## Description
Fixed an issue in the `prepare-content` command where wrong handling of the Image URL in the README.md file would cause the image file not to be created due to file naming issues.
Each image in the README.md file in being parsed and named temporarily after the the path value of the URL, in some cases, there are no further paths after the domain, which causes the image file name to be empty string or "/", which are invalid file names, and the image file ends-up not being created.
For example, given the URL: "www.example.test", the original logic would fail:
`Failed downloading the image in url ParseResult(scheme='https', netloc='www.example.test', path='', params='', query='', fragment=''). ........ No such file or directory: ''`
For URL with path values, like: "www.example.test/A/B/some-name", the file name value would be: "some-name".
This solution should generate random path value in such cases, in order for the image to be successfully created.

### DIFF
--- a/.changelog/4938.yml
+++ b/.changelog/4938.yml
@@ -1,0 +1,4 @@
+changes:
+- description: Fixed an issue in the ***prepare-content*** command where wrong handling of the Image URL in the README.md file would cause the image file not to be created.
+  type: internal
+pr_number: 4938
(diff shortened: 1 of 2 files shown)


israelpoli (approved):
Perfect 💪 
please run upload on the dev buckets before merging

PR #4958, opened 2025-05-26
CIAC-13872: Update the regexes related to the pack version and pack id

RotemAmit:
## Related Issues
relates:https://jira-dc.paloaltonetworks.com/browse/CIAC-13872

## Description
1. Updated demisto-sdk split, to catch "CONSTANT_PACK_VERSION" and remove it.
2. Updated demisto-sdk split, to catch the new format of the pack id and pack version and pack name pack version in the integration scripts and remove it.
3. Updated the contribution converter script to extract the pack version from the new format of pack id and pack version.
4. Added more logs to the contribution converter script.


RotemAmit:
A test PR, with the comment https://github.com/demisto/content/pull/40114

demisto_sdk/commands/init/tests/contribution_converter_test.py
```diff
@@ -871,7 +871,11 @@ def test_rearranging_before_conversion_indicator_and_incident_fields(repo: Repo)
     "input_script, output_version",
     [
         (
-            "This is a test script\n the script contains a pack version\n ### pack version: 3.4.5  TEST TEST",
```
eyalpalo, line 874: We should still support this type of pack version... Lets keep that test as well.

eyalpalo (changes requested):
Nice

PR #4969, opened 2025-06-11
init

RosenbergYehuda:
## Related Issues
fixes:


demisto_sdk/tests/test_files/conf.json
```diff
@@ -46,6 +46,9 @@
     "nightly_packs": [
         "CommonScripts"
     ],
+    "selected_nightly_tpb": [
+        "test test"
```
JasBeilin, line 50: what is this used for?

RosenbergYehuda: Now this field is required, so i need to add it to old tests (in the test data)  in order to pass

PR #4983, opened 2025-06-26
[Validate] GR107 Bi-directional validation

samuelFain:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-13659

## Description
Currently, GR107 only checks the following case:
```
Given validated item A, check if it uses any deprecated items.
```

GR107 will now additionally check a second case where:
```
Given validated item A has been deprecated, check if any items are using it.
```

### Additional information
- GR107 is set to run on deprecated items (validations default is usually False).
- GR107 will fail the `validate` command, but will be ignorable via `.pack-gnore`.
- GR107 will not run in nightly mode.
- When item A uses deprecated item B, and item A ignores `GR107` in its `.pack-ignore` file, the validation will not fail the relationship between A and B during validation of either A or B items.
  - To incorporate this logic, a closed list of validation error codes will be saved in `ALWAYS_RUN_ON_ERROR_CODE` constant to addresse unique cases where validation must always run first, then filter the relevant results afterward based on each validated content item's ignore list.

---
### Exmaple
Given `content_item` that uses `deprecated_content_item`.

GR107 can now output:
```
Packs/Path/content_item.yml: [GR107] - The item 'content_item' is using the following deprecated items: <deprecated_content_item_id>
Validate summary
The following errors were thrown as a part of this pr: GR107.
The following errors can be ignored: GR107.
The following errors don't run as part of the nightly flow and therefore can be force merged: GR107.
```
For both cases, when we run `demisto-sdk validate -i Packs/Path/content_item.yml` **OR** `demisto-sdk validate -i Packs/Path/deprecated_content_item.yml` (The second case was not validated until now)


demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -90,6 +93,10 @@ def run_validations(self) -> int:
                         in filtered_content_objects_for_validator
                     ]
                 try:
+                    validation_results = self.filter_validation_results(
```
YuvHayun, line 96: I think maybe it would be more efficient to check the the error_code is part of the constant from here, before calling filter_validation_results

demisto_sdk/commands/validate/validators/GR_validators/GR107_is_deprecated_content_item_in_usage.py
```diff
@@ -121,3 +122,5 @@ def obtain_invalid_content_items_using_graph(
             )
             for item_using_deprecated, data in grouped_results.items()
         ]
+
+    # def relevant_item_relationships(self, item_using_deprecated_ignored_errors, ):
```
YuvHayun, line 126: ?

demisto_sdk/commands/validate/validators/base_validator.py
```diff
@@ -269,7 +273,7 @@ def is_error_ignored(
     Returns:
         bool: True if the given error code should and allow to be ignored by the given item. Otherwise, return False.
```
YuvHayun, line 274: update docstrigns.

demisto_sdk/commands/validate/tests/GR_validators_test.py
```diff
@@ -804,7 +804,7 @@ def test_GR107_IsDeprecatedContentItemInUsageValidatorListFiles_valid_playbook(
         pack_objects
     )
 
-    assert len(validation_results) == 0
+    assert len(validation_results) == 1
```
YuvHayun, line 807: Let's make sure we have all the cases we talked about to ensure we're covered now, and fore the future:
a deprecated and being collected -> b is using a
a deprecated and being collected -> b is deprecated using a
a not deprecated being collected -> b is used by a, b is deprecated
a not deprecated being collected -> b is used by a, b is not deprecated
a is being deprecated -> b uses a, b is using c, c is deprecated.
a is being not deprecated -> b uses a, b is using c, c is deprecated.

YuvHayun (changes requested):
Great job!
Left some comments.

demisto_sdk/commands/validate/validators/GR_validators/GR107_is_deprecated_content_item_in_usage.py
samuelFain: Oops, deleted :)

demisto_sdk/commands/validate/validators/base_validator.py
samuelFain: Done.

demisto_sdk/commands/validate/tests/GR_validators_test.py
samuelFain: Done.

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -90,6 +93,17 @@ def run_validations(self) -> int:
...
+                    if any(
+                        error_code in ALWAYS_RUN_ON_ERROR_CODE
+                        for error_code in {
+                            validator.error_code for validator in self.validators
+                        }
+                    ):
+                        validation_results = self.filter_validation_results(
+                            validation_results
+                        )
+
```
YuvHayun, line 106: We are running in the scope of the specific validation, why do we need to check for every error code?

demisto_sdk/commands/validate/validate_manager.py
```diff
@@ -155,3 +169,23 @@ def add_invalid_content_items(self):
...
+        Returns:
+        List[ValidationResult]: Filtered validation results excluding ignored error codes
+        """
+        return [
+            result
+            for result in validation_results
+            if not (
+                result.validator.error_code in ALWAYS_RUN_ON_ERROR_CODE
+                and result.validator.error_code in result.content_object.ignored_errors
```
YuvHayun, line 189: why are we checking the first condition? if you got here it means we're already in an error code that is in ALWAYS_RUN_ON_ERROR_CODE isn't it?

YuvHayun (approved):
Looks good! Minor clarifications.

demisto_sdk/commands/validate/validate_manager.py
samuelFain: This is a change to https://github.com/demisto/demisto-sdk/pull/4983#discussion_r2179346526, where the idea was to first check if any of the validators are in ALWAYS_RUN_ON_ERROR_CODE before filtering the results.
This is due to the uniqe situation of GR107 and any future validations that need to run first, and filter its results later.

demisto_sdk/commands/validate/validate_manager.py
samuelFain: I think these are similar looking but different checks.
The initial check in the if statement:
```python
if any(
  error_code in ALWAYS_RUN_ON_ERROR_CODE
  for error_code in {
    validator.error_code for validator in self.validators
  }
):
```
Is just an initial indication that **at least** one validator has an error code in `ALWAYS_RUN_ON_ERROR_CODE`.
Since this code is in the main `run_validation` functionality, other validators can run along side GR107.

The code you mentioned in the `filter_validation_results` function actually goes over every `ValidationResult` object and filters according according to what is needed, and `result.validator.error_code in ALWAYS_RUN_ON_ERROR_CODE` is one of the necessary filtering conditions.

demisto_sdk/commands/validate/validate_manager.py
samuelFain: Fixed.

demisto_sdk/commands/validate/validate_manager.py
samuelFain: Fixed.

PR #4988, opened 2025-07-02
Fix handling of pack support level in xsoar-lint command

kamalq97:
## Related Issues
fixes: [link to issue](https://jira-dc.paloaltonetworks.com/browse/XSUP-52559)

## Description
Fixed an issue where the xsoar-lint command would not run if the pack support level was set to developer.


samuelFain (approved):
Nice one!

PR #4953, opened 2025-05-22
[feature] Preference Center

samuelFain:
## Related Issues
relates: https://jira-dc.paloaltonetworks.com/browse/CIAC-13824

## Description
SDK branch for the Preference Center feature support.
- Add schema support for the `compliantpolicies` yml key.
- Add graph support for the `compliantpolicies` yml key.

PR #5014, opened 2025-07-23
Fixed Layouts Schemas

ShacharKidor:
## Related Issues
fixes: [XSUP-53187](https://jira-dc.paloaltonetworks.com/browse/XSUP-53187).

## Description

- As described in this ticket - [XSUP-53187](https://jira-dc.paloaltonetworks.com/browse/XSUP-53187), the ST110 validation is failing for `layouts` that include `display filters` at the layout `section` level. 
It appears that the layout schemas allowed display filters only for the `section field` level, but not for the `section` level. 

- Added the filters schema to:
1. The Graph - strict objects layout schema (used in the validate command). 
2. The old layout container schema (used by the format command).

### DIFF
--- a/.changelog/5014.yml
+++ b/.changelog/5014.yml
@@ -1,4 +1,4 @@
 changes:
-- description: Fixed an issue where `display filters` were missing in layout containers' schemas.
+- description: Fixed an issue where `display filters` were missing in layout container's schemas.
   type: fix
 pr_number: 5014
--- a/.changelog/5014.yml
+++ b/.changelog/5014.yml
@@ -1,0 +1,4 @@
+changes:
+- description: Fixed an issue where `display filters` were missing in layout containers' schemas.
+  type: fix
+pr_number: 5014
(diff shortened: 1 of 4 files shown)


jbabazadeh (approved):
Great work

samuelFain:
@ShacharKidor Nice!
Updating an object schema in `strict_objects` typically necessitates a corresponding update for the related object's `parser`, in this case its a Layout object.
Meaning a change in `demisto_sdk/commands/content_graph/strict_objects/layout.py` should have a matching change in `demisto_sdk/commands/content_graph/parsers/layout.py`.
See for exmaple my recent PR where I added a new field for `integration` and `script` object schemas - https://github.com/demisto/demisto-sdk/pull/4953.
I'm not sure if this a must-have change, but it would be wise to verify that before we merge.

ShacharKidor:
Hi @samuelFain, 
As discussed offline with @RosenbergYehuda the Graph doesn't use the strict object (they are used only for validations).

 To be on the safe side - I triggered SDK nightly on this branch: 
 [link to pipeline](https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/pipelines/4314579).
 
 If it passes, I'll merge the PR.

PR #5024, opened 2025-08-06
Fix doc review with always true to ignore misspelled only

MosheEichler:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/XSUP-54389


samuelFain (approved):
Nice one! Approved.

PR #5055, opened 2025-08-31
fix issue

jbabazadeh:
## Related Issues
related: https://jira-dc.paloaltonetworks.com/browse/XSUP-55404


israelpoli (approved):
💪 

I would add a `key:value` pair `"test:blabla": {"test": {"test"}}` under the root of DATA obj in marketplace_suffix_preparer_test.py,
In addition, I would update the assertion `data=...` in all UT functions in that file that use DATA, so that this key:value pair is also included and the case is properly covered by the unit tests.

PR #5079, opened 2025-09-21
Align content to id and name

Shir2611:
## Related Issues
fixes: [CIAC-12645](https://jira-dc.paloaltonetworks.com/browse/CIAC-12645)

### DIFF
--- a/demisto_sdk/commands/common/schemas/layoutrule.yml
+++ b/demisto_sdk/commands/common/schemas/layoutrule.yml
@@ -1,11 +1,19 @@
 type: map
 mapping:
+  id:
+    type: str
+    required: true
   rule_id:
+    type: str
+    required: false
+    # DEPRECATED: Use 'id' field instead. This field is kept for backward compatibility with existing content items.
+  name:
     type: str
     required: true
   rule_name:
     type: str
-    required: true
+    required: false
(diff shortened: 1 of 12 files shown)


.changelog/5079.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Add validators to enforce standardized id/name fields.
```
rshunim: maybe here we need to change it now to something like:
"Updated  a few items so ...."
explaining the structure of some items has changed

Shir2611: will do!

rshunim (changes requested):
Great Job!
1. Considering add 1-2 tests for ST110 validation which check your case (success and failure)
2. if "name" and "trigger_name" for example should be the same (and in other items), we can add a validator in the  Pydantic object which checks this. let's discuss about it

rshunim (approved):
Great Job! Well Done!
Don't forget to release this SDK PR with your Content Repo changes at the same time, otherwise the nightly will fail

jbabazadeh (approved):
Great work

Shir2611:
A force merge is required because this PR introduces changes to the structure of the strict object. The new structure is incompatible with what currently exists in the content repository, causing standard merge checks to fail. We are also applying a matching update in the content repo to keep both sides aligned.

PR #5164, opened 2025-12-16
Fixes issue with interactive formatting

Anonymoo53:
When using the format command, passing the assume option does not trigger the formatting to be conducted in interactive mode, despite the documentation:

"Assumes automatic yes/no as an answer to all prompts and runs non-interactively."

Underlying modules attempt to check for the interactive flag having been set (e.g. https://github.com/Anonymoo53/demisto-sdk/blob/1c9e31da20c2c87e546131158f8687b05b260c33/demisto_sdk/commands/format/update_playbook.py#L325); However, the assume-yes/no arguments do not trigger this to be happen.

This is a simple change to automatically set the interactive flag based upon whether the assume yes/no flag is set.

relates: https://jira-dc.paloaltonetworks.com/browse/CIAC-15445


Anonymoo53:
@BEAdi Do you have any feedback?

BEAdi (approved):

PR #5206, opened 2026-01-18
Internal and Prompt

hyaffe839:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CRTX-218187


dantavori (approved):

demisto_sdk/commands/validate/validators/GR_validators/GR110_is_agentix_action_using_existing_content_item_valid.py
```diff
@@ -89,19 +89,28 @@ def obtain_invalid_content_items_using_graph(
...
             action_type = action.underlying_content_item_type
 
-            if action_type not in {"command", "script", "playbook"}:
+            if action_type not in {
+                "command",
+                "script",
+                "playbook",
+                "internal",
+                "prompt",
+            }:
```
jbabazadeh, line 98: do we have prompt and internal as content item in the graph? 
what about the dependency on them is it need to be also calculate?

jbabazadeh (commented):
please see my comment
and please add content pr/mr that have action with prompt and internal to see its not break other validations

demisto_sdk/commands/validate/validators/GR_validators/GR110_is_agentix_action_using_existing_content_item_valid.py
hyaffe839: We only include `prompt` as a content item in the graph (identified as `SCRIPT` via the `is_llm` field). Since `internal` items are not content items, I've added dependency logic solely for `prompt`.

hyaffe839: PR: https://github.com/demisto/content/pull/42792
MR: https://gitlab.xdr.pan.local/xdr/cortex-content/content-private/-/merge_requests/80

jbabazadeh (approved):

PR #5318, opened 2026-04-14
Add support for connectors object

YuvHayun:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-17010


demisto_sdk/scripts/validate_validation_config_file.py
```diff
@@ -59,6 +59,7 @@ def validate_all_validations_run_on_git_mode(
 ) -> int:
     """
     Validate that all validations configured in the path_based section are also configured in the use_git section.
```
joeymizrahi:
```suggestion
    Verifies that all validations configured in the path_based section are also configured in the use_git section.
```

demisto_sdk/scripts/validate_validation_config_file.py
```diff
@@ -74,6 +75,23 @@ def validate_all_validations_run_on_git_mode(
...
+    connector_path_based = set(
+        config_file_content.get("connectors_path_based_validations", {}).get(
+            "select", []
+        )
+    )
+    connector_use_git = set(
+        config_file_content.get("connectors_use_git", {}).get("select", [])
+    )
+    if non_configured_connector_codes := connector_path_based - connector_use_git:
+        logger.error(
```
joeymizrahi, line 89: can you make the error message more clear?

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+)
+
+# ============================================================
+# Shared field sub-models
+# ============================================================
+
+
+class FieldModifiers(BaseModel):
+    required: bool = False
+    disabled: bool = False
```
joeymizrahi: disabled was removed from field modifiers

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+
+class FieldOptions(BaseModel):
+    mask: Optional[bool] = None
+    placeholder: Optional[str] = None
+    description: Optional[str] = None
+    help_text: Optional[str] = None
+    default_value: Optional[Any] = None
+    hint: Optional[str] = None
+    values: Optional[List[dict]] = None
+    layout: Optional[dict] = None
```
joeymizrahi: were not gonna use this for July, and maybe not ever

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+class ConnectorField(BaseModel):
+    """A single field definition used across connection, capabilities, and configurations."""
+
+    id: str
+    title: str
+    field_type: str  # "input", "select", "checkbox", "switch"
+    metadata: Optional[dict] = None
+    options: Optional[FieldOptions] = None
+    validations: Optional[List[dict]] = None
+    behavior: Optional[dict] = None
```
joeymizrahi, line 68: not better to have this as its own type also?

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+    maintainers: List[str] = []
+
+
+class ConnectorMetadata(BaseModel):
+    title: str
+    description: str
+    version: str  # semver e.g. "1.0.0"
+    category: str
+    tags: List[str] = []
+    domain: Optional[str] = None
```
joeymizrahi, line 96: not for July

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+    vendor: str
+    publisher: str
+    author_image: Optional[str] = None
+    ownership: ConnectorOwnership
+
+
+class ConnectorSettings(BaseModel):
+    allow_skip_verification: bool = False
```
joeymizrahi: default should be true

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+
+class Connector(ContentItem, content_type=ContentType.CONNECTOR):  # type: ignore[call-arg]
+    """A unified connector content item.
+
+    Uses a hybrid approach:
+    - Pydantic sub-models for structured, queryable data
+    - RelatedFile instances for file-level concerns
+    """
+
+    # === Fields from connector.yaml ===
```
joeymizrahi, line 353: shouldnt there also be a field for "id" for the connector id?

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+# Serializer data — parsed from serializer.yaml
+# (defined before HandlerData to avoid forward references)
+# ============================================================
+
+
+class FieldMapping(BaseModel):
+    """Raw serializer entry from serializer.yaml."""
+
+    id: str  # connector field ID (connector_param_name)
+    field_name: str  # integration parameter name (content_param_name)
```
joeymizrahi:
```suggestion
    field_name: str  # value to transform integration parameter name (content_param_name) to
```

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+    Maps connector field IDs to integration parameter names.
+    If a field appears in the serializer, the names differ.
+    If not, both names equal the field ID.
+    """
+
+    connector_param_name: str  # field ID in connector YAML (e.g. "domain")
+    content_param_name: str  # param name in integration YAML (e.g. "InstanceURL")
+    field_value_transform: Optional[str] = None  # optional value transform
+    is_serialized: bool = False  # True if mapping came from serializer.yaml
+    source_file: str = ""  # which connector file defines this field
```
joeymizrahi, line 219: why is the source file needed? only for debug?

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+
+
+# ============================================================
+# Handler data — parsed from components/handlers/<name>/handler.yaml
+# ============================================================
+
+
+class HandlerTriggering(BaseModel):
+    type: str  # "ZERO_SCALE" or "PUB_SUB"
```
joeymizrahi: always PUB_SUB

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+    labels: Optional[Dict[str, str]] = None
```
joeymizrahi, line 246: always 
  labels:
    xsoar-integration-id: "Salesforce"
    xsoar-pack-id: "Salesforce"

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+    args: Optional[dict] = None
+
+
+class HandlerAuthOption(BaseModel):
+    id: str  # references connection profile ID
+    scopes: List[str] = []
```
joeymizrahi: optional

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+    workloads: List[str] = []
+    methods: Optional[List[Any]] = None
```
joeymizrahi: its a list of connection profile ids, so a list of strings. not Any

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+    service: Optional[str] = None
+    endpoint: str
+    headers: Optional[Dict[str, str]] = None
+
+
+class HandlerData(BaseModel):
+    """Parsed structured data from a handler.yaml file."""
+
+    id: str
+    metadata: dict  # raw metadata dict
```
joeymizrahi: i think also create a dto for this

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+        return self.metadata.get("module")
+
+    @property
+    def team(self) -> Optional[str]:
+        return self.metadata.get("ownership", {}).get("team")
+
+    @property
+    def is_xsoar(self) -> bool:
+        """Identify if this handler is XSOAR-related."""
+        return self.module == "xsoar" or self.team == "xsoar"
```
joeymizrahi: if one is xsoar and not the other then its a bug.

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+        return None
+
+    @property
+    def xsoar_pack_id(self) -> Optional[str]:
+        if self.triggering.labels:
+            return self.triggering.labels.get("xsoar-pack-id")
+        return None
+
+    @property
+    def xsoar_content_id(self) -> Optional[str]:
```
joeymizrahi: this isnt a thing. remove

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+            return self.triggering.labels.get("xsoar-content-id")
+        return None
+
+
+# ============================================================
+# Capability-handler mapping
+# ============================================================
+
+
+class CapabilityHandlerMapping(BaseModel):
```
joeymizrahi, line 317: have in docstring an example of this dto. 
not sure i understand what its supposed to hold.

demisto_sdk/commands/content_graph/parsers/related_files.py
```diff
@@ -439,3 +446,192 @@ def config_docstring(self) -> dict:
...
         docstring = ast.get_docstring(parsed_ast)
         return json5.loads(docstring) if docstring else {}
+
+
+# ============================================================
+# Connector-related file classes
+# ============================================================
+
+
+class ConnectorYAMLRelatedFile(RelatedFile):
```
joeymizrahi, line 474: shouldnt we also have a ConnectorRelatedFile for "connector.yaml"?

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,401 @@
...
+                if handler_dir.is_dir():
+                    h = HandlerRelatedFile(
+                        self.path, handler_dir.name, git_sha=self.git_sha
+                    )
+                    if h.exist:
+                        result.append(h)
+        return result
+
+    @staticmethod
+    def match(_dict: dict, path: Path) -> bool:
```
joeymizrahi, line 605: elaborate what this is for

demisto_sdk/commands/content_graph/tests/test_data/connector_capabilities.yaml
joeymizrahi, line 1: would do more complex test file, with general_config. sub_capabilities, etc..

demisto_sdk/commands/content_graph/tests/test_data/connector_connection.yaml
joeymizrahi, line 1: would do more complex test file

demisto_sdk/commands/validate/initializer.py
```diff
@@ -616,11 +623,22 @@ def get_items_from_packs(
...
+        ``ConnectorAwareInitializer.gather_objects_to_run_on``.
+
+        Args:
+            path_filter: Optional callable ``(Path) -> bool``.  When provided,
+                only paths for which the callable returns True are kept before
+                the expensive parsing step.
 
         Returns:
-            Tuple[Set[BaseContent], Set[Path], Set[Path]]: The sets of all the successful casts, the sets of all failed casts, and the set of non content items.
+            A statuses dict ready to be passed to ``git_paths_to_basecontent_set``.
```
joeymizrahi, line 642: elaborate what is returned with an example

demisto_sdk/commands/validate/initializer.py
```diff
@@ -1113,3 +1156,268 @@ def obtain_metadata_path(self, path: Path) -> Path:
...
+    def _is_relevant_path(path: Path) -> bool:
+        """Return True if the path is an Integration or Connector item."""
+        path_str = str(path)
+        return (
+            f"/{INTEGRATIONS_DIR}/" in path_str
+            or path_str.startswith(f"{INTEGRATIONS_DIR}/")
+            or "connectors/" in path_str
+        )
+
+    def gather_objects_to_run_on(self) -> Tuple[Set[BaseContent], Set[Path]]:
```
joeymizrahi, line 1202: add docstring explaining whats collected and whats filtered out

demisto_sdk/commands/validate/initializer.py
```diff
@@ -1113,3 +1156,268 @@ def obtain_metadata_path(self, path: Path) -> Path:
...
+                    )
+
+        # 3. Cross-match and expand with missing counterparts
+        filtered = self._cross_match_and_expand(
+            filtered_integrations, filtered_connectors
+        )
+
+        return filtered, invalid_items
+
+    def _cross_match_and_expand(
```
joeymizrahi, line 1272: try and sepereate to functions. 
docstring explaining example

demisto_sdk/commands/validate/validators/CO_validators/CO112_is_matching_license.py
```diff
@@ -0,0 +1,91 @@
...
+        "{handler_licenses}. Missing modules: {missing_modules}"
+    )
+    related_field = "config.required_license"
+    is_auto_fixable = False
+
+    def obtain_invalid_content_items(
+        self,
+        content_items: Iterable[ContentTypes],
+    ) -> List[ValidationResult]:
+        """Check per-handler that the capability license union covers supportedModules.
```
joeymizrahi: will it cover also the supportedModules of the packs?

demisto_sdk/commands/content_graph/parsers/connector.py
```diff
@@ -0,0 +1,603 @@
...
+            configurations.yaml     # per-capability config (optional)
+            triggers.yaml           # triggers (optional)
+            summary.yaml            # summary (optional)
+            components/handlers/
+                <handler-name>/
+                    handler.yaml    # handler config (required per handler)
+                    serializer.yaml # field mappings (optional)
+    """
+
+    def __init__(
```
joeymizrahi, line 70: add docstrings

demisto_sdk/commands/content_graph/parsers/connector.py
```diff
@@ -0,0 +1,603 @@
...
+            pack_marketplaces,
+            pack_supported_modules=pack_supported_modules or [],
+            git_sha=git_sha,
+        )
+        # Override structure_errors to be a proper list
+        # (BaseContentParser sets it to a FieldInfo object)
+        self.structure_errors = []
+
+        # Parse connector.yaml fields
+        self.connector_metadata: dict = self.yml_data.get("metadata", {})
```
joeymizrahi, line 93: what about the connector.yml id?

demisto_sdk/commands/content_graph/parsers/connector.py
```diff
@@ -0,0 +1,603 @@
...
+        self.settings: Optional[dict] = self.yml_data.get("settings")
+
+        # Create related files
```
joeymizrahi, line 96: why not have one for the connector.yaml

demisto_sdk/commands/content_graph/parsers/connector.py
```diff
@@ -0,0 +1,603 @@
...
+            configurations=configurations,
+        )
+
+    def _parse_capabilities_with_configs(self) -> List[CapabilityData]:
+        """Parse capabilities.yaml and merge with configurations.yaml.
+
+        Each capability gets a unified ``configurations`` list containing:
+        1. general_configurations from capabilities.yaml (shared across all)
+        2. general_configurations from configurations.yaml (shared across all)
+        3. Per-capability configurations from configurations.yaml (matching by ID)
```
joeymizrahi, line 226: what about the configuration params in the connection profile for this capability?
out of scope for now?

joeymizrahi (changes requested):

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: Moved completley? or somewhere else?

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: wdym by this? FieldOptions? or a specific field (which)?

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: yap

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: scopes? id? HandlerAuthOption?

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: fixing and adding a validation.

demisto_sdk/commands/content_graph/objects/connector.py
joeymizrahi: layout

demisto_sdk/commands/content_graph/objects/connector.py
joeymizrahi: there is no more field modifier called disabled. it was replaced by read_only

demisto_sdk/commands/content_graph/objects/connector.py
joeymizrahi: scopes

demisto_sdk/commands/validate/validators/CO_validators/CO112_is_matching_license.py
YuvHayun: Integration object either uses it's own supportedModules if exist, or inherit from the Pack in the parsing time.

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: have what? ConnectorField?

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: what? ConnectorMetadata?

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: It's referenced higher in the inheritance hierarchy

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: huh? a placeholder for a very specific integration? doesn't really make sense

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: You mean a handler metadata object?

demisto_sdk/commands/content_graph/objects/connector.py
joeymizrahi: no. i mean that for us the labels are not optional. they are required.
and the label always must contain 2 keys. 
xsoar-integration-id , xsoar-pack-id

demisto_sdk/commands/content_graph/parsers/related_files.py
YuvHayun: No, the connector.yml is the main file

demisto_sdk/commands/content_graph/objects/connector.py
joeymizrahi: yes

demisto_sdk/commands/content_graph/tests/test_data/connector_capabilities.yaml
YuvHayun: It's a template to use when building the mock object for testing. Not sure it makes sense to test sub-capabilities here, as it's a specific 1/2 test cases

demisto_sdk/commands/content_graph/tests/test_data/connector_connection.yaml
YuvHayun: Same answer as above.

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: so replace it with "read_only": bool = False under FieldModifiers?

demisto_sdk/commands/content_graph/objects/connector.py
YuvHayun: I think we already have it covered under CO100

demisto_sdk/commands/content_graph/parsers/connector.py
YuvHayun: Being added higher in the hierarchy.

demisto_sdk/commands/content_graph/parsers/connector.py
YuvHayun: a connector related file?

demisto_sdk/commands/content_graph/parsers/connector.py
YuvHayun: we have that implemented elsewhere.

demisto_sdk/commands/content_graph/objects/connector.py
joeymizrahi: no, just remove as you did. you already have read_only

joeymizrahi (approved):

rshunim (approved):
I reviewed most of the files. You did a good job! Well done!

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -0,0 +1,599 @@
...
+    - Pydantic sub-models for structured, queryable data
+    - RelatedFile instances for file-level concerns
+    """
+
+    # === Fields from connector.yaml ===
+    connector_metadata: ConnectorMetadata = Field(alias="metadata")
+    settings: Optional[ConnectorSettings] = None
+
+    @root_validator(pre=True)
+    def _rebuild_nested_from_neo4j(cls, values: Dict[str, Any]) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
roeishunim: this is the convention here

roeishunim:
1. New "Connector" Content Item Type
2. Content Graph Integration
3. New CO-Family Validations (Connector Validators)
4. Validate Command Infrastructure

demisto_sdk/commands/content_graph/interface/neo4j/neo4j_graph.py
```diff
@@ -305,7 +305,14 @@ def _add_nodes_to_mapping(self, nodes: Iterable[graph.Node]) -> None:
...
             )
-            logger.debug("{}", f"{self._id_to_obj=}")  # noqa: PLE1205
+            # NOTE: previously this also logged `f"{self._id_to_obj=}"`, which
+            # forces a full repr() of every cached BaseContent object. Pydantic
+            # models repr() their nested fields recursively, and the graph now
+            # contains items with circular relationships (e.g.
+            # Connector -> Pack -> Integration -> Pack -> ...), so the repr
+            # walks itself forever and crashes with RecursionError. Log only
+            # the cache size - the full dump was diagnostic, not actionable.
+            logger.debug(f"_id_to_obj cache size: {len(self._id_to_obj)}")
```
tcarmeli1: is it needed? if so, lets minimize the commented sentences..

demisto_sdk/commands/content_graph/commands/update.py
```diff
@@ -190,26 +290,81 @@ def _update_content_graph_inner(
...
+    # walks per `update_content_graph` call.
+    changed_pack_ids: Set[str] = set()
+    changed_connector_ids: Set[str] = set()
+    git_diff_failed = False
+    if (
+        use_git
+        and (commit := content_graph_interface.commit)
+        and not is_external_repo
+        and not explicit_changes_provided
+    ):
```
tcarmeli1, line 346: in what cases is this relevant? why did you add this and `not explicit_changes_provided`?

demisto_sdk/commands/content_graph/objects/integration.py
```diff
@@ -137,6 +137,9 @@ class Integration(IntegrationScript, content_type=ContentType.INTEGRATION):  # t
         False, alias="isCloudProviderIntegration"
     )
 
+    # === Cross-link to matched handler (set by ConnectorAwareInitializer) ===
+    related_content: Optional[Any] = Field(None, exclude=True)
+
```
tcarmeli1, line 142: why any?

demisto_sdk/commands/content_graph/parsers/connector.py
```diff
@@ -0,0 +1,611 @@
...
+    def __init__(
+        self,
+        path: Path,
+        pack_marketplaces: List[MarketplaceVersions] = list(MarketplaceVersions),
+        pack_supported_modules: Optional[List[str]] = None,
+        git_sha: Optional[str] = None,
+    ) -> None:
+        # Ensure path points to the connector directory
+        if path.name == "connector.yaml":
+            path = path.parent
```
tcarmeli1, line 79: what is this for?

demisto_sdk/commands/content_graph/parsers/repository.py
```diff
@@ -33,14 +36,19 @@ def __init__(self, path: Path) -> None:
...
         progress_bar: Optional[tqdm] = None,
+        connectors_to_parse: Optional[Tuple[Path, ...]] = None,
     ):
-        if not packs_to_parse:
-            # if no packs to parse were provided, parse all packs
+        if packs_to_parse is None:
+            # No caller intent provided -> default to parsing every pack.
+            # Mirror the ``connectors_to_parse is None`` check below so an
+            # explicit empty tuple (caller-says-"none") is honored instead
+            # of being silently re-expanded to a full scan.
```
tcarmeli1, line 51: why did you changed that?

demisto_sdk/commands/content_graph/parsers/repository.py
```diff
@@ -57,6 +65,20 @@ def parse(
...
+            connectors_to_parse = tuple(self.iter_connectors())
+        if connectors_to_parse:
+            logger.debug("Parsing connectors...")
+            for connector_path in connectors_to_parse:
+                connector = RepositoryParser.parse_connector(connector_path)
+                if connector:
+                    self.connectors.append(connector)
+                    if progress_bar:
+                        progress_bar.update(1)
+
```
tcarmeli1, line 81: if/else?

demisto_sdk/commands/content_graph/commands/update.py
rshunim: explicit_changes_provided was added to avoid running git diff when we already know what changed.

It becomes True in two cases:

The user provided specific packs/connectors using --packs / --connectors or passed them directly through update_content_graph.
DEMISTO_SDK_DIFF_FILES was set and already provided the changed pack/connector IDs.

In these cases, we already have the exact list of items that need updating. Running git diff again is unnecessary and can cause problems.

This is especially important in CI. The commit we need to compare against is often not available locally because of shallow clones, forks, or detached HEAD states. In that situation, git_util.get_all_changed_pack_ids(commit) can fail. The previous behavior would catch the error and fall back to a full rebuild, ignoring the explicit list that was already provided.

The other conditions are still needed:

use_git — respects cases where git checks are disabled.
content_graph_interface.commit exists — without a base commit, there is nothing to compare.
not is_external_repo — external repositories do not use the bucket commit history.

explicit_changes_provided is just another reason to skip git diff, so it is added as an additional condition.

This affects cases like:

demisto-sdk update-content-graph --packs Foo,Bar
CI flows that provide changed files through DEMISTO_SDK_DIFF_FILES
Code that calls update_content_graph(packs_to_update=[...])

Without this check, these flows could fail to find the git history and unnecessarily run a full content graph creation instead of updating only the requested items.

demisto_sdk/commands/content_graph/objects/integration.py
rshunim: related_content is always assigned a Connector object at runtime, so Any is not being used because the type is unknown.

The main reason is to avoid import and typing complications.

Connector is defined in connector.py, while Integration is defined in integration.py. Adding a direct type reference can introduce circular import issues between the content graph object modules.
Using Optional["Connector"] would require handling Pydantic forward references (update_forward_refs), which is not currently used by the other models in this area of the codebase.

For consistency, the opposite relationship in connector.py (related_integration) is also typed as Optional[Any] for the same reasons.

So Any was kept to avoid introducing additional import and Pydantic setup requirements for a field that already behaves correctly at runtime.

demisto_sdk/commands/content_graph/parsers/connector.py
rshunim: This is a defensive fix to ensure the parser always receives the correct input shape.

ConnectorParser is designed to work with a directory path (the connector root), not a file path. Inside the class, self.path is assumed to be the root folder and is used directly in multiple places, for example:

self.path / "connector.yaml"
self.path / "components" / "handlers"
passing self.path into related file handlers

If someone accidentally passes connector.yaml instead of the containing folder, all of these path operations break in subtle ways. You’d end up with invalid paths like:

.../connector.yaml/connector.yaml
.../connector.yaml/components/handlers

This doesn’t always fail immediately with a clear error, and can instead lead to confusing downstream FileNotFoundErrors.

So the normalization at the entry point ensures:

The parser always works with a directory, as it expects
Misuse at call sites is corrected early
Failures happen close to the source, not deep inside the parsing flow

It’s essentially a safety check to prevent incorrect inputs from silently producing broken or misleading results.

demisto_sdk/commands/content_graph/parsers/repository.py
rshunim: The old condition (if not packs_to_parse) conflates two different cases:

packs_to_parse is None → caller didn’t specify anything (use default: scan all packs)
packs_to_parse = () → caller explicitly asked for no packs

Both evaluate to False, so the code incorrectly treated an explicit empty tuple the same as “no input” and expanded it to:

tuple(self.iter_packs())

The bug this caused

In the connectors-only flow, we now rely on callers passing:

packs_to_parse = ()
connectors_to_parse = (changed_connector,)

**The intention is: don’t parse any packs, only process connectors.**

But with the old logic, () triggered the default branch, which re-expanded to all packs (~1300). This silently negated the optimization and turned a targeted run back into a full scan.

demisto_sdk/commands/content_graph/parsers/repository.py
rshunim: The two ifs check different things, not a two-way branch:

- if connectors_to_parse is None: -> "no caller intent, default to full scan" (assigns from iter_connectors())
- if connectors_to_parse: -> "there is actually work to do" (skips the loop and the log when the list is empty)

An if/else on is None wouldn't collapse them, because the None branch still needs the truthy guard afterwards (iter_connectors() can legitimately return ()). Keeping it as two ifs also matches the packs branch [repository.py:47](vscode-webview://0t6d2vpv7t7hfabm5fdnnd2b8esiv5umirjemmpf2vssf3cg0dor/demisto_sdk/commands/content_graph/parsers/repository.py:47) so both parameters read the same way.

tcarmeli1 (approved):

BEAdi (approved):

PR #5320, opened 2026-04-15
 The `generate-unit-tests` command is deprecated

shmuel44:
## Related Issues
Related: CIAC-12263

## Description
The generate-unit-tests command is deprecated and will be removed in the next version.

This command depends on the `klara` library, which has not been updated since 2021. `klara` in turn depends on the typed-ast library, which is deprecated in Python 3.13.

So it is blocking us from updating the demisto-sdk to support Python 3.13


demisto_sdk/commands/generate_unit_tests/generate_unit_tests_setup.py
```diff
@@ -62,6 +62,10 @@ def generate_unit_tests(
     >NOTE: The generate-unit-test command only works if demisto-sdk is installed with pip install demisto-sdk [generate-unit-tests].
 
     """
+    logger.warning(
+        "<yellow>WARNING: The `generate-unit-tests` command is deprecated and will be removed in the next version.</yellow>"
```
BEAdi, line 66: Instead of writing "in the next version" write "in a subsequent version", to not specify a specific version that we will remove it in

BEAdi (approved):

PR #5327, opened 2026-04-29
new validation for non mandatory supported modules

Shir2611:
## Related Issues
fixes: [CIAC-15887](https://jira-dc.paloaltonetworks.com/browse/CIAC-15887)

## Description
new validation for non mandatory supported modules


demisto_sdk/commands/content_graph/interface/neo4j/queries/validations.py
```diff
@@ -576,6 +576,37 @@ def get_supported_modules_mismatch_content_items(
     return results
 
 
+def get_non_mandatory_supported_modules_mismatch_dependencies(
```
BEAdi: Instead of adding a new function, use the existing one and add an argument to mark whether we are using mandatory or non mandatory dependacy, like we talked in the kickoff

BEAdi (changes requested):
Looks good!!
What about commands? Like in GR109

demisto_sdk/commands/content_graph/interface/neo4j/queries/validations.py
Shir2611: true. fixed thanks

demisto_sdk/commands/content_graph/interface/neo4j/neo4j_graph.py
```diff
@@ -717,8 +716,9 @@ def find_content_items_with_non_mandatory_module_mismatch_dependencies(
         """
         with self.driver.session() as session:
             results = session.execute_read(
-                get_non_mandatory_supported_modules_mismatch_dependencies,
+                get_supported_modules_mismatch_dependencies,
                 content_item_ids,
+                False,
```
BEAdi: Instead of adding here another function, we can also send True/False for the mandatory dependency for `find_content_items_with_module_mismatch_dependencies`

BEAdi (commented):
Instead of copying the GR109 validation entirely, to keep the validation the same, just inherit from it and add the mandatory value as a class member, which will be overridden by the new validation instance

demisto_sdk/commands/content_graph/interface/neo4j/neo4j_graph.py
Shir2611: true fixed

demisto_sdk/commands/validate/validators/GR_validators/GR114_is_non_mandatory_supported_modules_compatibility_all_files.py
```diff
@@ -0,0 +1,20 @@
...
+
+class IsNonMandatorySupportedModulesCompatibilityAllFiles(
+    IsNonMandatorySupportedModulesCompatibility
+):
+    expected_execution_mode = [ExecutionMode.ALL_FILES]
+
+    def obtain_invalid_content_items(
+        self, content_items: Iterable
+    ) -> List[ValidationResult]:
+        return self.obtain_invalid_content_items_using_graph(content_items, True)
```
BEAdi:
```suggestion
        return self.obtain_invalid_content_items_using_graph(content_items, False)
```

BEAdi (approved):

yedidyacohenpalo (approved):

PR #5333, opened 2026-05-03
Add param to capability mapper

YuvHayun:
## Related Issues
fixes:

### DIFF
--- a/demisto_sdk/scripts/connector_param_mapper.py
+++ b/demisto_sdk/scripts/connector_param_mapper.py
@@ -162,7 +162,9 @@ def _apply_manual_mapping(
     command_params: dict,
     manual_command_to_capability: Dict[str, List[str]],
 ) -> set:
-    """Step 2.1.5 — Apply manual command-to-capability overrides.
+    """
+    manual_command_to_capability - mapping command name -> list of capability names.
+    Step 2.1.5 — Apply manual command-to-capability overrides.
 
     Manual mapping is the source of truth for any listed command. For each entry:
       1. Ensure each listed capability exists in the result dict (create with []).
@@ -254,7 +256,9 @@ def _multi_capability_mapping(
     command_params: dict,
     handled_commands: Optional[set] = None,
 ) -> None:
-    """Step 2.3 - For each command, map its params to the matching capability
+    """
+    command_params structure- {integration: '', commands: {command: [params]}}
+    Step 2.3 - For each command, map its params to the matching capability
     (or ``Automation``).  Skips test-module (handled in 2.1) and any command
     already routed by manual mapping (Step 2.1.5).  Warns if the target
     capability is missing from the result mapping.
--- a/demisto_sdk/scripts/connector_param_mapper.py
+++ b/demisto_sdk/scripts/connector_param_mapper.py
@@ -33,7 +33,7 @@ EXCLUDED_AUTOMATION_PATTERNS: List[str] = [
 # ---------------------------------------------------------------------------
 # Step 1: Decide capabilities
 # ---------------------------------------------------------------------------
-def _is_pure_event_collector(integration_yml: dict) -> bool:
+def _is_pure_event_collector(integration_yml: dict, command_names) -> bool:
     """Check whether the integration is a pure event collector with no other fetch capabilities.
 
     Returns True only if the integration has isfetchevents but NO other fetch indicators
@@ -53,6 +53,9 @@ def _is_pure_event_collector(integration_yml: dict) -> bool:
     for param in integration_yml.get("configuration", []) or []:
         if param.get("name") == "isFetchCredentials":
             return False
+    get_events_cmd_count = sum(1 for n in command_names if "get-events" in n)
+    if (len(command_names) - get_events_cmd_count == 0):
+        return False
     return True
 
 
@@ -79,11 +82,7 @@ def decide_capabilities(integration_yml: dict) -> Dict[str, List[str]]:
     # Rule 2 - Log Collection (with possible early exit)
     if script.get("isfetchevents") is True:
         result["Log Collection"] = []
-        get_events_cmd_count = sum(1 for n in command_names if "get-events" in n)
-        if (
-            "Event Collector" in integration_name
-            or (len(command_names) - get_events_cmd_count == 0)
-        ) and _is_pure_event_collector(integration_yml):
+        if "Event Collector" in integration_name and _is_pure_event_collector(integration_yml, command_names):
             # Pure event collector — short-circuit to keep the result minimal
             return {"general_configurations": [], "Log Collection": []}
 
@@ -127,15 +126,18 @@ def _handle_test_module(
     param_defaults: dict,
 ) -> None:
     """Step 2.1 - Add params from ``test-module`` without a default to
-    ``general_configurations``."""
+    ``general_configurations``.
+
+    Uses a local ``general_set`` for O(1) membership checks instead of
+    repeatedly scanning the underlying list.
+    """
     commands_section: dict = command_params.get("commands") or {}
     test_module_params: List[str] = commands_section.get("test-module", []) or []
+    general_set: set = set(result["general_configurations"])
     for param in test_module_params:
-        if (
-            param not in param_defaults
-            and param not in result["general_configurations"]
-        ):
+        if param not in param_defaults and param not in general_set:
             result["general_configurations"].append(param)
+            general_set.add(param)
 
 
 def _apply_manual_mapping(
@@ -153,6 +155,9 @@ def _apply_manual_mapping(
     Returns the set of command names that were handled here, so subsequent steps
     (2.2 / 2.3) can skip them and avoid double-routing.
 
+    Uses a ``placed_per_cap`` dict-of-sets for O(1) per-capability dedup
+    instead of scanning ``result[cap]`` linearly on every check.
+
     No-op when ``manual_command_to_capability`` is empty.
     """
     handled_commands: set = set()
@@ -160,6 +165,7 @@ def _apply_manual_mapping(
         return handled_commands
 
     commands_section: dict = command_params.get("commands") or {}
+    placed_per_cap: Dict[str, set] = {}
     for cmd_name, capability_list in manual_command_to_capability.items():
         # Ensure each capability exists.
         for cap in capability_list:
@@ -168,9 +174,11 @@ def _apply_manual_mapping(
         # Route this command's params to each listed capability.
         params = commands_section.get(cmd_name) or []
         for cap in capability_list:
+            cap_set = placed_per_cap.setdefault(cap, set(result[cap]))
             for param in params:
-                if param not in result[cap]:
+                if param not in cap_set:
                     result[cap].append(param)
+                    cap_set.add(param)
         handled_commands.add(cmd_name)
     return handled_commands
 
@@ -232,18 +240,26 @@ def _multi_capability_mapping(
     """Step 2.3 - For each command, map its params to the matching capability
     (or ``Automation``).  Skips test-module (handled in 2.1) and any command
     already routed by manual mapping (Step 2.1.5).  Warns if the target
-    capability is missing from the result mapping."""
+    capability is missing from the result mapping.
+
+    Uses a ``placed_per_cap`` dict-of-sets for O(1) per-capability dedup
+    instead of scanning ``result[target]`` linearly on every check.
+    """
     handled_commands = handled_commands or set()
     commands_section: dict = command_params.get("commands") or {}
+    placed_per_cap: Dict[str, set] = {}
     for cmd_name, params in commands_section.items():
         if cmd_name == "test-module" or cmd_name in handled_commands:
             continue
         target = _resolve_target_capability(cmd_name, result)
-        for param in params or []:
-            if target in result:
-                if param not in result[target]:
+        if target in result:
+            cap_set = placed_per_cap.setdefault(target, set(result[target]))
+            for param in params or []:
+                if param not in cap_set:
                     result[target].append(param)
-            else:
+                    cap_set.add(param)
+        else:
+            for param in params or []:
                 logger.warning(
                     f"{param} failed to add to {target} because it doesn't "
                     f"exist although it's a part of {cmd_name}."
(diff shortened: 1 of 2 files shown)


demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+
+COMMAND_TO_CAPABILITY: Dict[str, str] = {
+    "fetch-incidents": "Fetch Issues",
+    "fetch-events": "Log Collection",
+    "fetch-credentials": "Fetch Secrets",
+    "fetch-indicators": "Threat Intelligence & Enrichment",
+    "fetch-assets": "Fetch Assets and Vulnerabilities",
+}
+
+EXCLUDED_AUTOMATION_PATTERNS: List[str] = [
```
joeymizrahi, line 30: what about "fetch-indicators"?

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+        return False
+    if script.get("feed"):
+        return False
+    if script.get("isfetchassets"):
+        return False
+    # Check for isFetchCredentials param
+    for param in integration_yml.get("configuration", []) or []:
+        if param.get("name") == "isFetchCredentials":
+            return False
+    get_events_cmd_count = sum(1 for n in command_names if "get-events" in n)
```
joeymizrahi, line 71: we can assume that if the command name contains "get-events" its a pure event collector?

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+            return {"general_configurations": [], "Log Collection": []}
+
+    # Rule 3 - Fetch Issues
+    if script.get("isfetch") is True and script.get("isfetch:platform") is not False:
+        result["Fetch Issues"] = []
+
+    # Rule 4 - Threat Intelligence & Enrichment (with possible early exit)
+    if script.get("feed") is True:
+        result["Threat Intelligence & Enrichment"] = []
+        get_indicators_cmd_count = sum(
```
joeymizrahi, line 113: can assume that if the string "get-indicators" is in the command name then the integration is a feed?
Look at `Packs/Anomali_ThreatStream/Integrations/AnomaliThreatStreamv3/AnomaliThreatStreamv3.yml`
i dont see `feed:true`

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+            1 for n in command_names if "get-indicators" in n
+        )
+        if "feed" in integration_name and (
```
joeymizrahi, line 116: see `Packs/VirusTotal/Integrations/FeedLivehunt/FeedLivehunt.yml` which has command `fetch-indicators` is this also a feed command?

demisto_sdk/scripts/connector_param_mapper.py
joeymizrahi, line 30: do we want command `long-running-execution` to be outside of "automation" also?

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+            len(command_names) - get_indicators_cmd_count == 0
+        ):
+            return {
+                "general_configurations": [],
+                "Threat Intelligence & Enrichment": [],
+            }
+
+    # Rule 5 - Fetch Assets and Vulnerabilities
+    if script.get("isfetchassets") is True:
+        result["Fetch Assets and Vulnerabilities"] = []
```
joeymizrahi: i think these strings should be constants i.e. "Fetch Assets and Vulnerabilities"

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+
+    # Rule 6 - Automation
+    for name in command_names:
```
joeymizrahi:
```suggestion
    for command_name in command_names:
```

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+
+# ---------------------------------------------------------------------------
+# Step 2: Map params to capabilities
+# ---------------------------------------------------------------------------
+def _handle_test_module(
+    result: Dict[str, List[str]],
+    command_params: dict,
+    param_defaults: dict,
+) -> None:
+    """Step 2.1 - Add params from ``test-module`` without a default to
```
joeymizrahi: @JudahSchwartz we finalized on this decision?

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+    ``general_configurations``.
+
+    Uses a local ``general_set`` for O(1) membership checks instead of
```
joeymizrahi, line 148: this "general_set" object reflects the general configurations in the connection or configuration step of the connector?

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+    for param in test_module_params:
+        if param not in param_defaults and param not in general_set:
+            result["general_configurations"].append(param)
+            general_set.add(param)
+
+
+def _apply_manual_mapping(
+    result: Dict[str, List[str]],
+    command_params: dict,
+    manual_command_to_capability: Dict[str, List[str]],
```
joeymizrahi, line 163: can you add example in docstring of how manual_command_to_capability looks?

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+
+    # Step 2.1
+    _handle_test_module(result, command_params, param_defaults)
+
+    # Step 2.1.5 - manual override (source of truth for listed commands)
+    handled_commands = _apply_manual_mapping(
+        result, command_params, manual_command_to_capability
+    )
+
+    if len(result) == 2:
```
joeymizrahi: i would wrap `if len(result) == 2:` into some function like _is_single_capability

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+    already routed by manual mapping (Step 2.1.5).  Warns if the target
+    capability is missing from the result mapping.
+
+    Uses a ``placed_per_cap`` dict-of-sets for O(1) per-capability dedup
+    instead of scanning ``result[target]`` linearly on every check.
+    """
+    handled_commands = handled_commands or set()
+    commands_section: dict = command_params.get("commands") or {}
+    placed_per_cap: Dict[str, set] = {}
+    for cmd_name, params in commands_section.items():
```
joeymizrahi, line 272: can you share how command_params json file looks?
i dont understand how cmd_names and params are taken from it

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+         ``"Threat Intelligence & Enrichment"`` exists in the capabilities →
+         ``"Threat Intelligence & Enrichment"``.
+    3. Fallback: ``"Automation"``.
+    """
+    if cmd_name in COMMAND_TO_CAPABILITY:
+        return COMMAND_TO_CAPABILITY[cmd_name]
+    if "get-events" in cmd_name and "Log Collection" in result:
+        return "Log Collection"
+    if "get-indicators" in cmd_name and "Threat Intelligence & Enrichment" in result:
+        return "Threat Intelligence & Enrichment"
```
joeymizrahi: capability commands should be an ENUM. i.e. enum for "Threat Intelligence & Enrichment"

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -0,0 +1,497 @@
...
+        "-o",
+        "--output",
+        help="Output JSON file path.",
+    ),
+) -> None:
+    """Generate the connector parameter mapping from the integration YML and
+    the supplied command/defaults JSON inputs (with optional manual overrides)."""
+    logging_setup(calling_function=__name__)
+
+    command_params: Dict[str, Any] = json.loads(command_params_json)
```
joeymizrahi, line 488: these json files are for a single integration?

demisto_sdk/scripts/tests/connector_param_mapper_test.py
joeymizrahi, line 1: i think you should have here test json files.

joeymizrahi (changes requested):

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: Good catch.

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: The assumption here is that if all commands in yml contains get-events, then it's a pure event collector

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: But here we're looking at the sub-group where feed: true and they have get-indicators commands.

demisto_sdk/scripts/connector_param_mapper.py
joeymizrahi: yeah i get that, but is it something we can rely on? 
did you check with idex if there are commands that contain "get-events" string and its not an event-collector?

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: It does, but fetch commands usually doesn't appear in the yml as a command with the name fetch-<type>, they appear as the bool flag under script (feed, isfetch, isfetchevents, etc)

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: No, as discussed with @JudahSchwartz, we will not have a stand-alone capability for that.
However, we have the ability to add manual input where we can decide to map long-running-execution to some other capability if this is the purpose of long-running-execution.
It'll be a column in the CSV, I'll be responsible to map all integration with long-running-exec myself.

demisto_sdk/scripts/connector_param_mapper.py
joeymizrahi: ack, but we can assume that if command contains "get-indicators" then its classic feed?

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: Good idea, done.

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: You can see that in the generate_param_mapping func - "JSON string mapping command name -> list of capability names. "

demisto_sdk/scripts/tests/connector_param_mapper_test.py
YuvHayun: As per @JudahSchwartz we expect the inputs as json strings that needs to be loaded.

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: Yes, it's JSON strings, not files.

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: can be found in generate_param_mapping

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: The configuration step.

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: you just proved we can't

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: If we have "Event Collector" in the integration name, and the integration fetches events, and no other fetch type is configured, and only commands apears in the yml are get-events commands, then yeah. we can assume it's an event-collector.

demisto_sdk/scripts/tests/connector_param_mapper_test.py
joeymizrahi: doesnt mean we cant have test files with this data

demisto_sdk/scripts/connector_param_mapper.py
```diff
@@ -5,20 +5,26 @@
...
 import typer
-import yaml
 
 from demisto_sdk.commands.common.handlers import DEFAULT_JSON_HANDLER as json
+from demisto_sdk.commands.common.handlers import DEFAULT_YAML_HANDLER as yaml
 from demisto_sdk.commands.common.logger import logger, logging_setup
 
 main = typer.Typer()
 
+FETCH_ASSETS_CAPABILITIES = "Fetch Assets and Vulnerabilities"
```
joeymizrahi, line 15: wouldnt enum be nicer?

demisto_sdk/scripts/connector_param_mapper.py
joeymizrahi: not sure this is what we decided. 
why do we need to move params that are used in test-module without a default to general section?

demisto_sdk/scripts/connector_param_mapper.py
joeymizrahi: i think its worth having here too for clarity

demisto_sdk/scripts/connector_param_mapper.py
joeymizrahi: think its worth adding here for clarity

joeymizrahi (changes requested):

demisto_sdk/scripts/connector_param_mapper.py
YuvHayun: Because we need the params for all capabilities to run the tests.

joeymizrahi (approved):

PR #5341, opened 2026-05-07
AS109 - improve validation precision

MLainer1:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CRTX-248365

## Description
AS109 was producing false positive errors because it failed to detect context key references in several common patterns:

- Single-segment keys (e.g., SuspiciousCommandLines) used as simple: values in scriptarguments or conditions

- Shared-prefix keys (e.g., ${File.SHA256} when another task uses root: "File")

- DT filter expressions (e.g., ${FileEnrichment(val.TIMScore == 0).Value}) iscontext: true fields where the flag is in a parent dict

- Set command key arguments (e.g., scriptarguments.key.simple: SuspiciousCommandLines)

Rewrote the detection to use root-level matching: extract the first identifier from each context reference and check if it appears in any other task. This reduced false positives from 34 to 2 genuine errors across the full AutonomousPlaybooksPack.

Also simplified the implementation from a recursive dict walker to a string-regex approach (121 lines, 3 regexes).


MLainer1:
Ran `validate -a -sv AS109` on content (master) + private (Arik's mr - minorfixes) successfully:

image

yedidyacohenpalo (approved):

MLainer1:
This PR needs forcemerge due to an error in the Demisto sdk build:
image

PR #5346, opened 2026-05-10
Strip internal:true from script YAML and pack metadata during upload

rshunim:
- BaseScript.prepare_for_upload now removes the `internal` key from the prepared script data so the uploaded script is visible to users.
- Pack._dump_pack_metadata replaces the verbatim shutil.copy of pack_metadata.json with a JSON-aware copy that strips `internal`. Falls back to a plain copy if the source isn't valid JSON.
- Pack.dump_metadata adds `internal` to the excluded fields so the generated metadata.json never contains the flag either.
- Adds unit tests covering all three code paths.
- Adds .changelog/5328.yml.

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-16736


israelpoli (approved):
💪 Great

rshunim:
Force Merge is needed due to a known issue with the GR109 validation. Unrelated to my PR

jbabazadeh (approved):

PR #5389, opened 2026-05-28
fix: register all commands before attempting to dump the API

GraemeMeyerCAA:
## Description
The `export-api` command is supposed to [dump the API for every registered command](https://github.com/demisto/demisto-sdk/blob/3e6da00a37840d188bfb9ad418e64de50f1ddbee/demisto_sdk/commands/dump_api/dump_api_setup.py#L22), but only the `export-api` command is registered before the dump runs, so the output describes only export-api command itself.

My proposed fix is to include `export-api` in the list of scenarios which require all commands to be registered - treating `export-api ` the same as `--help` and test mode. This causes every command to be registered for an `export-api` command, and fixes the bug.

My first contribution to the project - open to feedback. Wanted to start with something small. Will sign the CLA as soon as the bot sends it to me, and will try the `poetry run sdk-changelog --init` command as well as soon as the PR is up, per the [Contributing to Demisto SDK](https://xsoar.pan.dev/docs/contributing/sdk) guide.

relates: https://jira-dc.paloaltonetworks.com/browse/CIAC-16935

### DIFF
--- a/.changelog/5389.yml
+++ b/.changelog/5389.yml
@@ -1,0 +1,4 @@
+changes:
+- description: 'fix: register all commands before attempting to dump the API'
+  type: fix
+pr_number: 5389
--- a/demisto_sdk/__main__.py
+++ b/demisto_sdk/__main__.py
@@ -153,7 +153,8 @@ def register_commands(_args: list[str] = []):  # noqa: C901
 
     is_test = not _args
     is_help = "-h" in _args or "--help" in _args
-    register_all = any([is_test, is_help])
+    is_export_api = command_name == "export-api"
+    register_all = any([is_test, is_help, is_export_api])
 
     # Pre-commit runs a few commands as hooks.
     is_pre_commit = "pre-commit" == command_name


GraemeMeyerCAA:
Signed the CLA, and added the .changelog/5389.yml. The contributing guide doesn't actually say the file should be included in the PR, but I'm assuming it should be because otherwise I don't know why I would be supposed to generate it and keep it locally...

GraemeMeyerCAA:
Hey folks the docs state that PRs will receive an initial response within [5 business days](https://xsoar.pan.dev/docs/contributing/sla) - is that still accurate? This isn't an urgent issue, but I'd like to understand how contributions and issues with the SDK can be raised effectively. I'm now receiving messages from the bot saying the issues are going stale from inactivity @kamalq97  @jbabazadeh ?

apollocreed:
Hey I'm interrested as well please fix it

kamalq97:
Hi @GraemeMeyerCAA and @apollocreed,

Thanks for the ping and for your patience. You are right that this slipped past our usual response SLA, so thank you for keeping it on our radar.

I appreciate you submitting this fix and handling the CLA and changelog so smoothly for your first contribution. I am currently working internally to get this PR re-assigned to the correct maintainer on the SDK team so we can get it moving again!

ShacharKidor (changes requested):
Hi @GraemeMeyerCAA,
first of all - apologies for the long delay in getting to this review, thanks for your patience!

Great work and thanks you for the fix - the diagnosis is spot-on and the change is nicely minimal. 👍

One thing I'd like to see before we merge: a small regression test so this can't silently break again if someone touches register_commands() in the future.

**What to test**:
The bug is essentially "export-api produces a JSON with only one key". A good test asserts the opposite: after invoking register_commands(["export-api"]), the Typer app is populated with many commands (not just export-api itself).

**Where to add it**:
Create a new test file next to the command implementation, so the test lives with the code it covers:

demisto_sdk/commands/dump_api/tests/dump_api_test.py

**What the test should look like**
Something along these lines:
```

# demisto_sdk/commands/dump_api/tests/dump_api_test.py
import json
from pathlib import Path

from typer.main import get_command


def test_export_api_registers_all_commands(tmp_path: Path):
    """
    Given: The user invokes the `export-api` command.
    When:  `register_commands` is called with ["export-api"].
    Then:  All SDK commands are registered on `app` (not just `export-api`),
           so that the resulting dump contains the full API surface.
    """
    # Import inside the test so register_commands runs with a fresh app state
    from demisto_sdk.__main__ import app, register_commands

    register_commands(["export-api"])

    typer_app = get_command(app)
    command_names = set(typer_app.commands.keys())  # type: ignore[attr-defined]

    # Sanity: export-api itself is there
    assert "export-api" in command_names
    # Regression guard: several unrelated commands must also be registered
    assert {"upload", "download", "validate", "format"}.issubset(command_names)
    # And the total should clearly be more than just export-api
    assert len(command_names) > 5

```
Happy to help if any of this is unclear or you have any question. 
Just ping me here or DM via DFIR.

PR #5405, opened 2026-06-09
fix(pre-commit): preserve ApiModule non-code files and pull sibling R…

rshunim:
Summary

Fixes demisto-sdk pre-commit silently skipping hooks targeting non-code files (e.g. markdownlint-cli2 on README.md) when the file lives inside an ApiModules pack folder. Affects both -i and --use-git flows.

Before: markdownlint-cli2 ....(no files to check) Skipped
After: markdownlint-cli2 ....Passed (or Failed if the file has violations)

Root cause
In group_by_language(), files under Packs/ApiModules/Scripts/<X>/ are stored in integrations_scripts_mapping. The API-module-expansion block only re-emits .py/.yml/.ps1 via add_related_files() into downstream integrations, and the second loop skips the ApiModules pack entirely — so any README/description/image/json sitting in an ApiModule folder is dropped before reaching pre-commit.

Changes
group_by_language() — rescue non-.py/.yml/.ps1 files inside an ApiModule folder onto infra_files, so they survive and land in language_to_files. Source-agnostic (works for -i, --use-git, --staged-only, --all-files).

add_related_files() — when the input is a .yml, also include the sibling README.md, so README-targeting hooks fire whenever an integration manifest is touched.

Tests
test_api_module_readme_is_preserved — repro of the bug report.
test_yml_change_pulls_readme — covers the symmetric fix.
Updated test_preprocess_files_with_input_yml_files to expect the README in the result.

## Related Issues
fixes:

### DIFF
--- a/.changelog/5405.yml
+++ b/.changelog/5405.yml
@@ -1,4 +1,4 @@
 changes:
-- description: Fixed an issue where `pre-commit` silently skipped hooks targeting non-code files (e.g. `markdownlint-cli2` on `README.md`) sitting inside an `ApiModules` pack folder.
+- description: Fixed an issue where `pre-commit` silently skipped hooks targeting non-code files (e.g. `markdownlint-cli2` on `README.md`).
   type: fix
 pr_number: '5405'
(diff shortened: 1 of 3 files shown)


jbabazadeh (approved):
Great work!

PR #5406, opened 2026-06-10
Added pack-level ignored validations

ShacharKidor:
## Related Issues
fixes: [CIAC-17006](https://jira-dc.paloaltonetworks.com/browse/CIAC-17006).

## Description

1. Expands the mechanism of the .pack-ignore file to support pack-level ignores (ignoring certain validations for all content items within a pack).
Tests were executed on this content PR - [pull/44594/](https://github.com/demisto/content/pull/44594/).

2. Enforce a force merge for a PR that added a pack-level ignored validation - Implemented by the new PA135 validation.
Tested on content-private in this MR - [merge_requests/522](https://gitlab.xdr.pan.local/xdr/cortex-content/content-private/-/merge_requests/522).


ShacharKidor:
Hi @jbabazadeh, @yedidyacohenpalo, 
I implemented the first phase of the task - Added support for pack-level ignored validation. 
Local tests have passed successfully.
Build test has passed successfully as you can see in this content PR - [pull/44594](https://github.com/demisto/content/pull/44594).

Please start the code review for the first phase.

rshunim (changes requested):
Good Job! Well done!

🔴 Must fix before merge

1. Add a test for content_item.in_pack is None
The new pack is not None guard in is_error_ignored is the only branch in this PR with zero coverage, and it is reachable in production - several content items legitimately have in_pack == None (see the existing if self.in_pack else "" guards in content_item.py). A future refactor could trivially drop the guard and crash every validator that runs on a pack-less item.

Suggested test (drop into TestIsErrorIgnoredPackLevel):

def test_item_with_no_pack_is_not_ignored(self):
    """
    Given a content item whose `in_pack` is None,
    When is_error_ignored is called,
    Then the pack-level check is skipped safely (no AttributeError)
    and the result is False.
    """
    item = _FakeContentItem(pack=None)
    assert is_error_ignored("BA101", ALLOWED, item) is False

2. Drop the # pragma: no cover on pack.py:193
test_exception_is_logged_and_empty_list_returned explicitly exercises that except branch, so the pragma is now inaccurate and will mislead anyone reading coverage reports. One-line deletion:

-        except Exception as e:  # pragma: no cover - defensive
             logger.debug(
                 f"Failed to extract pack-level ignored errors for {self.object_id}: {e}"
             )

🟡 Nice to have - won't block
Deduplicate the [pack] parsing logic
The "iterate section, find ignore key, split-and-strip" walk is hand-rolled in two places:

Pack.pack_level_ignored_errors in pack.py
extract_error_codes_from_file in tools.py
They are equivalent today, but they will drift the next time the [pack] grammar changes (e.g. supporting comments, line continuations, globs). Worth a small helper like _parse_ignore_list(value: str) -> list[str] shared by both call sites.

rshunim (approved):
Well Done!

ShacharKidor:
Hi @rshunim, @yedidyacohenpalo, 
Would appreciate your review of the addition of the new PA135 validation. 

Tested it on the following content private MR (on a private pack, and also for the split pack case) - [merge_requests/522](https://gitlab.xdr.pan.local/xdr/cortex-content/content-private/-/merge_requests/522).

demisto_sdk/commands/content_graph/objects/pack.py
```diff
@@ -167,6 +171,101 @@ def ignored_errors(self) -> List[str]:
...
+        """
+        try:
+            section = self.ignored_errors_dict.get("pack")
+            if section is None:
+                return []
+            for key, value in section.items():
+                if key == "ignore":
+                    return parse_ignore_list(value)
+            return []
+        except Exception as e:
```
rshunim: 1. The except is too broad and swallows everything
except Exception catches any error, including bugs that are not "the .pack-ignore was malformed":

AttributeError (e.g. someone accidentally sets ignored_errors_dict = None on the Pack)
TypeError (e.g. a future refactor changes SectionProxy to something that doesn't support .items())
MemoryError, KeyboardInterrupt, custom exceptions from unrelated subsystems, etc.
All of them are silently converted to "return empty list" and a debug log. The user never sees an error, and the validation flow happily continues without applying the pack-wide ignores the user asked for.

Concrete failure scenario: a user writes [pack] ignore=BA101 in .pack-ignore and expects BA101 to be silenced pack-wide. A future bug in the property causes an AttributeError. Result: the user's ignore is silently dropped, BA101 fires on every file, and no one knows why - not even the developer, unless they happen to grep debug logs.

Broad except Exception is a well-known anti-pattern for exactly this reason: it hides bugs instead of surfacing them.

rshunim: 2. The try/except isn't actually needed
Look at the operations inside the try:

self.ignored_errors_dict.get("pack") - .get() on a dict cannot raise.
if section is None: return [] - a comparison cannot raise.
section.items() - SectionProxy (the documented type) always supports .items().
str(value).split(",") - always safe on any value.
code.strip() - always safe on any string.
Under the documented input contract (which the docstring on [lines 177-181](vscode-webview://0t6d2vpv7t7hfabm5fdnnd2b8esiv5umirjemmpf2vssf3cg0dor/demisto_sdk/commands/content_graph/objects/pack.py:177) explicitly spells out), there is no code path that can raise an exception. The try/except is guarding against a bug that cannot happen unless someone violates the invariant - and if they do, we want to know, not silently pretend it worked.

Contrast with [get_ignored_errors](vscode-webview://0t6d2vpv7t7hfabm5fdnnd2b8esiv5umirjemmpf2vssf3cg0dor/demisto_sdk/commands/content_graph/objects/pack.py:205) right below it, which uses a bare except because it does inherently unsafe indexing (list(...)[0][1].split(",")) that can IndexError. That defensive block is justified. This one is not.

rshunim: Option A - drop the try/except entirely
Option B - keep the guard but narrow it

rshunim: except (AttributeError, TypeError) as e

rshunim (approved):
Well Done! I left one comment

demisto_sdk/commands/validate/sdk_validation_config.toml
```diff
@@ -220,6 +220,7 @@ select = [
   "PA131",
   "PA132",
   "PA134",
+  "PA135",
```
yedidyacohenpalo:
```suggestion
```
this validation should only run in use git

BEAdi (approved):

PR #5412, opened 2026-06-14
Add support dockerimage:demistoextended 

YairGlick:
## Related Issues
related: https://jira-dc.paloaltonetworks.com/browse/CIAC-16725


demisto_sdk/commands/common/tests/docker_helper_test.py
```diff
@@ -199,3 +199,371 @@ class MockedDockerClient:
...
+         - a demistoextended/ base image
+
+        When:
+         - the test image name is constructed in get_or_create_test_image
+
+        Then:
+         - the prefix should be devtestdemistoextended/ (not devtestdemistoextended/)
+        """
+        base_image = "demistoextended/accessdata:1.1.0.10177564"
+        # Simulate the logic from get_or_create_test_image
```
BEAdi: If you are simulating the logic and not using the function, so you are not checking that if it will change, the code might break 🙂 You can extract this logic from `get_or_create_test_image` to a helper function and test it.
Same in `test_demisto_image_still_uses_devtestdemisto_prefix, test_naive_replace_would_break_demistoextended`

demisto_sdk/commands/common/tests/docker_helper_test.py
```diff
@@ -199,3 +199,371 @@ class MockedDockerClient:
...
+         - the prefix should be devtestdemisto/ (existing behavior)
+        """
+        base_image = "demisto/python3:3.10.11.54799"
+        if base_image.startswith("demistoextended/"):
+            result = base_image.replace("demistoextended", "devtestdemistoextended")
+        else:
+            result = base_image.replace("demisto", "devtestdemisto")
+        assert result == "devtestdemisto/python3:3.10.11.54799"
+
+    def test_naive_replace_would_break_demistoextended(self):
```
BEAdi: [AI comment]
This test documents the old bug against inline code rather than the fix. I think it can be removed - the two behavior tests above already cover the correct outcome once they call the real function.

demisto_sdk/commands/common/tests/docker_helper_test.py
```diff
@@ -199,3 +199,371 @@ class MockedDockerClient:
...
+            "type": "python",
+            "dockerimage": "demistoextended/accessdata:1.1.0.10177564",
+        }
+        original_image = script_obj["dockerimage"]
+        ScriptYMLFormat.update_docker_image_in_script(
+            script_obj, "/fake/path/script.yml"
+        )
+        assert script_obj["dockerimage"] == original_image
+
+    def test_does_not_skip_demisto_image(self):
```
BEAdi: [AI comment]
This test has no assertion and relies on real network (try to query DockerHub and may fail). What is it actually verifying? I'd mock the DockerHub call and assert the image was updated (or at least that the early-return branch was not taken), otherwise it can pass or flake for the wrong reasons.

demisto_sdk/commands/common/docker/tests/docker_image_test.py
```diff
@@ -69,3 +72,203 @@ def test_docker_image_parse_invalid_docker_images(docker_image: str):
...
+         - checking is_trusted_repository
+
+        Then:
+         - returns False
+        """
+        image = DockerImage("someother/image:1.0")
+        assert image.is_trusted_repository is False
+
+
+class TestGetExtendedClient:
```
BEAdi, line 162: [AI comment]
Nice coverage of the None path. Can we also assert the registry_api_url override is built correctly (e.g. example-registry.io/test-project → https://example-registry.io/v2/test-project)? That is the derived value this method exists to produce.

demisto_sdk/commands/common/docker/docker_image.py
```diff
@@ -109,6 +149,14 @@ def is_tag_latest(self) -> bool:
     def is_demisto_repository(self) -> bool:
         return self.repository == "demisto"
 
+    @property
+    def is_demistoextended_repository(self) -> bool:
+        return self.repository == "demistoextended"
+
+    @property
+    def is_trusted_repository(self) -> bool:
+        return self.repository in {"demisto", "demistoextended"}
```
BEAdi: Save both `"demisto", "demistoextended"` to consts

demisto_sdk/commands/common/docker/docker_image.py
```diff
@@ -165,19 +211,16 @@ def is_image_exist(self) -> bool:
...
     @property
     def latest_docker_image(self) -> "DockerImage":
         """
         Returns the docker image with the latest tag
         """
-        return DockerImage(
-            self._get_dockerhub_client().get_latest_docker_image(self.name)
-        )
+        latest_tag = self._get_client().get_latest_docker_image_tag(self.name)
+        return DockerImage(f"{self.name}:{latest_tag}")
```
BEAdi: Why not keep using `get_latest_docker_image` which appends the latest tag itself?

demisto_sdk/commands/common/docker/dockerhub_client.py
```diff
@@ -406,7 +408,11 @@ def get_image_tags(self, docker_image: str) -> List[str]:
                 exception=error,
             )
 
-        return response.get("tags") or []
+        tags = response.get("tags") or []
+        logger.info(
+            f"get_image_tags | {docker_image} | registry={self.registry_api_url} | found {len(tags)} tags"
+        )
```
BEAdi: Needed? Or was only for debugging?

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -502,6 +503,37 @@ def create_image(
 
     @staticmethod
     def get_image_registry(image: str) -> str:
```
BEAdi, line 563: I think it will be more readable to extract the normalize CR prefix part and the apply extended registry part into small helpers functions. It is up to you because most of it will be removed soon

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -502,6 +503,37 @@ def create_image(
...
+        if image.startswith(CR_REGISTRY_PREFIX):
+            stripped = image[len(CR_REGISTRY_PREFIX) :]
+            logger.debug(
+                f"get_image_registry | stripped CR prefix: {image} -> {stripped}"
+            )
+            image = stripped
+        if any(
+            segment in image
+            for segment in ("demistoextended/", "devtestdemistoextended/")
+        ):
```
BEAdi: `"demistoextended/"` is in `"devtestdemistoextended/"` so you can only check it

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -502,6 +503,37 @@ def create_image(
...
+            # Already carries the extended registry prefix - return as-is to avoid
+            # double-prefixing it with the Docker Hub proxy registry.
+            if extended_registry and extended_registry in image:
+                logger.debug(f"get_image_registry | returned: {image}")
+                return image
+            if extended_registry and image.startswith(
+                ("demistoextended/", "devtestdemistoextended/")
```
BEAdi: and `devdemistoextended`?

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -502,6 +503,37 @@ def create_image(
...
+        extended_registry = os.getenv("DEMISTO_SDK_EXTENDED_REGISTRY", "")
+        # Normalize images that hardcode the container-registry prefix
+        # (e.g. "gcr.io/xsoar-registry/demistoextended/accessdata-p:tag") back to
+        # their canonical "demistoextended/..." form, so routing is uniform whether
+        # or not DEMISTO_SDK_EXTENDED_REGISTRY is set. Without this, an unset env var
+        # would leave the hardcoded prefix in place and the runner's Docker mirror
+        # would prepend its own registry, producing a broken double-registry path.
+        if image.startswith(CR_REGISTRY_PREFIX):
+            stripped = image[len(CR_REGISTRY_PREFIX) :]
+            logger.debug(
```
BEAdi: Only keep the debug logs that will be needed for the future

demisto_sdk/commands/format/update_script.py
```diff
@@ -71,6 +71,11 @@ def update_docker_image_in_script(
                 "<yellow>Skipping docker image update as default docker image is being used.</yellow>"
             )
             return
+        if dockerimage.startswith("demistoextended/"):
+            logger.info(
+                "<yellow>Skipping docker image update for demistoextended image.</yellow>"
```
BEAdi: We are keeping this for the next part?

BEAdi (changes requested):
Great work! Please aee my comments

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -51,6 +51,7 @@
 TEST_REQUIREMENTS_DIR = Path(__file__).parent.parent / "pre_commit" / "resources"
 DOCKER_CONTAINER_TIMEOUT = int(os.getenv("DOCKER_CONTAINER_TIMEOUT") or 300)
 CR_REGISTRY_PREFIX = "gcr.io/xsoar-registry/"
+EXTENDED_REPOSITORY_SEGMENT = "demistoextended/"
```
BEAdi: Reuse the `DEMISTO_EXTENDED_REPOSITORY` const when creating this const

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -501,47 +502,56 @@ def create_image(
...
             if extended_registry and extended_registry in image:
-                logger.debug(f"get_image_registry | returned: {image}")
                 return image
-            if extended_registry and image.startswith(
-                ("demistoextended/", "devtestdemistoextended/")
-            ):
-                logger.debug(
-                    f"get_image_registry | returned: {extended_registry}/{image}"
-                )
+            if extended_registry and image.startswith(EXTENDED_REPOSITORY_SEGMENT):
```
BEAdi: This is not the same as `image.startswith(
                ("demistoextended/", "devtestdemistoextended/")
            )`

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -501,47 +502,56 @@ def create_image(
...
+        Extended images ("demistoextended/...") map to "devtestdemistoextended/...".
+        A naive replace("demisto", "devtestdemisto") would corrupt them into
+        "devtestdemistoextended" via the inner "demisto", so extended images are
+        handled explicitly. All other images keep the existing "demisto" ->
+        "devtestdemisto" mapping.
+        """
+        if base_image.startswith(EXTENDED_REPOSITORY_SEGMENT):
+            renamed = base_image.replace("demistoextended", "devtestdemistoextended")
+        else:
+            renamed = base_image.replace("demisto", "devtestdemisto")
```
BEAdi: Use consts here (the ones that you already created)

demisto_sdk/commands/format/update_script.py
```diff
@@ -71,43 +72,91 @@ def update_docker_image_in_script(
...
+        ``_get_extended_client``. Returns an empty string on any failure (e.g. the
+        extended registry is not configured), so the caller leaves the image
+        unchanged instead of writing a broken tag.
+        """
+        # Imported lazily to avoid a heavy import on the common format hot path.
+        from demisto_sdk.commands.common.docker.docker_image import DockerImage
+
+        try:
+            image = DockerImage(dockerimage)
+            latest_tag = str(image.latest_tag)
```
BEAdi: why not use `get_latest_docker_image  here too?

demisto_sdk/commands/format/update_script.py
```diff
@@ -120,18 +120,13 @@ def _get_extended_image_latest_full_name(dockerimage: str) -> str:
...
+            return str(DockerImage(dockerimage).latest_docker_image)
         except Exception as e:
             logger.info(
                 f"<yellow>Failed getting latest tag for extended image "
                 f"{dockerimage}. {e}</yellow>"
             )
             return ""
-        if not latest_tag:
-            logger.info("<yellow>Failed getting docker image latest tag</yellow>")
-            return ""
```
BEAdi: Now nothing will be logged in this case

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -511,33 +558,26 @@ def create_image(
...
-            # double-prefixing it with the Docker Hub proxy registry.
+            # Return as-is if already prefixed, to avoid double-prefixing.
             if extended_registry and extended_registry in image:
                 return image
             if extended_registry and image.startswith(
-                (EXTENDED_REPOSITORY_SEGMENT, DEVTEST_EXTENDED_REPOSITORY_SEGMENT)
+                (
+                    EXTENDED_REPOSITORY_SEGMENT,
+                    DEVTEST_EXTENDED_REPOSITORY_SEGMENT,
+                )
```
BEAdi: What about `devdemistoextended`?

demisto_sdk/commands/common/docker/dockerhub_client.py
BEAdi, line 1: Remove the log lines that are not needed for the future

demisto_sdk/commands/common/docker/dockerhub_client.py
```diff
@@ -699,24 +728,36 @@ def get_gcloud_access_token() -> Optional[str]:
                    is caught and logged, but not re-raised.
     """
     try:
-        logger.debug("Trying to retrieve a Google Cloud access token.")
+        logger.debug(
+            "get_gcloud_access_token | resolving credentials via google.auth.default()"
+        )
         # Automatically obtain credentials from the environment
         credentials, project_id = google.auth.default()
+        service_account = getattr(credentials, "service_account_email", None)
```
BEAdi: Where is it used? I see it only in the debug logs

demisto_sdk/commands/common/docker/dockerhub_client.py
```diff
@@ -66,18 +65,14 @@ def __init__(
...
-        # Note: callers (DockerImage, DO104) always pass registry=DOCKER_REGISTRY_URL,
-        # which equals the GAR path in CI, so bool(registry) alone is insufficient.
+        # A custom registry is one explicitly provided by the user (e.g., JFrog):
+        # NOT Docker Hub (demisto/) and NOT a GAR target (demistoextended/), since
+        # GAR uses a gcloud bearer token and must never get Basic Auth.
         self._is_custom_registry = (
             bool(registry)
             and not IS_CONTENT_GITLAB_CI
             and self.DEFAULT_REGISTRY not in self.registry_api_url
+            and not _is_gar_registry(self.registry_api_url)
```
BEAdi: I'm not sure this is correct and will not break for customers, what is it needed for?

demisto_sdk/commands/common/docker_helper.py
YairGlick: I added

demisto_sdk/commands/common/docker/dockerhub_client.py
YairGlick: I removed

demisto_sdk/commands/common/docker/dockerhub_client.py
YairGlick: removed

demisto_sdk/commands/common/docker/dockerhub_client.py
YairGlick: removed

demisto_sdk/commands/common/constants.py
```diff
@@ -56,6 +56,11 @@
 CR_REGISTRY_PREFIX = f"{DEFAULT_EXTENDED_REGISTRY}/"
 
 
+def strip_cr_registry_prefix(image: str) -> str:
+    """TEMPORARY: strip the CR host prefix back to the canonical "demistoextended/" form."""
```
BEAdi: Open a ticket (can be a sub-task for your task) for you to remove this support from the sdk, and add the ticket number here in the 2 code comments that reference it (where you wrote TEMPORARY)

demisto_sdk/commands/common/docker/dockerhub_client.py
```diff
@@ -706,11 +708,14 @@ def get_gcloud_access_token() -> Optional[str]:
...
-            return None
+        logger.warning(
+            "get_gcloud_access_token | credentials refreshed but token is empty"
+        )
+        return None
     except Exception as e:
-        logger.debug(f"Failed to get access token: {str(e)}")
+        logger.warning(
+            f"get_gcloud_access_token | failed to get access token: {str(e)}"
+        )
```
BEAdi: Is this change mandatory?

demisto_sdk/commands/common/docker/docker_image.py
```diff
@@ -35,6 +41,33 @@ def _get_dockerhub_client(cls):
...
+            )
+            if not extended_registry:
+                return None
+            client = DockerHubClient(registry=extended_registry)
+            # get_registry_api_url() resolves to the GAR proxy in CI, so override
+            # with the actual registry V2 endpoint: https://{host}/v2/{path}
+            parts = extended_registry.rstrip("/").split("/", 1)
+            host = parts[0]
+            path = parts[1] if len(parts) > 1 else ""
+            client.registry_api_url = f"https://{host}/v2/{path}".rstrip("/")
```
BEAdi:
```suggestion
            host, *path = extended_registry.rstrip("/").split("/", 1)
            client.registry_api_url = f"https://{host}/v2{f'/{path[0]}' if path else ''}"
```
Image

demisto_sdk/commands/common/docker/docker_image.py
```diff
@@ -35,6 +41,33 @@ def _get_dockerhub_client(cls):
...
+            cls._extended_client = client
+        return cls._extended_client
+
+    def _get_client(self):
+        """Routes to the correct registry client based on repository prefix."""
+        if self.is_demistoextended_repository:
+            if extended_client := self._get_extended_client():
+                return extended_client
+            logger.warning(f"No DEMISTO_SDK_EXTENDED_REGISTRY configured for {self}")
```
BEAdi: The warning should be inside `_get_extended_client` where this condition is checked, and use the const value `DEMISTO_SDK_EXTENDED_REGISTRY_ENV` instead of writing it again

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -502,14 +561,28 @@ def create_image(
...
-        logger.debug(f"get_image_registry | returned: {image}")
+        image = strip_cr_registry_prefix(image)
+        # "demistoextended" images -> extended (GAR) registry; everything else -> Docker.
+        registry = (
+            os.getenv(DEMISTO_SDK_EXTENDED_REGISTRY_ENV, DEFAULT_EXTENDED_REGISTRY)
+            if DEMISTO_EXTENDED_REPOSITORY in image
+            else DOCKER_REGISTRY_URL
+        )
+        if registry and registry not in image:
+            return f"{registry}/{image}"
```
BEAdi, line 575: Why do you strip it and then re-add it? For the future, when we will not strip anymore?

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -584,9 +658,16 @@ def get_or_create_test_image(
                 )
             except (docker.errors.BuildError, docker.errors.APIError, Exception) as e:
                 errors = str(e)
-                logger.exception(  # noqa: PLE1205
-                    "{}", f"<red>{log_prompt} - Build errors occurred: {errors}</red>"
-                )
+                if EXTENDED_REPOSITORY_SEGMENT in base_image:
+                    # Quiet for GAR images; the caller decides skip (local) vs fail (CI).
```
BEAdi: How does this condition check whether we are in local or CI?

BEAdi (commented):
Looks good! Please see my small comments

demisto_sdk/commands/common/constants.py
YairGlick: opened a ticket

demisto_sdk/commands/common/docker_helper.py
YairGlick: Yes, the strip is only for the temporary case where the docker name already includes the CR prefix

demisto_sdk/commands/common/docker/docker_image.py
YairGlick: added

BEAdi (approved):

PR #5424, opened 2026-06-22
Fix GR109

shmuel44:
## Related Issues
fix: line 961 in the build

## Description
Fixed an issue where the GR109 (IsSupportedModulesCompatibility) validator
    falsely reported module-compatible commands (e.g. Set, Print, ReadFile) as incompatible
    when another validator had already loaded the content item's full set of USES relationships
    into the shared graph cache. The list of incompatible commands is now computed directly
    by the graph query instead of relying on the shared cache.


demisto_sdk/commands/validate/validators/GR_validators/GR109_is_supported_modules_compatibility.py
```diff
@@ -153,26 +153,29 @@ def get_missing_modules_by_command(self, content_item) -> dict[str, list[str]]:
         return missing_modules_by_item
 
     def get_commands_with_missing_modules_by_content_item(
-        self, item, commands_with_missing_modules_by_content_item: dict
+        self,
+        item,
+        mismatched_command_ids,
```
BEAdi: Add type hinting

demisto_sdk/commands/validate/tests/GR_validators_test.py
```diff
@@ -1788,6 +1788,138 @@ def test_SupportedModulesCompatibility_invalid_list_files_mismatch_playbook(
...
+                },
+            },
+        },
+    }
+    pack_a.create_playbook("playbook1", yml=playbook_yml)
+
+    return graph_repo
+
+
+def test_GR109_ignores_safe_commands_after_cache_pollution(
```
BEAdi, line 1883: Add an example of this test failing before the fix

demisto_sdk/commands/content_graph/interface/neo4j/queries/validations.py
BEAdi, line 544: Did you check all the places that are using this function? Specifically check GR114

BEAdi (commented):
Looks good, please add a build test of the failure before and after the fix, which shows the correct commands.

BEAdi (approved):

demisto_sdk/commands/content_graph/interface/neo4j/queries/validations.py
shmuel44: It should not break it

PR #5430, opened 2026-06-25
feat(prepare-content): support per-marketplace hidden values on integ…

rshunim:
…ration commands

Mirror the existing parameter-level behavior on commands: the command-level `hidden` field now accepts a list of marketplace names in addition to a plain boolean. At unify time, list-typed values are resolved to a boolean that matches the current marketplace - listed marketplaces become `hidden: true`, others become `hidden: false`. Pre-existing boolean values are left untouched.

The `xsoar` family aliases (`xsoar`, `xsoar_saas`, `xsoar_on_prem`) are expanded using the same rules as parameter-level hiding, so a single `xsoar` entry hides the command on all XSOAR-flavored marketplaces.

Adds 8 parametrized unit tests covering all six marketplaces plus the no-marketplace and missing-commands edge cases.

## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CRTX-260035


israelpoli (approved):
💪 Great

BEAdi (approved):

PR #5434, opened 2026-06-29
improve gr109 validator

Shir2611:
## Related Issues
fixes: oncall issue

## Description
Improve GR109 validation by leveraging get_content_item_supported_modules from tools.py


ShacharKidor (approved):
@Shir2611 - Looks good!

Just make sure that all existing unit tests regarding the GR109 validation and supported modules passes. 

Also, might worth adding a test for the specific case where supported modules are None (what causes the original issue with the new agentic skill object) - If we don't have any test for that case. 

Thanks.

PR #5437, opened 2026-07-02
Yc/managed field support suffix

yedidyacohenpalo:
## Related Issues
fixes: [CIAC-16730](https://jira-dc.paloaltonetworks.com/browse/CIAC-16730)

## Description
Adds support for marketplace-suffixed `managed` and `source` pack metadata fields (e.g. `managed:platform`, `source:platform`), allowing a pack to declare a per-marketplace `managed`/`source` value.

**What changed:**
- `MarketplaceSuffixPreparer.prepare_managed_and_source` resolves the suffixed `managed`/`source` keys into the plain `managed`/`source` fields for the current marketplace, with validation (invalid suffix, suffixed `managed` without a plain default, `source` without `managed`, managed-without-source, source-while-unmanaged).
- Marketplaces in `ALWAYS_UNMANAGED_MARKETPLACES` are forced to `managed: false` with `source` removed.
- `Pack.dump_metadata` and `Pack._dump_pack_metadata` call the resolver so the dumped `metadata.json` and the copied `pack_metadata.json` stay consistent.
- The suffixed fields are declared on the parser (`PackMetadataParser`) and the strict model (`StrictPackMetadata`) so they pass structure validation (`extra = forbid`).

**Tests:** parametrized valid/invalid cases in `marketplace_suffix_preparer_test.py`, plus dump/strip-internal cases in `upload_internal_field_test.py`.


demisto_sdk/commands/content_graph/parsers/pack.py
```diff
@@ -205,6 +205,11 @@ def __init__(self, path: Path, metadata: Dict[str, Any]) -> None:
         self.source: str = metadata.get("source", "")
         self.managed: bool = metadata.get("managed", False)
         self.internal: bool = metadata.get("internal", False)
+        # Marketplace-suffixed managed/source fields. Kept as-is here; they are
```
ShacharKidor: Not sure why the fields are under the "# For private packs" section here, but in demisto_sdk/commands/content_graph/objects/pack_metadata.py they aren't.

demisto_sdk/commands/content_graph/strict_objects/pack_meta_data.py
```diff
@@ -103,3 +103,8 @@ def validate_managed_pack_has_source(cls, values):
     source: Optional[str] = Field("", alias="source")
     managed: Optional[bool] = Field(False, alias="managed")
     internal: Optional[bool] = Field(False, alias="internal")
+
+    # Marketplace-suffixed managed/source fields (e.g. ``managed:platform``).
+    # Resolved into the plain managed/source per-marketplace during dump.
```
ShacharKidor, line 108: Why are we adding those fields as strict objects? 
IIRC, they aren't mandatory...

demisto_sdk/commands/content_graph/objects/pack.py
```diff
@@ -302,7 +310,13 @@ def _dump_pack_metadata(
...
             return
 
-        if pack_metadata.pop("internal", None):
+        # Resolve marketplace-suffixed managed/source fields for the current
+        # marketplace, consistent with the dumped metadata.json.
+        pack_metadata = MarketplaceSuffixPreparer.prepare_managed_and_source(
+            pack_metadata, marketplace
+        )
+
+        if strip_internal and pack_metadata.pop("internal", None):
```
ShacharKidor, line 319: Why did we change the 'strip internal' logic? please explain.

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
```diff
@@ -5,6 +5,20 @@
 
 SEPARATOR = ":"
 
+MANAGED_KEY = "managed"
+SOURCE_KEY = "source"
+
+# Marketplaces that are always considered unmanaged. For these marketplaces the
+# pack is forced to `managed: false` and the `source` field is removed, without
+# any further suffix resolution.
```
ShacharKidor:
```suggestion
# Marketplaces designated as unmanaged. For these marketplaces, 
# packs are forced to `managed: false` and the `source` field is removed, without
# any further suffix resolution.
```

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
```diff
@@ -89,3 +103,143 @@ def fix_recursively(datum: Any) -> Any:
...
+    def prepare_managed_and_source(
+        data: dict,
+        current_marketplace: MarketplaceVersions,
+    ) -> dict:
+        """
+        Resolves the marketplace-suffixed ``managed`` and ``source`` fields of a
+        pack metadata into their final, marketplace-specific values.
+
+        The pack metadata may contain ``managed``/``source`` keys with an optional
+        marketplace suffix (e.g. ``managed:platform``, ``source:platform``). This
```
ShacharKidor:
```suggestion
        marketplace suffix (e.g. ``managed:platform:true``, ``source:platform:foo``). This
```

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
```diff
@@ -89,3 +103,143 @@ def fix_recursively(datum: Any) -> Any:
...
+        - A ``source`` (plain or suffixed) without any ``managed`` at all.
+        - A resolved ``managed: true`` without a resolved ``source``.
+        - A resolved ``source`` while the resolved ``managed`` is ``false``.
+
+        Args:
+            data: The pack metadata dictionary.
+            current_marketplace: The marketplace the pack is being prepared for.
+
+        Returns:
+            The (possibly modified) pack metadata dictionary.
```
ShacharKidor: What do you mean by 'possibly modified'?

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
```diff
@@ -89,3 +103,143 @@ def fix_recursively(datum: Any) -> Any:
...
+                    collected[suffix] = key
+            return collected
+
+        def remove_managed_and_source_keys() -> None:
+            """Removes every ``managed``/``source`` key from ``data`` - both the
+            plain keys and any suffixed ones (including always-unmanaged and
+            ``None``-valued keys). The resolved values are set back afterwards."""
+            for key in list(data.keys()):
+                for field in (MANAGED_KEY, SOURCE_KEY):
+                    if key == field or key.startswith(f"{field}{SEPARATOR}"):
```
ShacharKidor:
```suggestion
               if key in (MANAGED_KEY, SOURCE_KEY) or key.startswith(f"{MANAGED_KEY}{SEPARATOR}") or key.startswith(f"{SOURCE_KEY}{SEPARATOR}"):
```

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
```diff
@@ -89,3 +103,143 @@ def fix_recursively(datum: Any) -> Any:
...
+            )
+
+        if not managed_keys:
+            return data
+
+        # A suffixed managed requires a plain managed as the default for all
+        # other marketplaces.
+        if any(managed_keys) and "" not in managed_keys:
+            raise ValueError(
+                "Pack metadata has a marketplace-suffixed 'managed' field but is "
```
ShacharKidor: not critical, but might be worth adding the managed suffixed field itself

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
```diff
@@ -89,3 +103,143 @@ def fix_recursively(datum: Any) -> Any:
...
+
+            Keys whose value is ``None`` are treated as absent (they only exist
+            because the model declares them). An invalid suffix raises.
+            """
+            collected: dict = {}
+            for key in data:
+                if data[key] is None:
+                    continue
+                if key == field:
+                    collected[""] = key
```
ShacharKidor: Maybe I'm wrong but shouldn't it be collected[""] = data[key] 
Meaning: where do we save the value itself? 

Let's discuss that

demisto_sdk/commands/content_graph/tests/upload_internal_field_test.py
```diff
@@ -428,8 +432,9 @@ def test_dump_pack_metadata_keeps_internal_when_flag_false(tmp_path):
           ``strip_internal`` (the default for prepare-content / artifact
           builds).
     Then:
-        - The destination ``pack_metadata.json`` is a verbatim copy of the
-          source - the ``internal`` field is preserved.
+        - The destination ``pack_metadata.json`` preserves the ``internal``
```
ShacharKidor, line 435: Do we have new tests for all the cases of the managed and source fields? 
If not, let's add them

ShacharKidor (changes requested):
Great Work! 

Please add a short description to the PR body, and also add a changelog.

ShacharKidor:
@israelpoli or @tcarmeli1 I would appriciate that if one of you could have a look too.

demisto_sdk/commands/content_graph/parsers/pack.py
yedidyacohenpalo: Good catch - not private-pack fields. Separated them into their own block, fixed in next push.

demisto_sdk/commands/content_graph/strict_objects/pack_meta_data.py
yedidyacohenpalo: They're optional (default `None`). But `BaseStrictModel` uses `extra=forbid`, so `managed:platform`/`source:platform` would fail structure validation unless declared. Declaring them just allows the keys, it doesn't make them required.

demisto_sdk/commands/content_graph/objects/pack.py
yedidyacohenpalo: To stay consistent with `dump_metadata`: we resolve managed/source first, and only `pop('internal')` + log when it's actually present, avoiding a misleading log. `strip_internal=False` behavior is unchanged.

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
yedidyacohenpalo: Intentional - `collect_keys` maps suffix -> key name (not value); the value is read later via `data.get(...)`. Keeping the key name is what lets us pop the exact keys afterward. Can add an inline comment if helpful.

demisto_sdk/commands/content_graph/tests/upload_internal_field_test.py
yedidyacohenpalo: Covered by the parametrized `test_prepare_managed_and_source_valid`/`_invalid` (override, always-unmanaged, managed-without-source, source-without-managed, invalid suffix, missing plain managed). Let me know if a specific case is missing.

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
yedidyacohenpalo: Applied in the next push. Thanks!

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
yedidyacohenpalo: Applied in the next push. Thanks!

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
yedidyacohenpalo: Clarified the wording in the docstring - it just means the same dict is returned (unchanged when neither managed nor source is present). Applied in the next push.

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
yedidyacohenpalo: Applied in the next push. Thanks!

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
yedidyacohenpalo: Good idea - added the offending suffixed field name to the error message. Applied in the next push.

ShacharKidor (approved):

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
```diff
@@ -89,3 +103,151 @@ def fix_recursively(datum: Any) -> Any:
...
+            unchanged when neither ``managed`` nor ``source`` is present.
+        """
+        valid_suffixes = {
+            f"{SEPARATOR}{mp.value}"
+            for mp in MarketplaceVersions
+            if mp not in ALWAYS_UNMANAGED_MARKETPLACES
+        }
+
+        def collect_keys(field: str) -> dict:
+            """Maps each set key of ``field`` to its suffix (``""`` for the plain
```
ShacharKidor: TBH this docstring is unclear to me... would rather have one like: "Goes over the pack metadata fields and does this and that..." 
Also - add a description to the arg 'field' itself, which field is it?

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
```diff
@@ -89,3 +103,151 @@ def fix_recursively(datum: Any) -> Any:
...
+                    if suffix not in valid_suffixes:
+                        raise ValueError(
+                            f"Invalid marketplace suffix in pack metadata field "
+                            f"'{key}'. The suffix must be one of: "
+                            f"{sorted(s.lstrip(SEPARATOR) for s in valid_suffixes)}."
+                        )
+                    collected[suffix] = key
+            return collected
+
+        def remove_managed_and_source_keys() -> None:
```
ShacharKidor: It would be better to move the inner functions here outside the prepare_managed_and_source function for better readability.

demisto_sdk/commands/prepare_content/preparers/marketplace_suffix_preparer.py
```diff
@@ -89,3 +103,151 @@ def fix_recursively(datum: Any) -> Any:
...
+            resolved to their plain, marketplace-specific values. It is returned
+            unchanged when neither ``managed`` nor ``source`` is present.
+        """
+        valid_suffixes = {
+            f"{SEPARATOR}{mp.value}"
+            for mp in MarketplaceVersions
+            if mp not in ALWAYS_UNMANAGED_MARKETPLACES
+        }
+
+        def collect_keys(field: str) -> dict:
```
ShacharKidor: this function should have a better and more specific to the case name.

demisto_sdk/commands/validate/validators/RM_validators/RM109_is_readme_exists.py
```diff
@@ -46,6 +47,6 @@ def obtain_invalid_content_items(
                 )
                 and (not content_item.readme.exist)
                 and (not content_item.is_silent)
-                and getattr(content_item, "source", "") != "autonomous"
+                and not is_autonomous_pack(content_item.in_pack)
```
ShacharKidor, line 50: I want to ask you something about that.

demisto_sdk/commands/validate/tools.py
```diff
@@ -37,6 +37,53 @@
...
+
+    Args:
+        pack: The pack object (may carry suffixed managed/source attributes).
+
+    Returns:
+        tuple: (managed, source) resolved for the platform marketplace.
+    """
+    managed = pack.managed_platform
+    if managed is None:
+        managed = pack.managed
```
ShacharKidor, line 62: Do we have a case where the managed platform field not set, but we have managed:koi:true and managed:true?

demisto_sdk/commands/validate/tools.py
```diff
@@ -37,6 +37,53 @@
...
+
+    source = pack.source_platform
+    if source is None:
+        source = pack.source
+
+    return bool(managed), source or ""
+
+
+def is_autonomous_pack(pack: Optional[Pack]) -> bool:
```
ShacharKidor, line 74: clean and elegant solution for all the validations - nice.

ShacharKidor (changes requested):
Nice! 

Let's go over the minor comments together before the merge.

demisto_sdk/commands/validate/tools.py
```diff
@@ -37,6 +37,53 @@
...
+    Args:
+        pack: The pack object (may carry suffixed managed/source attributes).
+
+    Returns:
+        tuple: (managed, source) resolved for the platform marketplace.
+    """
+    managed = pack.managed_platform
+    if managed is None:
+        managed = pack.managed
+
```
ShacharKidor, line 63: The logic should be: If managed platform is False - source is "".

ShacharKidor (approved):

PR #5438, opened 2026-07-02
CIAC 17053 - Add BA133 validation for fetch-type supported modules

dtroushinsky:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-17053

## Description
Adds the BA133 validation, which ensures an integration's fetch-related configuration parameters exist as required by their fetch type and resolve to `supportedModules` consistent with that fetch type (e.g. Fetch Events → xsiam only; Fetch Assets → xsiam + exposure_management). Auto-fixable. Warning-only on the nightly.


demisto_sdk/commands/validate/validators/BA_validators/BA133_invalid_supported_modules_for_fetch_type.py
```diff
@@ -0,0 +1,462 @@
...
+        "display": "Fetch incidents",
+        "type": 8,
+        "required": False,
+        "section": "Collect",
+    },
+    "incidentType": {
+        "display": "Incident type",
+        "type": 13,
+        "required": False,
+        "section": "Connect",
```
kamalq97:
```suggestion
        "section": "Collect",
```

kamalq97 (approved):
Approved from my end! 
The logic appears solid and matches the spec.

yedidyacohenpalo (approved):

demisto_sdk/commands/validate/validators/BA_validators/BA133_invalid_supported_modules_for_fetch_type.py
```diff
@@ -0,0 +1,462 @@
...
+        allowed_modules=XSIAM_ONLY_MODULES,
+        required_params={"isFetchEvents", "eventFetchInterval"},
+        optional_params={MAX_FETCH_PARAM},
+    ),
+    FetchType(
+        flag_attr="is_fetch_assets",
+        display_name="Fetch Assets",
+        allowed_modules=XSIAM_AND_EXPOSURE_MANAGEMENT_MODULES,
+        required_params={"isFetchAssets", "assetsFetchInterval"},
+    ),
```
BEAdi, line 209: Have you checked this?

demisto_sdk/commands/validate/validators/BA_validators/BA133_invalid_supported_modules_for_fetch_type.py
```diff
@@ -0,0 +1,462 @@
...
+        "type": 19,
+        "required": False,
+        "defaultvalue": "240",
+        "section": "Collect",
+    },
+    "feedExpirationPolicy": {
+        "display": "",
+        "type": 17,
+        "required": False,
+        "options": ["never", "interval", "indicatorType", "suddenDeath"],
```
BEAdi, line 129: Sure about "indicatorType" and "suddenDeath"?

demisto_sdk/commands/validate/validators/BA_validators/BA133_invalid_supported_modules_for_fetch_type.py
```diff
@@ -0,0 +1,462 @@
...
+        "required": False,
+        "additionalinfo": "Supports CSV values.",
+        "section": "Collect",
+    },
+    "tlp_color": {
+        "display": "Traffic Light Protocol Color",
+        "type": 15,
+        "required": False,
+        "additionalinfo": "The Traffic Light Protocol (TLP) designation to apply to indicators fetched from the feed",
+        "options": ["RED", "AMBER", "GREEN", "WHITE"],
```
BEAdi: There is also the "AMBER+STRICT" option for TLP color, but I'm not sure if it is used in content, please verify that

demisto_sdk/commands/validate/validators/BA_validators/BA133_invalid_supported_modules_for_fetch_type.py
```diff
@@ -0,0 +1,462 @@
...
+            "feedBypassExclusionList",
+            "feedTags",
+            "tlp_color",
+        },
+        optional_params={"feedIncremental", "first_fetch"},
+    ),
+    FetchType(
+        activation_param=IS_FETCH_CREDENTIALS_PARAM,
+        display_name="Fetch Credentials",
+        allowed_modules=XSIAM_AND_AGENTIX_MODULES,
```
BEAdi, line 230: Have you checked this?

BEAdi (approved):

demisto_sdk/commands/validate/validators/BA_validators/BA133_invalid_supported_modules_for_fetch_type.py
dtroushinsky: Yes, I verified the Fetch Assets functionality. In a live run I confirmed:
- A correctly-configured asset-fetching integration passes with no false positives.
- Only xsiam and exposure_management are accepted; any other module is flagged — including modules like agentix that are valid for other fetch types but not for Assets — and the error clearly lists the allowed modules.
- Multiple disallowed modules on one parameter are all reported together.
- Both required parameters (isFetchAssets and assetsFetchInterval) are checked individually; if either is missing, it's reported by name against the "Fetch Assets" type.

The Fetch Assets type — activation, its allowed-module set, and both required parameters — behaves correctly.

demisto_sdk/commands/validate/validators/BA_validators/BA133_invalid_supported_modules_for_fetch_type.py
dtroushinsky: I'm updating the auto-fix template to inject the newer set (RED, AMBER+STRICT, AMBER, GREEN, CLEAR, default CLEAR).

demisto_sdk/commands/validate/validators/BA_validators/BA133_invalid_supported_modules_for_fetch_type.py
dtroushinsky: Yes, I checked the Fetch Credentials type. I verified that:
- An integration with an invalid module on isFetchCredentials (e.g. edr) is flagged, with a clear message listing the allowed modules (agentix, xsiam).
- An optional credentials param with an invalid module (e.g. secrets with edr) is also flagged.
- A valid config (isFetchCredentials with xsiam) passes with no false positives.
- Activation works by param existence, not a script flag: when isFetchCredentials is present the credentials params are checked, and when it's absent they're correctly skipped.
- Credentials has no required params, so only invalid-module errors apply — there's no missing-required path for it.

So the Fetch Credentials type behaves as expected.

demisto_sdk/commands/validate/validators/BA_validators/BA133_invalid_supported_modules_for_fetch_type.py
dtroushinsky: Yes — indicatorType and suddenDeath are correct. I verified them against the actual feedExpirationPolicy parameters across the content repo, and both the option list and the exact camelCase spelling match.

demisto_sdk/commands/validate/validators/BA_validators/BA133_invalid_supported_modules_for_fetch_type.py
BEAdi: My question is actually on the assumption here- have you checked on a machine with one of those licenses that it works on it, and that it really doesn't work on a machine without those licenses?

PR #5451, opened 2026-07-14
fix(content_graph): compute git-diff after import so interface.commit…

rshunim:
… is populated

Fixes a silent no-op in validate -g where a modified pack (e.g. deprecation flip on an integration) was ignored by the graph, causing graph validations such as GR107 to fire against unrelated content items with stale deprecated=false values inherited from the bucket's master state.

Root cause
----------
_update_content_graph_inner computed the git-diff BEFORE calling import_graph. ContentGraphInterface.commit reads the pinned commit from metadata.json under the import directory - a file only written by import_graph. On a fresh CI runner with an empty import dir, interface.commit therefore returned None at diff time, the diff was skipped, and after the connectors refactor added a hard early-return on empty inputs to ContentGraphBuilder.update_graph, the whole update pipeline silently returned without touching the graph.

Fix
---
1. Move the git-diff computation to AFTER import_graph populates metadata.json (and therefore interface.commit).
2. Add a safety-net: if use_git=True and interface.commit is still None post-import (corrupted metadata, shallow clone where the pinned commit is unknown, etc.), fall back to create_content_graph (full rebuild) instead of no-op'ing with a stale bucket graph. Mirrors the existing git_diff_failed handler.
3. Add a debug log in ContentGraphBuilder.update_graph's empty-input early-return so future silent no-ops are visible in debug logs.
4. Add regression tests covering both the post-import diff and the safety-net fallback.

## Related Issues
fixes:

### DIFF
--- a/.changelog/5451.yml
+++ b/.changelog/5451.yml
@@ -1,0 +1,10 @@
+changes:
+- description: >-
+    Fixed an issue where **validate -g** silently skipped applying git-diff
+    updates to the content graph when the imported graph had no pinned
+    commit yet (e.g. fresh CI runner). The diff is now computed after
+    **import_graph** populates the commit, and a full-rebuild fallback
+    protects against the same silent no-op if the commit is still unknown
+    post-import.
+  type: fix
+pr_number: 5451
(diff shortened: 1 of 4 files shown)


rshunim:
SDK Nightly on this branch (Content and Infra are master) 

https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/pipelines/11024823

PR #5458, opened 2026-07-16
retry failed docker image pull

BarGali:
test changes https://github.com/demisto/content/pull/45097


demisto_sdk/commands/common/docker_helper.py
BEAdi, line 1: Please add UTs for the changes that you added

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -444,6 +447,87 @@ def push_image(self, image: str, log_prompt: str = "") -> None:
...
+            )
+
+        # After a successful push, verify the image is pullable from the registry.
+        # Registry propagation can take a few minutes, so we retry with delays.
+        self._verify_image_available_after_push(
+            test_image_name_to_push, log_prompt=log_prompt
+        )
+
+    @staticmethod
+    def _verify_image_available_after_push(
```
BEAdi, line 567: Can we use `is_image_available` instead?

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -444,6 +447,87 @@ def push_image(self, image: str, log_prompt: str = "") -> None:
...
+            image (str): The image name (without registry prefix), e.g.
+                ``devtestdemisto/python3:3.10.0.12345-abcdef``.
+            log_prompt (str): Log prompt prefix for messages.
+            max_retries (int): Maximum number of verification attempts. Defaults to 10.
+            delay_seconds (int): Seconds to wait between retries. Defaults to 30.
+        """
+        if ":" not in image:
+            repo, tag = image, "latest"
+        else:
+            repo, tag = image.rsplit(":", 1)
```
BEAdi: We shouldn't allow images with more than 1 `:`, like in `is_image_available`

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -444,6 +447,87 @@ def push_image(self, image: str, log_prompt: str = "") -> None:
...
+
+        logger.info(
+            f"{log_prompt} - Verifying pushed image {image} is available on DockerHub "
+            f"(up to {max_retries} attempts, {delay_seconds}s apart)."
+        )
+
+        for attempt in range(1, max_retries + 1):
+            try:
+                token = _get_docker_hub_token(repo)
+                if not token:
```
BEAdi: `_get_docker_hub_token` raises `RuntimeError` on failure and shouldn't return a falsy value, so the `if not token` can't be reached and can it be removed

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -444,6 +447,87 @@ def push_image(self, image: str, log_prompt: str = "") -> None:
...
+                    f"failed due to {type(e).__name__}. Retrying in {delay_seconds}s...",
+                    exc_info=True,
+                )
+
+            if attempt < max_retries:
+                time.sleep(delay_seconds)
+
+        raise DockerException(
+            f"{log_prompt} - Image verification failed: {image} was not found on DockerHub "
+            f"after {max_retries} attempts ({max_retries * delay_seconds}s total). "
```
BEAdi: `_get_docker_hub_token` also has retry configured, so the number of seconds might not be accurate. The log can mention that this is the delay time and not necessarily the run time

.changelog/5458.yml
BEAdi, line 1: [AI comment]
Please shorten the descriptions to state the new behaviour rather than the implementation detail. e.g. drop "in DockerBase.push_image" / "registry propagation before proceeding" and phrase it as user-facing behaviour (docker image push now verifies the image is pullable and fails loudly if the push did not succeed).

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -444,6 +447,87 @@ def push_image(self, image: str, log_prompt: str = "") -> None:
...
+            repo, tag = image.rsplit(":", 1)
+
+        logger.info(
+            f"{log_prompt} - Verifying pushed image {image} is available on DockerHub "
+            f"(up to {max_retries} attempts, {delay_seconds}s apart)."
+        )
+
+        for attempt in range(1, max_retries + 1):
+            try:
+                token = _get_docker_hub_token(repo)
```
BEAdi: The push should also be verified for `devtestdemistoextended`, which is pushed to GAR and not to DockerHub (use `get_image_registry`). We can add this option to `is_image_available`

BEAdi (commented):
Looks good! Please see my comments

demisto_sdk/commands/common/tests/docker_helper_test.py
BEAdi, line 1: Please retrieve the removed tests

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -503,6 +552,102 @@ def push_image(self, image: str, log_prompt: str = "") -> None:
...
+            try:
+                if is_gar_image:
+                    if not DockerBase.is_image_available(
+                        image, use_registry_prefix=True
+                    ):
+                        raise RuntimeError(
+                            f"Image {image} not yet available in registry"
+                        )
+                else:
+                    DockerBase._is_image_available_on_registry(repo, tag)
```
BEAdi: You need to check the return value from this function call.

[AI comment]
Add a case where `_is_image_available_on_registry` returns `False` on every attempt and assert that it retries and finally raises `DockerException`.

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -369,10 +370,57 @@ def pull_image(image: str) -> docker.models.images.Image:
...
+            bool: True if the image is available.
+        """
+        if use_registry_prefix:
+            registry_image = DockerBase.get_image_registry(image)
+            try:
+                DockerBase.pull_image(registry_image)
+                return True
+            except (docker.errors.NotFound, docker.errors.ImageNotFound):
+                logger.debug(f"Image {registry_image} not found in registry")
+                return False
```
BEAdi, line 422: [AI comment]
`is_image_available(..., use_registry_prefix=True)` only swallows `NotFound/ImageNotFound` and returns `False`; any other `docker.errors.DockerException` (e.g. a failed `gar_daemon_login / APIError`) propagates out of the loop and is not caught by the `except RuntimeError / except (ConnectionError, Timeout, RequestException)` handlers, so it will crash instead of retrying. Was that intended, or should a failed GAR pull also be retried?

demisto_sdk/commands/common/docker_helper.py
```diff
@@ -503,6 +552,102 @@ def push_image(self, image: str, log_prompt: str = "") -> None:
...
+                    f"{log_prompt} - Verification attempt {attempt}/{max_retries}: "
+                    f"failed due to {type(e).__name__}. Retrying in {delay_seconds}s...",
+                    exc_info=True,
+                )
+
+            if attempt < max_retries:
+                time.sleep(delay_seconds)
+
+        raise DockerException(
+            f"{log_prompt} - Image verification failed: {image} was not found on DockerHub "
```
BEAdi: [AI comment]
You compute registry_name = "the registry (GAR)", but the retry log ("not yet available on DockerHub") and the final DockerException ("was not found on DockerHub") are still hardcoded to DockerHub. For a GAR image these will be misleading. Suggest reusing registry_name in both messages.

BEAdi (approved):

PR #5461, opened 2026-07-20
Remove source auto-add (format) and MC100 validation, keep source fie…

yedidyacohenpalo:
## Related Issues
related to: [CIAC-16730](https://jira-dc.paloaltonetworks.com/browse/CIAC-16730)

## Description
This PR is a prerequisite for the main [PR](https://github.com/demisto/demisto-sdk/pull/5437/) that introduces the managed-per-marketplace feature.

Removed the automatic addition of the **source** field in the **format** command, and removed the **MC100** validation.
Since having a `source` field in managed content items is no longer needed. 
The source is now determined according to the pack metadata file only.


.changelog/5461.yml
```diff
@@ -0,0 +1,4 @@
+changes:
+- description: Removed the automatic addition of the **source** field in the **format** command, and removed the **MC100** validation.
```
ShacharKidor, line 2:
```suggestion
- description: Removed the automatic addition of the **source** field to managed content items from the **format** command, and removed the **MC100** validation that validates the existence of a **source** field in managed content items.
```

demisto_sdk/commands/format/update_generic.py
```diff
@@ -473,45 +473,6 @@ def check_server_version(self):
                     self.data[self.from_version_key] = current_from_server_version
                     self.data.pop(self.json_from_server_version_key)
 
-    def add_source_field_for_managed_pack(self):
```
ShacharKidor, line 476: I can see that we call this function only for playbooks and scripts. What about other content items such as integrations? 
I just want to be sure that we are not missing another function that should be removed in addition to that one.

ShacharKidor (changes requested):
Nice!
Added two comments - please review. 

Also, please update the issue body with a description and links to the relevant ticket / other related PRs.

demisto_sdk/commands/format/update_generic.py
yedidyacohenpalo: Checked - it was only called from the playbook and script formatters. Integrations and other content items never used it, nothing else to remove.

ShacharKidor (approved):

PR #5469, opened 2026-07-30
demisto-sdk upload-custom-integration

Shir2611:
## Related Issues
fixes:
[CIAC-17163](https://jira-dc.paloaltonetworks.com/browse/CIAC-17163)

## Description
Added a new `demisto-sdk upload-custom-integration` command that safely uploads a custom integration to the Cortex Platform. Before uploading, it verifies that both `commonfields.id` and `name` in the integration YAML end with the `_copy` suffix, preventing ID conflicts with system pack integrations that would otherwise cause system pack downloads to fail. Accepts a YAML file path or an integration directory. Pass `--force-id` to bypass the check with a warning when the ID is known to be unique.

image

image


demisto_sdk/__main__.py
```diff
@@ -172,6 +172,23 @@ def register_commands(_args: list[str] = []):  # noqa: C901
...
+        from demisto_sdk.commands.upload.upload_custom_integration_setup import (
+            upload_custom_integration,
+        )
+
+        app.command(
+            name="upload-custom-integration",
+            help=(
+                "Upload a custom integration to Cortex Platform with '_copy' marker "
+                "safety enforcement. Targets the Cortex Platform marketplace exclusively "
+                "(marketplace=platform, cannot be overridden). "
```
ShacharKidor: Regarding the 
```
Targets the Cortex Platform marketplace exclusively "
                "(marketplace=platform, cannot be overridden). "
```
I haven't seen a reference for that in the code - do we really pass mp=platform? 
Please check that it is needed, and if so, let's add it.

demisto_sdk/commands/upload/upload_custom_integration_setup.py
```diff
@@ -0,0 +1,83 @@
+"""
+Typer CLI setup for the ``upload-custom-integration`` command.
+
+Exposes exactly two user-facing flags:
+  -i / --input    Path to the integration YAML file or directory.
+  --force-id      Bypass the ``_copy`` marker check (with a warning).
```
ShacharKidor, line 6: Let's add here in the doc that this arg is not recommended.

demisto_sdk/commands/upload/upload_custom_integration_setup.py
```diff
@@ -0,0 +1,83 @@
...
+
+All other upload options (marketplace, zip, insecure, etc.) are handled
+internally by the core upload pipeline with its defaults.
+
+The existing ``demisto-sdk upload`` command is **not modified** by this module.
```
ShacharKidor:
```suggestion

```

demisto_sdk/commands/upload/upload_custom_integration_setup.py
```diff
@@ -0,0 +1,83 @@
...
+    ctx: typer.Context,
+    input_path: Path = typer.Option(
+        ...,
+        "--input",
+        "-i",
+        exists=True,
+        resolve_path=True,
+        help=(
+            "Path to the integration YAML file or its parent directory. "
+            "The file's 'commonfields.id' and 'name' must end with '_copy' "
```
ShacharKidor: 'commonfields.id' and 'name' - isn't name also a common field? I'm not sure why they're different. 
It appears in several places throughout the PR.

demisto_sdk/commands/upload/upload_custom_integration_setup.py
```diff
@@ -0,0 +1,83 @@
...
+            "unless --force-id is passed."
```
ShacharKidor:
```suggestion
```

demisto_sdk/commands/upload/upload_custom_integration_setup.py
```diff
@@ -0,0 +1,83 @@
...
+        ),
+    ),
+    force_id: bool = typer.Option(
+        False,
+        "--force-id",
+        help=(
+            "Bypass the '_copy' marker validation. "
+            "WARNING: Uploading without '_copy' risks conflicting with "
+            "official system pack IDs and may cause download failures."
```
ShacharKidor:
```suggestion
            "official system pack IDs and may cause pack installation failures."
```

demisto_sdk/commands/upload/upload_custom_integration_setup.py
```diff
@@ -0,0 +1,83 @@
...
+    # The three options below are injected by @logging_setup_decorator and must
+    # be present in the signature, but are hidden from --help output so the
+    # user only sees -i/--input and --force-id.
+    console_log_threshold: str = typer.Option(
+        None, "--console-log-threshold", hidden=True
+    ),
+    file_log_threshold: str = typer.Option(None, "--file-log-threshold", hidden=True),
+    log_file_path: str = typer.Option(None, "--log-file-path", hidden=True),
+) -> None:
+    """Upload a custom integration to Cortex XSOAR/XSIAM with safety enforcement.
```
ShacharKidor: why XSOAR/XSIAM?
```suggestion
    """Upload a custom integration to Cortex Platform with safety enforcement.
```

demisto_sdk/commands/upload/upload_custom_integration_setup.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Validates that the integration's ``commonfields.id`` and ``name`` fields
+    end with the ``_copy`` marker before uploading, preventing ID conflicts
+    with official system pack integrations.
+
+    \b
+    RISK: If a custom integration is uploaded with an ID that matches a system
+    integration's ID, subsequent downloads of the system pack containing that
```
ShacharKidor:
```suggestion
    integration's ID, subsequent installations of the system pack containing that
```

demisto_sdk/commands/upload/upload_custom_integration_setup.py
```diff
@@ -0,0 +1,83 @@
...
+    integration will fail with a system error.
+
+    \b
+    ENVIRONMENT VARIABLES:
+      DEMISTO_BASE_URL   Cortex XSOAR/XSIAM instance URL (required)
```
ShacharKidor:
```suggestion
      DEMISTO_BASE_URL   Cortex Platform instance URL (required)
```

demisto_sdk/commands/upload/upload_custom_integration_setup.py
```diff
@@ -0,0 +1,83 @@
...
+    force_id: bool = typer.Option(
+        False,
+        "--force-id",
+        help=(
+            "Bypass the '_copy' marker validation. "
+            "WARNING: Uploading without '_copy' risks conflicting with "
+            "official system pack IDs and may cause download failures."
+        ),
+    ),
+    # The three options below are injected by @logging_setup_decorator and must
```
ShacharKidor, line 54: why do we add those?

demisto_sdk/commands/upload/upload_custom_integration_setup.py
```diff
@@ -0,0 +1,83 @@
...
+    \b
+    RISK: If a custom integration is uploaded with an ID that matches a system
+    integration's ID, subsequent downloads of the system pack containing that
+    integration will fail with a system error.
+
+    \b
+    ENVIRONMENT VARIABLES:
+      DEMISTO_BASE_URL   Cortex XSOAR/XSIAM instance URL (required)
+      DEMISTO_API_KEY    Valid API key for the instance (required)
+      XSIAM_AUTH_ID      Auth ID for XSIAM or XSOAR 8.x (optional)
```
ShacharKidor: I think it is required for the platform
```suggestion
      XSIAM_AUTH_ID      Auth ID for Cortex Platform (required)
```

demisto_sdk/commands/upload/upload_custom_integration.py
```diff
@@ -0,0 +1,72 @@
+"""
+Core logic for the ``upload-custom-integration`` command.
+
+This module is a thin safety wrapper around the existing
+:func:`~demisto_sdk.commands.upload.upload.upload_content_entity` function.
+It enforces the ``_copy`` naming convention on integration IDs and display
+names **before** delegating to the standard upload pipeline.
```
ShacharKidor:
```suggestion
It enforces the ``_copy`` naming convention on integration IDs and
name **before** delegating to the standard upload pipeline.
```

demisto_sdk/commands/upload/upload_custom_integration.py
```diff
@@ -0,0 +1,72 @@
...
+
+The existing ``demisto-sdk upload`` command is **not modified** by this module.
```
ShacharKidor:
```suggestion
```

demisto_sdk/commands/upload/upload_custom_integration.py
```diff
@@ -0,0 +1,72 @@
...
+    # Resolve directory → YAML file transparently so that both
+    #   -i Integrations/MyIntegration_copy/
+    #   -i Integrations/MyIntegration_copy/MyIntegration_copy.yml
+    # work identically.
+    resolved_path = resolve_integration_yaml(input_path)
+
+    if not is_integration_yaml(resolved_path):
+        raise typer.BadParameter(
+            f"'{input_path}' does not appear to be an integration YAML "
+            f"(missing 'commonfields' / 'script'+'category' keys). "
```
ShacharKidor:
```suggestion
"
```

demisto_sdk/commands/upload/upload_custom_integration.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # Resolve directory → YAML file transparently so that both
+    #   -i Integrations/MyIntegration_copy/
+    #   -i Integrations/MyIntegration_copy/MyIntegration_copy.yml
+    # work identically.
+    resolved_path = resolve_integration_yaml(input_path)
+
+    if not is_integration_yaml(resolved_path):
+        raise typer.BadParameter(
+            f"'{input_path}' does not appear to be an integration YAML "
```
ShacharKidor:
```suggestion
            f"'{input_path}' does not appear to be an integration YAML. "
```

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
+"""
+Validation logic for the upload-custom-integration command.
+
```
ShacharKidor: Not sure we need to have this in a separate file; why not add the function to the upload_custom_integration.py file? What do you think?

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+Enforces the '_copy' marker naming convention on integration IDs and display
+names before upload, preventing ID conflicts with official system pack integrations.
```
ShacharKidor:
```suggestion
Enforces the '_copy' marker naming convention on integration IDs and
name before upload, preventing ID conflicts with official system pack integrations.
```

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+
+Background
+----------
+If a custom integration is uploaded with an ``id`` that matches a system
+integration's ``id``, any subsequent attempt to download the system pack that
```
ShacharKidor:
```suggestion
integration's ``id``, any subsequent attempt to install the system pack that
```

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+contains that integration will fail with a system error.  The ``_copy`` suffix
+is the established XSOAR convention that distinguishes custom (user-owned)
```
ShacharKidor:
```suggestion
is the established convention that distinguishes custom (user-owned)
```

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+# The required suffix that marks an integration as a user-owned copy.
+COPY_MARKER: str = "_copy"
+
+# YAML key path for the canonical integration ID.
+_COMMONFIELDS_KEY = "commonfields"
+_ID_KEY = "id"
+_NAME_KEY = "name"
+
+
+def _load_yaml(path: Path) -> dict:
```
ShacharKidor: Please check if we already have a similar function to use instead of implementing a new one here

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+    with path.open("r", encoding="utf-8") as fh:
+        return yaml.load(fh) or {}
+
+
+def resolve_integration_yaml(path: Path) -> Path:
+    """If *path* is a directory, resolve the single integration YAML file inside it.
+
+    Users often pass the integration folder (e.g.,
+    ``-i Packs/MyPack/Integrations/MyIntegration_copy/``) rather than the
+    explicit ``.yml`` file path.  This helper transparently handles both forms.
```
ShacharKidor: Also here - if we already have this logic in the upload command, it's better to use the existing function.

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+
+from demisto_sdk.commands.common.logger import logger
+
+# The required suffix that marks an integration as a user-owned copy.
+COPY_MARKER: str = "_copy"
+
+# YAML key path for the canonical integration ID.
+_COMMONFIELDS_KEY = "commonfields"
+_ID_KEY = "id"
+_NAME_KEY = "name"
```
ShacharKidor:
```suggestion
COMMONFIELDS_KEY = "commonfields"
ID_KEY = "id"
NAME_KEY = "name"
```

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+    Returns:
+        ``True`` when the file looks like an integration YAML, ``False``
+        otherwise (including when the file cannot be read or parsed).
+    """
+    target_path = resolve_integration_yaml(path)
+    if not target_path.is_file():
+        return False
+    try:
+        data = _load_yaml(target_path)
+        return _COMMONFIELDS_KEY in data or ("script" in data and "category" in data)
```
ShacharKidor: Are we sure about this condition for verifying that this is an integration yml?

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+        typer.BadParameter: When the YAML cannot be parsed or the expected
+            fields are missing.
+    """
+    target_path = resolve_integration_yaml(path)
+    try:
+        data = _load_yaml(target_path)
+    except Exception as exc:
+        raise typer.BadParameter(
+            f"Could not parse YAML file '{target_path}': {exc}"
+        ) from exc
```
ShacharKidor: This part is redundant since we already have it in 
is_integration_yaml
and in 
resolve_integration_yaml

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+        )
+        return
+
+    raise typer.BadParameter(
+        f"\n"
+        f"  Integration field(s) missing the '{COPY_MARKER}' marker:\n"
+        f"    {missing_summary}\n"
+        f"\n"
+        f"  RISK: Uploading a custom integration whose ID matches a system\n"
+        f"  integration ID will cause subsequent downloads of the system pack\n"
```
ShacharKidor:
```suggestion
        f"  integration ID will cause subsequent installations of the system pack\n"
```

demisto_sdk/__main__.py
```diff
@@ -172,6 +172,23 @@ def register_commands(_args: list[str] = []):  # noqa: C901
...
 
+    if command_name == "upload-custom-integration" or register_all:
+        from demisto_sdk.commands.upload.upload_custom_integration_setup import (
+            upload_custom_integration,
+        )
+
+        app.command(
+            name="upload-custom-integration",
+            help=(
+                "Upload a custom integration to Cortex Platform with '_copy' marker "
```
ShacharKidor: Let's keep only the first sentence here since you already have detailed help sections in the command itself for the args.
`"Upload a custom integration to Cortex Platform with '_copy' marker safety enforcement. "`

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+----------
+If a custom integration is uploaded with an ``id`` that matches a system
+integration's ``id``, any subsequent attempt to download the system pack that
+contains that integration will fail with a system error.  The ``_copy`` suffix
+is the established XSOAR convention that distinguishes custom (user-owned)
+copies from system-managed originals.
+"""
+
+from pathlib import Path
+from typing import Tuple
```
ShacharKidor: Use tuple[str, str] instead of from typing import Tuple. Python 3.10 is the minimum, and the repo style prefers built-in generics. 
```suggestion
```

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+    if path.is_file():
+        return path
+    if path.is_dir():
+        yml_files = [
+            f
+            for f in list(path.glob("*.yml")) + list(path.glob("*.yaml"))
+            if not f.name.startswith(".")
+        ]
+        if len(yml_files) == 1:
+            return yml_files[0]
```
ShacharKidor: what if we have a directory with more than one yml file?

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+        The resolved ``.yml`` / ``.yaml`` file path when *path* is a directory
+        containing exactly one non-hidden YAML file; otherwise *path* unchanged.
+    """
+    if path.is_file():
+        return path
+    if path.is_dir():
+        yml_files = [
+            f
+            for f in list(path.glob("*.yml")) + list(path.glob("*.yaml"))
+            if not f.name.startswith(".")
```
ShacharKidor: should be covered in the path.glob("*.yaml")
```suggestion
```

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+    Returns:
+        A ``(integration_id, integration_name)`` tuple.
+
+    Raises:
+        typer.BadParameter: When the YAML cannot be parsed or the expected
+            fields are missing.
+    """
+    target_path = resolve_integration_yaml(path)
+    try:
+        data = _load_yaml(target_path)
```
ShacharKidor: We already loaded the yaml before ... maybe better to do it once.

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+        ``True`` when the file looks like an integration YAML, ``False``
+        otherwise (including when the file cannot be read or parsed).
+    """
+    target_path = resolve_integration_yaml(path)
+    if not target_path.is_file():
+        return False
+    try:
+        data = _load_yaml(target_path)
+        return _COMMONFIELDS_KEY in data or ("script" in data and "category" in data)
+    except Exception:
```
ShacharKidor: Bare except Exception: masks real failures. If the file exists but has a YAML syntax error, is_integration_yaml returns False and the caller then raises the misleading "does not appear to be an integration YAML" error instead of "invalid YAML on line X". Please narrow to except (yaml.YAMLError, OSError): (or the equivalent for YAML_Handler), or log at debug level before returning False.

demisto_sdk/commands/upload/integration_copy_validator.py
```diff
@@ -0,0 +1,203 @@
...
+            "Proceeding because --force-id was explicitly set."
+        )
+        return
+
+    raise typer.BadParameter(
+        f"\n"
+        f"  Integration field(s) missing the '{COPY_MARKER}' marker:\n"
+        f"    {missing_summary}\n"
+        f"\n"
+        f"  RISK: Uploading a custom integration whose ID matches a system\n"
```
ShacharKidor: L177-183 & L186-203 - Duplicated "why _copy matters" prose in both the warning and the error message. Extract the risk explanation into a module-level constant (_COPY_MARKER_RISK_EXPLANATION) and reference it from both branches - single source of truth, and the wording won't drift between the two paths.

ShacharKidor (changes requested):
Hi @Shir2611, 
Nice work! 

Added several comments please see below. 

Also, we discussed setting the **marketplace to platform** in the call to the upload command - please add that. 

**Edge cases the author may have missed**:

1. Marketplace default silently becomes XSOAR - single most important issue; documented above.
2. Directory with 2+ YAML files - returned unchanged; produces a misleading downstream error.
3. YAML file with a syntax error - is_integration_yaml swallows it and reports "not an integration YAML".
4. commonfields present but not a dict - guarded in code but not tested.
5. id literally equal to "_copy" - passes validation; almost certainly not intended.
6. Case sensitivity - _COPY, _Copy, etc. silently fail.
7. Symlinked directories / relative-path resolution - Typer's resolve_path=True handles the CLI path, but resolve_integration_yaml doesn't call .resolve() on the glob results; likely fine, worth a mental check.
8. upload_content_entity raises typer.Exit - Uploader inside always raises typer.Exit(code) on completion. Any exception-based caller of upload_custom_integration_entity (e.g. a Python API user, not the CLI) will see a typer.Exit bubble up. Consider documenting this in the docstring or catching + re-raising a clearer exception for programmatic use.

demisto_sdk/__main__.py
Shir2611: true, this is now fixed
```
# The marketplace is fixed for this command — Cortex Platform only.
# Passing any other value to upload_content_entity is intentionally prevented
# by the explicit function signature (no marketplace parameter exposed).
_MARKETPLACE = "platform"
```

demisto_sdk/commands/upload/upload_custom_integration_setup.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration_setup.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration_setup.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration_setup.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration_setup.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration_setup.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration_setup.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration_setup.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration.py
Shir2611: done

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: done

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: done

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: done

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: done

demisto_sdk/__main__.py
Shir2611: done

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: done

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: done. now raises typer.BadParameter with a clear message listing the conflicting files when a directory contains more than one YAML.

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: pathlib glob("*.yml") matches hidden files like ".hidden.yml",so we explicitly exclude them here.

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: done

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: done

demisto_sdk/commands/upload/upload_custom_integration_setup.py
Shir2611: these are mandatory params in the func's decorator `@logging_setup_decorator` 
we must have them mentioned, other wise we get this error 
`TypeError: upload_custom_integration() got an unexpected keyword argument 'console_log_threshold'`
also this is the implementation in other functions in the repo that use this decorator, so i am aligned with the repo's logic

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: sure, added to the upload_custom_integration.py file

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: done

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: No equivalent exists in the codebase.

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: according to idex this is the convention for module-private constants

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: keep only this 
return _COMMONFIELDS_KEY in data

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: this is the updated code
`try:
        data = get_yaml(target_path) or {}
        return _COMMONFIELDS_KEY in data
    except OSError:
        return False
    except Exception as exc:
        raise typer.BadParameter(
            f"YAML syntax error in '{target_path}': {exc}"
        ) from exc
`

demisto_sdk/__main__.py
ShacharKidor: Do you have any idea why the command worked before the change? 
Meaning - why it is needed to set a marketplace if it was uploaded successfully to a platform tenant without the flag? Can you please check that?

demisto_sdk/commands/upload/integration_copy_validator.py
ShacharKidor: TBH - I'm not sure. 
@Shir2611  -Do you know how we verify such things in other commands?

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: done

ShacharKidor: @Shir2611 missed that one

ShacharKidor (approved):
@Shir2611, 
Great work! 
Just review the 3 unresolved comments left from the review.

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: this is fixed, this file is outdated (does not exist anymore)

demisto_sdk/__main__.py
Shir2611: this is as the .env file had the DEMISTO_BASE_URL set to platform 
so the upload was successful even when we did not explicitly mentioned the mp

demisto_sdk/commands/upload/integration_copy_validator.py
Shir2611: we have to have this cond as the id field is in the commonfields section
```
commonfields:
  id: Polar Security_copy
```
i dont see any other commands that check for yml file

PR #5477, opened 2026-08-03
Jmizrahi/connectus validators

joeymizrahi:


demisto_sdk/commands/common/constants.py
```diff
@@ -1328,7 +1328,7 @@ class FileType(StrEnum):
 
 # A list of validation error codes that must always execute, regardless of ignore settings.
 # This addresses unique cases where validation must run first, then filter the relevant results afterward.
-ALWAYS_RUN_ON_ERROR_CODE = ["GR107", "GR109", "CO157"]
+ALWAYS_RUN_ON_ERROR_CODE = ["GR107", "GR109"]
```
JudahSchwartz, line 1331: explain?

demisto_sdk/commands/validate/validators/CO_validators/CO108_is_valid_general_config_description.py
```diff
@@ -0,0 +1,80 @@
...
+            if capabilities_metadata is None:
+                # No capabilities.yaml - nothing to validate here.
+                continue
+
+            details: List[str] = []
+
+            general_configurations = capabilities_metadata.general_configurations
+            if general_configurations is None:
+                details.append(
+                    "general_configurations must be present with description "
```
JudahSchwartz, line 54: what should be underneath these general configurations that it must always be present?

demisto_sdk/commands/validate/validators/CO_validators/CO120_is_proxy_and_insecure_exists.py
```diff
@@ -0,0 +1,208 @@
...
+
+        required_families = self._integration_families(handler)
+        if not required_families:
+            return results  # Integration declares no proxy/insecure param.
+
+        exposed_content_names = self._resolved_content_names(handler)
+
+        for family in sorted(required_families):
+            aliases = PROXY_ALIASES if family == "proxy" else INSECURE_ALIASES
+            if aliases & exposed_content_names:
```
JudahSchwartz, line 184: its not a legitimate case to reference this not from the yml?
I actually think in the future proxy and insecure should only be handled inside the baseclient, not sure if requiring it in the parameters makes sense.

Think we should only check that they are available fields in the connector

demisto_sdk/commands/validate/validators/CO_validators/CO121_is_valid_interpolation.py
```diff
@@ -0,0 +1,290 @@
...
+    ".identifier",
+    ".password",
+    ".credentials.sshkey",
+)
+
+# XSOAR ParameterType.AUTH = 9 (see demisto_sdk.commands.common.constants).
+_CREDENTIALS_PARAM_TYPE = 9
+
+
+class IsValidInterpolationValidator(ConnectorsValidator[ContentTypes]):
```
JudahSchwartz, line 39: think we should add here that fields that have ispublish true shouldnt be in the interpolation mapping

demisto_sdk/commands/validate/validators/CO_validators/CO123_is_profile_fields_covered.py
```diff
@@ -0,0 +1,173 @@
...
+
+ContentTypes = Connector
+
+# Non-auth params must publish to the running integration - except for
+# 'engine_mode', which is a UI-only field controlling the engine picker
+# rather than an integration parameter.
+_PUBLISH_EXEMPT_IDS = {"engine_mode"}
+
+
+class IsProfileFieldsCoveredValidator(ConnectorsValidator[ContentTypes]):
```
JudahSchwartz, line 27: should we add a validation that if the field matches a real field in the profile,

like client_id for oauth profile etc, 
that it cant have ispublish?

Should help keep some secrets safe

demisto_sdk/commands/validate/validators/CO_validators/CO171_is_collection_sub_capability_fetch_flag_valid.py
```diff
@@ -0,0 +1,200 @@
+"""CO171 - forward direction: every subscribed collection sub-capability
+must have a matching fetch-flag emission in the handler's serializer.
```
JudahSchwartz, line 2: didnt we check this under is valid fetch?

demisto_sdk/commands/validate/validators/CO_validators/CO175_no_removed_connector_params.py
```diff
@@ -0,0 +1,99 @@
+from __future__ import annotations
+
+from typing import Iterable, List, Set, cast
+
+from demisto_sdk.commands.common.constants import GitStatuses
```
JudahSchwartz, line 5: think we also need to have a no added withoujt default and required

JudahSchwartz (approved):

demisto_sdk/commands/common/constants.py
joeymizrahi: idk what this is. maybe @YuvHayun or @yedidyacohenpalo can tell us?
i just removed the CO157 because it no longer exists

demisto_sdk/commands/validate/validators/CO_validators/CO120_is_proxy_and_insecure_exists.py
joeymizrahi: Not sure i get what you mean. 
If these fields are in the integration, we are ensuring they are in the connector. 
Thats what we discussed in the review.

JudahSchwartz: i thought this is checking the other way?

demisto_sdk/commands/validate/validators/CO_validators/CO108_is_valid_general_config_description.py
joeymizrahi: instance name

demisto_sdk/commands/validate/validators/CO_validators/CO121_is_valid_interpolation.py
joeymizrahi: sure, ill add it

demisto_sdk/commands/validate/validators/CO_validators/CO175_no_removed_connector_params.py
joeymizrahi: added in CO179

demisto_sdk/commands/validate/validators/CO_validators/CO171_is_collection_sub_capability_fetch_flag_valid.py
joeymizrahi: CO130 owns the fetch-issues UI contract (four required fields with the right shape — CO171 knows nothing about this); CO171 owns the fetch-family serializer-flag contract across all 5 collection capabilities (CO130 covers only 1 of the 5). Removing either leaves a real coverage gap.

demisto_sdk/commands/validate/validators/CO_validators/CO123_is_profile_fields_covered.py
joeymizrahi: Adi said he would add something like that

demisto_sdk/commands/validate/validators/CO_validators/CO120_is_proxy_and_insecure_exists.py
joeymizrahi: no.

demisto_sdk/commands/common/constants.py
yedidyacohenpalo: It’s a slightly different ignore mechanism. Usually, it’s used for graph validations, where the errors are removed at the end if an ignore is defined.

yedidyacohenpalo (approved):

demisto_sdk/commands/content_graph/objects/connector.py
```diff
@@ -799,6 +818,9 @@ def _rebuild_nested_from_neo4j(cls, values: Dict[str, Any]) -> Dict[str, Any]:
 
     # === Parsed sub-models (populated by parser, excluded from serialization) ===
     connection: Optional[ConnectorConnectionData] = Field(None, exclude=True)
+    capabilities_metadata: Optional[ConnectorCapabilitiesData] = Field(
```
ShacharKidor, line 821: I think we need to add this to the strict objects as well. 
@yedidyacohenpalo - please verify my thoughts.

ShacharKidor (changes requested):

demisto_sdk/commands/content_graph/objects/connector.py
yedidyacohenpalo: strict_objects holds the Pydantic schemas we enforce via ST110 validation. There is currently no demisto_sdk/commands/content_graph/strict_objects/connector.py, which means there is no Pydantic enforcement on connectors: nothing prevents adding unknown fields or omitting required ones. 
@joeymizrahi @JudahSchwartz Your call whether to add such a schema and validation, or to fold it into ST110 so it's covered together.

joeymizrahi: i think this is more of an infra decision. 
Prefer your team to make the decision as you have more knowledge in this area.

demisto_sdk/commands/common/constants.py
joeymizrahi: So what should we be doing with this list?
Dont fully understand, whats the rule of thumb, what validations to add here?

demisto_sdk/commands/content_graph/objects/connector.py
ShacharKidor: As discussed offline, we should add an ST validation for Connectors - need to open a ticket for that.

ShacharKidor (approved):

PR #5478, opened 2026-08-04
fix(validate): honor per-file .pack-ignore for related-file validators

BEAdi:
is_error_ignored returned False for any validator declaring related_file_type as soon as none of the related files matched, so it never consulted the content item's own [file:...] per-file ignore section. For AgentixActions (which have no SKILL_CONTENT related file) this made AG112 impossible to ignore via .pack-ignore.

Fall through to the main content per-file ignore when no related file matches. Add is_error_ignored regression tests.

## Related Issues
fixes:

## Description

test MR - [link](https://gitlab.xdr.pan.local/xdr/cortex-content/content-private/-/merge_requests/584/pipelines)
failure before the fix -[link](https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/jobs/69649672)
AG112 passing with the fix -  [link](https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/jobs/69614984)

## SDK Nightly

SDK Nightly link:

### DIFF
--- a/demisto_sdk/commands/validate/validators/base_validator.py
+++ b/demisto_sdk/commands/validate/validators/base_validator.py
@@ -297,10 +297,13 @@ def is_error_ignored(
         3. If ``err_code`` is listed under the new ``[pack]`` section of the
            pack's ``.pack-ignore`` file -> ignore for every item in the pack
            (and for related files of those items).
-        4. If a ``related_file_type`` is provided -> ignore only when the
-           code is listed under the related file's per-file section.
-        5. Otherwise -> ignore only when the code is listed under the
-           content item's per-file section.
+        4. If a ``related_file_type`` is provided -> ignore when the code is
+           listed under any of the related files' per-file sections. If none
+           of the related files match (e.g. the related file does not exist
+           for this content type, such as an AgentixAction which has no
+           SKILL_CONTENT), fall through to rule 5.
+        5. Ignore when the code is listed under the content item's own
+           per-file section.
 
     Args:
         err_code: The validation's error code.
@@ -340,10 +343,13 @@ def is_error_ignored(
                     return True
             except Exception:
                 continue
-        return False
-    else:
-        # If the validation should run on the main content, will check if the validation's error code is ignored by the file.
-        return err_code in content_item.ignored_errors
+        # None of the related files carried the ignore (e.g. the related file
+        # does not exist for this content type, such as an AgentixAction which
+        # has no SKILL_CONTENT). Fall through to the main content's per-file
+        # ignore so a `[file:...]` section on the item itself is still honored.
+
+    # If the validation should run on the main content, will check if the validation's error code is ignored by the file.
+    return err_code in content_item.ignored_errors
 
 
 class ValidationResult(BaseResult, BaseModel):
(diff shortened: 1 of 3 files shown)


demisto_sdk/commands/validate/validators/base_validator.py
```diff
@@ -453,10 +453,13 @@ def is_error_ignored(
...
-           code is listed under the related file's per-file section.
-        5. Otherwise -> ignore only when the code is listed under the
-           content item's per-file section.
+        4. If a ``related_file_type`` is provided -> ignore when the code is
+           listed under any of the related files' per-file sections. If none
+           of the related files match (e.g. the related file does not exist
+           for this content type, such as an AgentixAction which has no
+           SKILL_CONTENT), fall through to rule 5.
+        5. Ignore when the code is listed under the content item's own
+           per-file section.
```
ShacharKidor, line 461:
```suggestion
        4. If a ``related_file_type`` is provided -> ignore when the code is
           listed under any of the related files' per-file sections. If none
           of the related files match (e.g. the related file does not exist
           for this content type), fall through to rule 5.
        5. Ignore when the code is listed under the content item's own
           per-file section.
```

demisto_sdk/commands/validate/validators/base_validator.py
```diff
@@ -496,10 +499,13 @@ def is_error_ignored(
...
-        return False
-    else:
-        # If the validation should run on the main content, will check if the validation's error code is ignored by the file.
-        return err_code in content_item.ignored_errors
+        # None of the related files carried the ignore (e.g. the related file
+        # does not exist for this content type, such as an AgentixAction which
+        # has no SKILL_CONTENT). Fall through to the main content's per-file
+        # ignore so a `[file:...]` section on the item itself is still honored.
+
+    # If the validation should run on the main content, will check if the validation's error code is ignored by the file.
```
ShacharKidor, line 509: Two clarifying questions:

1. Why does AgentixAction declare SKILL_CONTENT as a related file if it doesn't own one? Is this a new case where the related file is optional per-item (an action may or may not be paired with a skill)?

2. Was the previous behavior fine until now because, for every other validator, the declared related_file_type was always guaranteed to exist on the item — making AgentixAction the first content type where that assumption breaks?

demisto_sdk/commands/validate/validators/base_validator.py
```diff
@@ -496,10 +499,13 @@ def is_error_ignored(
                     return True
             except Exception:
```
ShacharKidor: Might be worth adding a debug log here: 
```
except Exception as e:
    logger.debug(f"is_error_ignored: skipped related_file={related_file.value!r} on {content_item!r}: {e!r}")
    continue
```

Since errors are silently swallowed here, the blast radius now grows after the change.

ShacharKidor (changes requested):
Very nice! 

Added minor suggestions.

demisto_sdk/commands/validate/validators/base_validator.py
BEAdi: 1. https://github.com/demisto/demisto-sdk/pull/5491 Inbal will remove it
2. Yes

ShacharKidor (approved):

PR #5480, opened 2026-08-06
improvements_connectus_validate

rshunim:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-17519

## Description

- description: The **validate** command now builds/updates the content graph automatically
    when a selected validation needs it (graph validations and the connectors flow),
    so the `--graph` flag is no longer required for those. `--graph` remains as an
    optional connect-only shortcut for CI where the graph is built beforehand.
  type: feature
- description: Reduced the number of content-graph queries performed by the connectors
    validation flow by batching the handler/integration cross-matching lookups into
    a single query per content type instead of one query per item.

## SDK Nightly

SDK Nightly link:


yedidyacohenpalo (approved):

ShacharKidor (changes requested):
Only two things I would actually block a merge on. Everything else is polish.

1. The failure path closes a graph the caller owns
In [initializer.py](vscode-webview://0hbisp6d0es61rlr4bpq38ptu0dnau6ud9vihu05jna05er7gogc/demisto-sdk/demisto_sdk/commands/validate/initializer.py:1469):

```
BaseValidator.ensure_graph_initialized()
try:
    filtered = self._cross_match_and_expand(...)
except Exception:
    if BaseValidator.graph_interface:
        BaseValidator.graph_interface.close()
        BaseValidator.graph_interface = None
    raise

```
ensure_graph_initialized is idempotent - if --graph was passed in CI, the graph is already wired by the caller. If _cross_match_and_expand then raises, this block closes the caller's graph and nulls the class variable. ValidateManager.run_validations's own finally will then silently no-op, and any subsequent code that touches BaseValidator.graph_interface will fault.

Fix: only close if this call was the one that opened it.
```

opened_here = BaseValidator.graph_interface is None
BaseValidator.ensure_graph_initialized()
try:
    filtered = self._cross_match_and_expand(...)
except Exception:
    if opened_here and BaseValidator.graph_interface:
        BaseValidator.graph_interface.close()
        BaseValidator.graph_interface = None
    raise

```
2. The perf test does not actually test the perf fix
[test_connector_table_is_fetched_once_across_unmatched_integrations](vscode-webview://0hbisp6d0es61rlr4bpq38ptu0dnau6ud9vihu05jna05er7gogc/demisto-sdk/demisto_sdk/commands/validate/tests/validators_test.py:1421) claims to prove the connector table is fetched once "regardless of how many unmatched integrations there are" - but the test only sets up one unmatched integration (Orphan). _all_graph_connectors would be called once even under the old buggy per-integration-scan code, so this test would pass against the regression it is meant to guard against.

Fix: put at least two unmatched integrations into the working set and assert mock_connectors.call_count == 1. Without that, the perf regression can silently come back and the test suite will stay green.

rshunim:
Both are real catches, fixed in the latest push.

1. Failure path closing a caller-owned graph - you're right, and I traced the exact path: attach_graph_interface_if_requested() in validate_setup.py wires BaseValidator.graph_interface when --graph is passed, and that happens before ValidateManager.__init__ runs collection. So in CI, ensure_graph_initialized() returns early and the initializer owns nothing - but the old except block closed and nulled it anyway. run_validations()'s finally would then see None and silently no-op, and every later reader (_graph_search_integration, _all_graph_connectors, all graph validators) would hit a dead driver. Applied your opened_here guard verbatim, plus a regression test test_caller_owned_graph_is_not_closed_when_cross_matching_fails that pre-wires a mock graph and asserts it is neither closed nor nulled on failure. The existing test_graph_is_closed_when_cross_matching_fails still covers the opened-here path.

2. Perf test was vacuous - confirmed. _graph_expand_connectors loops over unmatched integrations, so with a single Orphan the call count is 1 either way. Test now seeds two unmatched integrations and asserts mock_connectors.call_count == 1. I sanity-checked it by temporarily putting the per-integration scan back: the test fails with assert 2 == 1, so it now genuinely guards the regression.

@ShacharKidor

ShacharKidor (approved):

PR #5488, opened 2026-08-10
Fix integration support_level_header

kamalq97:
## Related Issues
fixes: [CIAC-17642](https://jira-dc.paloaltonetworks.com/browse/CIAC-17642)

## Description
Fixed an issue where the `ST110` validation was incorrectly flagging valid `supportlevelheader` values (community, partner, developer, xsoar) as false positives in integration YMLs.


BEAdi (commented):
Overall it looks good, you will need to connect this PR to a content PR (using a connecting infra branch, there is a skill in content-configuration that does this) in order to check that nothing broke and that uploading a pack to a tenant behaves the same way (change content items with different SupportLevels in the content PR that you opened and check that their installation on the machine is working in the build), and upload to a tenant (best is using the bypass url created in the PR) and install the pack in order to see that everything works the same way as before.

PR #5492, opened 2026-08-11
feat(script): support promptConfig.modelTier for AI prompts

mayyagoldman:
Add the modelTier field to the Script promptConfig object, so AI prompt content can pin an LLM tier instead of a concrete model id.

The server resolves the tier at run time against the agentix-hub models cache (ServiceAI.resolveModel), which means content pins the tier and the backing model can change without a content release. The legacy top-level `model` field is deprecated server-side and is left untouched here.

Valid tiers are Flash, Thinking and Pro. An absent tier stays valid and resolves to the hub-advertised default model, so existing AI prompts are unaffected.

Changes:
- schemas/script.yml: modelTier under promptConfig.mapping, enum-constrained
- content_graph/objects/script.py: model_tier on PromptConfig
- content_graph/strict_objects/script.py: ModelTier StrEnum + model_tier on PromptConfig, so an unsupported tier surfaces as an ST110 error

Note: the SDK is stricter than the server on casing. The server matches tiers with strings.EqualFold, while this enum requires the canonical form.

Tests: parser round-trip with and without a tier, plus ST110 coverage for valid, invalid and absent tiers.

## Related Issues
fixes:

## SDK Nightly

SDK Nightly link:

### DIFF
--- a/.changelog/5492.yml
+++ b/.changelog/5492.yml
@@ -1,0 +1,4 @@
+changes:
+- description: Added the **modelTier** field under the **promptConfig** object in the Script schema, allowing AI Tasks to pin an LLM tier (Flash, Thinking or Pro) instead of a concrete model id. An absent tier remains valid and resolves to the default model.
+  type: feature
+pr_number: 5492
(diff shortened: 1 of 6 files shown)


tcarmeli1 (approved):
Good!

ShacharKidor (approved):

PR #5504, opened 2026-08-20
fix(git_util): derive pack first-created date from the add commit

rshunim:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CRTX-221269

## Description
Derive the file creation date from the commit that first added the file (following renames) instead of the oldest reachable commit, and log a warning when the result may be unreliable (for example, a shallow clone).

### DIFF
--- a/demisto_sdk/commands/common/git_util.py
+++ b/demisto_sdk/commands/common/git_util.py
@@ -1,6 +1,6 @@
 import os
 import re
-from datetime import datetime
+from datetime import datetime, timezone
 from functools import lru_cache
 from pathlib import Path
 from typing import List, Optional, Sequence, Set, Tuple, Union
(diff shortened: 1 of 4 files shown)


rshunim:
Checked for both cases, when a pack like Base contains the "created" filed in its metadata and the other case for "AHA" and "Traceable" which have not this field and made sure in this case the first commit is being taken.

demisto_sdk/commands/common/tests/git_util_test.py
```diff
@@ -166,3 +180,167 @@ def test_get_file_creation_date(git_repo: Repo):
...
+    - Retrieving the creation time of the given file.
+
+    Then:
+    - The current time is returned, in the expected format.
+    """
+    file_creation_date = git_repo.git_util.get_file_creation_date(
+        Path("no_such_file.json")
+    )
+
+    datetime.strptime(file_creation_date, ISO_TIMESTAMP_FORMAT)  # raises if invalid
```
BEAdi: Use `freeze_time` fixture  to assert the fallback returns the current time

demisto_sdk/commands/common/git_util.py
```diff
@@ -1264,8 +1264,67 @@ def stage_file(self, file_path: Union[Path, str]):
...
         if commits := list(self.repo.iter_commits(paths=file_path)):
+            logger.warning(
+                f"Could not find the commit that added '{file_path}', "
+                "falling back to the oldest reachable commit. The date may be inaccurate."
+            )
             first_commit = commits[-1]
-            return first_commit.authored_datetime.strftime(ISO_TIMESTAMP_FORMAT)
+            return self._normalize_iso_date(first_commit.authored_datetime.isoformat())
+
         return datetime.now().strftime(ISO_TIMESTAMP_FORMAT)
```
BEAdi: Do you want to `_normalize_iso_date` this date too?
It also doesn't have a timezone, which can be fixed now

BEAdi (commented):
Great work!
Did you check this on a machine?
Can you check in a real test upload flow, that it takes the correct value and doesn't log the error?

rshunim:
@BEAdi I checked again after fix on XSOAR tenant, 2 Packs: Slack (has "created" field) and AHA (does not have it)
and as expected, the Slack has the right "created" value on the UI and AHA as well with the first absolute commit.
Now I'm waiting for the "test-upload-flow" and then we can move on

Here is a link

https://github.com/demisto/content/pull/45618

I used the fixed SDK version and made some changes in the 2 Packs I mentioned above

PR #5508, opened 2026-08-26
feat(content-graph): add optional supportedFeatures field

rshunim:
Add an optional `supportedFeatures` field (list[str]) to every content item type and to pack_metadata.json, so content items can declare which platform features they require.

Semantics:
- An item that does not author the key inherits its pack's value.
- Strip only on None, never on falsiness, so an explicitly authored empty list is preserved as an intentional "no features" declaration and is distinct from an absent key.
- When neither the item nor its pack author it, the key is omitted entirely from the output and marketplace artifacts - not [], not null.

Inheritance is resolved at the single ContentItemParser.parse seam rather than by threading a new argument through ~35 parser subclasses.

Connectors are excluded: Connector subclasses ContentItem so it inherits the attribute, but no connector schema or strict model gained the field, so it can never be authored there.

Covers all 40 content-item schemas and 35 strict-model files. Note this is wider than the supportedModules precedent, which only exists in 21 schemas - the types lacking it would otherwise reject the new key as an extra field.

## Related Issues
fixes:

## SDK Nightly

SDK Nightly link:
https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/pipelines/12937184


rshunim:
Link to SDK Nightly
https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/pipelines/12692635
branch "feat/supported-features-field" for all Infra, SDK and Content

israelpoli (approved):
Excellent 💪 
Great job 🥇

ShacharKidor (approved):

PR #5520, opened 2026-09-10
feat(validate): add temp GR117 warning for duplicate AgentixAction IDs

rshunim:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-18005

## Description

AgentixAction items with duplicate IDs currently exist in the content-private repo and are scheduled to be fixed in December. As a stopgap, all of GR105 was downgraded to a warning, which stopped it from blocking merges for every content type.

This PR narrows the blast radius: GR105 now skips AgentixAction items so it can be restored to an error for everything else, and a new temporary GR117 reports duplicate AgentixAction IDs as a warning until the duplications are resolved.

GR117 is intentionally throwaway - when the duplications are fixed, revert this commit to delete the three GR117 files, drop GR117 from the two warning lists, and restore AgentixAction in GR105.

## SDK Nightly

SDK Nightly link:


rshunim:
Testing it in a Content branch
https://github.com/demisto/content/pull/45862
want to make sure GR105 now passes and the temp one GR117 warns

demisto_sdk/commands/validate/validators/GR_validators/GR105_duplicate_content_id.py
```diff
@@ -118,5 +115,10 @@ def obtain_invalid_content_items_using_graph(
             for content_item, duplicates in self.graph.validate_duplicate_ids(
                 paths_of_content_items_to_validate
             )
+            # AgentixAction and AgentixAgent duplicate IDs are temporarily reported by
+            # GR117 (a warning) until the existing duplications in the private content
+            # repo are fixed.
+            if content_item.content_type
+            not in (ContentType.AGENTIX_ACTION, ContentType.AGENTIX_AGENT)
```
yedidyacohenpalo, line 122:
```suggestion
```
The condition appears to be redundant. The method parameters strictly filter content_items by the ContentTypes union type, which does not include AGENTIX_ACTION or AGENTIX_AGENT. We can safely remove this check.

demisto_sdk/commands/validate/validators/GR_validators/GR117_duplicate_agentix_action_id.py
```diff
@@ -0,0 +1,60 @@
...
+                    content_item.object_id,
+                    get_relative_path_from_packs_dir(str(duplicate.path)),
+                ),
+                content_object=content_item,  # type: ignore[arg-type]
+            )
+            for content_item, duplicates in self.graph.validate_duplicate_ids(
+                paths_of_content_items_to_validate
+            )
+            if content_item.content_type
+            in (ContentType.AGENTIX_ACTION, ContentType.AGENTIX_AGENT)
```
yedidyacohenpalo:
```suggestion
```
Similar to GR105, the condition here is also redundant. Since ContentTypes for this validator is explicitly restricted to Union[AgentixAction, AgentixAgent], all items being processed are guaranteed to be of these types. We can safely remove this check.

yedidyacohenpalo (approved):

BEAdi (approved):

PR #5521, opened 2026-09-15
fix(validate): run PA114 on deprecated packs

rshunim:
## Related Issues
fixes: https://jira-dc.paloaltonetworks.com/browse/CIAC-18237

## Description

PA114 was silently skipped on deprecated packs (the base deprecated filter dropped the Pack object while keeping its content items, causing the KeyError) - it now runs on them via run_on_deprecated, so changes to a deprecated pack correctly require a version bump.

## SDK Nightly

SDK Nightly link:


rshunim:
@yedidyacohenpalo 
Here you can see the issue which I fixed
https://gitlab.xdr.pan.local/xdr/cortex-content/content/-/jobs/78720605

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
```diff
@@ -185,13 +197,27 @@ def obtain_invalid_content_items(
...
                     content_packs_ids_to_bump.add(content_item.pack_id)  # type: ignore[union-attr]
+                    bump_trigger_by_pack_id[content_item.pack_id] = content_item  # type: ignore[union-attr,index]
 
         # Go over all the pack ids that need to be bumped.
         for pack_id in content_packs_ids_to_bump:
-            # Access them via the dict that was created earlier
-            pack = content_packs[pack_id]
+            # Access them via the dict that was created earlier.
+            pack = content_packs.get(pack_id)
+            if pack is None or pack.old_base_content_object is None:
```
BEAdi, line 206: `pack.old_base_content_object` will also be None for new packs. New packs don't require RN, so they should be excluded previously in this validation; just add a test for this case to make sure the validation doesn't fail for new packs.

BEAdi (approved):

demisto_sdk/commands/validate/validators/PA_validators/PA114_pack_metadata_version_should_be_raised.py
rshunim: Done

